PR #2, opened 2020-08-04
Initial Github Actions workflow

jordan-griffin:
This PR will add our initial Github Actions CI workflows to the repo.

We need two workflows to start with:
- One to run tests whenever new code is pushed to a PR.
- Another to run tests and build out an AAR file when code is merged into master.

These pipelines can be expanded in the future but the aim for now is to get some simple pipelines up.

### DIFF
--- a/.github/workflows/android-feature.yml
+++ b/.github/workflows/android-feature.yml
@@ -12,17 +12,17 @@ jobs:
     name: Run Unit Tests
     
     # The type of runner that the job will run on
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     steps:
     # Checks-out your repository under $GITHUB_WORKSPACE, so the job can access it
     - uses: actions/checkout@v1
   
     # Sets up a JDK on the job so that we can run Java code
-    - name: Set up JDK 11
+    - name: Set up JDK 1.8
       uses: actions/setup-java@v1
       with:
-        java-version: 11
+        java-version: 1.8
 
     # Call Gradle to run the unit tests
     - name: Unit Tests
--- a/.github/workflows/android-feature.yml
+++ b/.github/workflows/android-feature.yml
@@ -2,13 +2,9 @@
 
 name: Android SDK Feature CI
 
-# Triggers the workflow on push to our working branches
+# Triggers the workflow whenever a pull request is opened, reopened or has commits pushed to it
 on:
-  push:
-    branches: 
-      - '*'
-      - '!master'
-      - '!release'
+  pull_request
 
 jobs:
   # This workflow contains a single job for running our tests
--- a/.github/workflows/android-feature.yml
+++ b/.github/workflows/android-feature.yml
@@ -1,6 +1,6 @@
 # A workflow for controlling the acions taken by CI whenever we push a commit to an open pull request
 
-name: Branch CI
+name: Android SDK Feature CI
 
 # Triggers the workflow on push to our working branches
 on:
--- a/.github/workflows/android-feature.yml
+++ b/.github/workflows/android-feature.yml
@@ -16,7 +16,7 @@ jobs:
 
     steps:
     # Checks-out your repository under $GITHUB_WORKSPACE, so the job can access it
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
   
     # Sets up a JDK on the job so that we can run Java code
     - name: Set up JDK 1.8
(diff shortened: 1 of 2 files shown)


.github/workflows/android-feature.yml
```diff
@@ -0,0 +1,30 @@
...
+on:
+  pull_request
+
+jobs:
+  # This workflow contains a single job for running our tests
+  test:
+    name: Run Unit Tests
+    
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
```
kennyklaviyo, line 15: we might want to specify a version to keep things consistent

.github/workflows/android-master.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      steps:
+        # Checks-out your repository under $GITHUB_WORKSPACE, so the job can access it
+        - uses: actions/checkout@v1
+
+        # Sets up a JDK on the job so that we can run Java code
+        - name: Set up JDK 11
+          uses: actions/setup-java@v1
+          with:
+              java-version: 11
```
kennyklaviyo, line 46: possibly not a problem but not too sure how well android build tools plays with java 11, i remember have issues with proper linking of classes but maybe that's changed. just pointing out we might end up changing this to 8

.github/workflows/android-master.yml
```diff
@@ -0,0 +1,58 @@
...
+
+        # Call Gradle to build our our debug AAR
+        - name: Build debug AAR
+          run: bash ./gradlew assembleDebug --stacktrace
```
kennyklaviyo, line 50: we probably need to setup android build tools first, this will probably fail due to missing license agreement

.github/workflows/android-feature.yml
jordan-griffin: Fair. I will update these to run against ubuntu-20.04 which supports the version of Java we want to build against.

.github/workflows/android-master.yml
jordan-griffin: Uuh good point. Android's Java support should be more up-to-date than it was before. Their support of JDK versions slowed down to push resources into the adoption of Kotlin I heard and with Java 8's end of life approaching at the end of 2020 I would really hope Google updated their support for higher JDK versions by now

.github/workflows/android-master.yml
jordan-griffin: Ya, I'm aware. Unfortunately I'll need to push out another PR first, just something that I had forgotten to ticket when storymapping.

.github/workflows/android-master.yml
kennyklaviyo: ah interesting, yeah we should definitely keep an eye out for java 11 support since thats the current LTS but to get going without issues Java 8 is the way to go

.github/workflows/android-master.yml
kennyklaviyo: sounds good

.github/workflows/android-master.yml
```diff
@@ -0,0 +1,58 @@
...
+          with:
+              java-version: 1.8
+
+        # Call Gradle to build our our debug AAR
+        - name: Build debug AAR
+          run: bash ./gradlew assembleDebug --stacktrace
+
+        # Stores the generated AAR to an outputs folder
+        - name: Store AAR
+          uses: actions/upload-artifact@v1
```
kennyklaviyo, line 54: we'll want to use v2 of the action https://github.com/actions/upload-artifact note here about version syntax: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-versioned-actions

kennyklaviyo: also one thing to note about artifacts, they will automatically expire and be deleted after 90 days

.github/workflows/android-master.yml
```diff
@@ -0,0 +1,58 @@
...
+  # This workflow contains a single job for running our tests
+  test:
+    name: Run Unit Tests
+
+    # The type of runner that the job will run on
+    runs-on: ubuntu-20.04
+
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so the job can access it
+      - uses: actions/checkout@v1
```
kennyklaviyo, line 21: also want to use v2 of checkout as well https://github.com/actions/checkout

.github/workflows/android-master.yml
jordan-griffin: That's ok with me I think. The storing of artifacts currently is a precursor step to us sending them off to Firebase Test Lab when we have that infrastructure set up. I don't think we'll need old SDKs for that.

PR #7, opened 2020-08-05
Jenny/08 05 2020/added new example

JenRen-tech:
# Purpose

**Type of PR**
* [ ] Bug Fix
* [X] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

TP-45807


.github/workflows/example-build.yml
```diff
@@ -0,0 +1,26 @@
...
+    # Sets up a JDK on the job so that we can run Java code
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+
+    # Compile the APK for testing
+    - name: Compile APK
+      run: bash ./example :example:assembleDebug --stacktrace
+
```
jordan-griffin, line 26: Can you add a simple step for running unit tests here? Should be able to basically copy/paste whats in the other Github Actions workflows we made for that

JenRen-tech: Added! At ln 28

jordan-griffin: Just requesting one more tiny change. Where you run the test can you add `- name: Unit Tests` just before? So that each step of the pipeline is named and labelled to make it easier to understand where things fail if they do.

PR #8, opened 2020-08-07
Jenny/08 06 2020/networking features

JenRen-tech:
# Purpose

**Type of PR**
* [ ] Bug Fix
* [X] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

TP-49005 Added network features. This is still a draft and not finished yet. As per the TP specs:

Adding the Android OS permissions for accessing network state and internet to the AndroidManifest of the core SDK module
Create a function that checks if the device is currently connected to the network
Create a private function which builds a simple HttPUrlConnection to perform our network requests. These should be wrapped in a function that accepts a URL and a payload (for POST requests). The request should have read and write timeouts attached. Requests should also have request properties attached to build out a User Agent string that we pump user information into.
Create two public functions called track and identify which call into the previously created abstract private network request function. They should accept body information as params but they should hit the pre-assigned URLs for the Klaviyo track and identify endpoints.
Unit Tests should exist for all functions. Actual networking calls and responses can be mocked out.
Performing any network batching is not a part of this ticket.


core/network.kt
```diff
@@ -0,0 +1,46 @@
...
+        val networkInfo: NetworkInfo? = connectivityManager.activeNetworkInfo
+        return networkInfo?.isConnected ?: false
+    }
+}
+
+/** This returns the userAgent. I wasn't sure if that's what we wanted */
+private fun performNetworkRequests(url: URL, payload: Payload): String {
+    private fun buildRequest(): String {
+        val urlConnection = (HttpURLConnection) url.openConnection()
+        urlConnection.setReadTimeout(5000);
```
ndurell, line 21: These should be constants.

core/network.kt
```diff
@@ -0,0 +1,46 @@
...
+        /** the time to establish the connection with the remote host */
+        urlConnection.setConnectTimeout(5000);
```
ndurell, line 23: I think we want a writeTimeout too...been a while since I've used this api though.

core/network.kt
```diff
@@ -0,0 +1,46 @@
...
+        try {
+            val inputStream = BufferedInputStream(urlConnection.getInputStream())
+            /** This only works for above Android 2.1, from what I could find
+            Alternative is to use a WebView: String userAgent=new WebView(this).getSettings().getUserAgentString();
+             */
+            String userAgent = System.getProperty("http.agent");
```
ndurell, line 29: this seems strange...we should set the user agent to something. We can probably do that in another PR.

core/network.kt
```diff
@@ -0,0 +1,46 @@
+private fun isInternetConnected(context: Context): Boolean {
```
ndurell, line 1: Are you missing imports? Surprised this works?

JenRen-tech: sorry sorry! It's just a draft. I need to clean things up and test

ndurell: Oh my b!

ndurell: you can put it in draft state if you want.

JenRen-tech: ty tho! I'll add in the imports

core/network.kt
JenRen-tech: I was actually curious bout that, since I couldn't find a writeTimeout

core/network.kt
ndurell: I'm also just excited to read some non-python code. Are you working in Android Studio? It can help you resolve these errors.

core/network.kt
ndurell: Totally misremembering...you are right I think these will suffice actually!

core/network.kt
JenRen-tech: :D ikrrr! I'm actually on vscode, but I'll make the switch! Kotlin is a beauty

core/src/main/AndroidManifest.xml
```diff
@@ -1 +1,3 @@
-<manifest package="com.klaviyo.coresdk" />
\ No newline at end of file
+<manifest package="com.klaviyo.coresdk" />
+<uses-permission android:name="android.permission.INTERNET" />
+<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
```
kennyklaviyo, line 3: i think you'll need to put the uses-permission tags inside of manifest:
`<manifest ..>
    <uses-permission... />
</manifest>`

core/network.kt
```diff
@@ -0,0 +1,46 @@
...
+    var result = false
+    val connectivityManager = context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
```
kennyklaviyo, line 3: might be an android kotlin extension library but theres nice kotlin helpers that make this really easy:
`context.getSystemService<ConnectivityManager>()`
also technically this is nullable

core/network.kt
```diff
@@ -0,0 +1,46 @@
+private fun isInternetConnected(context: Context): Boolean {
+    var result = false
```
kennyklaviyo, line 2: i dont think youre using this var

core/network.kt
```diff
@@ -0,0 +1,46 @@
...
+
+/** This returns the userAgent. I wasn't sure if that's what we wanted */
+private fun performNetworkRequests(url: URL, payload: Payload): String {
+    private fun buildRequest(): String {
+        val urlConnection = (HttpURLConnection) url.openConnection()
+        urlConnection.setReadTimeout(5000);
+        /** the time to establish the connection with the remote host */
+        urlConnection.setConnectTimeout(5000);
+        try {
+            val inputStream = BufferedInputStream(urlConnection.getInputStream())
```
kennyklaviyo, line 25: might be a good idea to close this stream as well, kotlin provides a nice way to do this with `use`:
```
inputStream.use {
    ...do stuff with stream
}
```
it will automatically close the stream when completing the block

core/network.kt
```diff
@@ -0,0 +1,46 @@
...
+    buildRequest()
+}
+
+public void track(Payload payload) {
+    performNetworkRequests('klaviyo.com/track', payload)
+}
+
+public void identify(Payload payload) {
+    performNetworkRequests('klaviyo.com/identify', payload)
+}
```
kennyklaviyo, line 46: wait is this java?

jordan-griffin:
Closing this cause I continued the world from another branch myself.

PR #9, opened 2020-08-07
Add gitignore

jordan-griffin:
# Purpose
Adding a gitignore file to the repo for all the generated files we don't care about.

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,24 @@
+# Built application files
+*.apk
+*.aar
+*.aab
+
+# Delvik VM files
+*.dex
+
+Class files for any Java work
+*.class
+
+# Generated files
+.gradle/
+build/
+
+# Local config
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -0,0 +1,24 @@
+# Built application files
+*.apk
+*.aar
+*.aab
+
+# Delvik VM files
+*.dex
```
ndurell, line 7: does dex show up somewhere else?

jordan-griffin: Shows up under `build/intermediates` so I thought I'd separate it out  from the `Built application files` section since it's not set of build files, not a final built application file

PR #12, opened 2020-08-11
Adds a basic configuration builder

jordan-griffin:
# Purpose
We want to provide the user with the easiest means of configuring the SDK before use. One means of doing this which would be relatively clean would be with a Builder pattern constructing a configuration class for us. That configuration class can then be passed into whatever KlaviyoInstance we create to control the application.

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Changelog / Code Overview

- Added a KlaviyoConfig class with an internal builder class
- Added Kotlin dependencies to the Gradle build for the core SDK
- Added basic test cases for the KlaviyoConfig class.

## Related Issues/Tickets

https://klaviyo.tpondemand.com/entity/49008-add-a-basic-configuration-builder

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-    ext.kotlin_version = '1.3.31'
+    ext.kotlin_version = '1.3.72'
     repositories {
         google()
         jcenter()
--- a/build.gradle
+++ b/build.gradle
@@ -1,14 +1,15 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
+    ext.kotlin_version = '1.3.31'
     repositories {
         google()
         jcenter()
-        
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:3.4.2'
-        
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
     }
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -22,12 +22,24 @@ android {
 }
 
 dependencies {
+    /***** Main Dependencies *****/
     implementation fileTree(dir: 'libs', include: ['*.jar'])
+    implementation 'androidx.appcompat:appcompat:1.2.0'
+    //Kotlin Dependencies
     implementation "androidx.core:core-ktx:1.3.1"
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.3.72"
-    implementation 'androidx.appcompat:appcompat:1.2.0'
+
+    /***** Testing Dependencies *****/
     testImplementation 'junit:junit:4.12'
+    testImplementation 'androidx.test:runner:1.2.0'
+    testImplementation 'androidx.test:rules:1.2.0'
+    // Mocking Dependencies - Mockito-inline is needed to navigate around Kotlin final class errors
+    testImplementation 'org.mockito:mockito-inline:3.4.0'
     testImplementation "com.nhaarman.mockitokotlin2:mockito-kotlin:2.2.0"
+    // JSON is bundled with the Android SDK so we need real objects to test against
+    testImplementation "org.json:json:20200518"
+
+    /***** Instrumentation Testing Dependencies *****/
     androidTestImplementation 'androidx.test:runner:1.2.0'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.2.0'
 }
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -24,7 +24,7 @@ android {
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation "androidx.core:core-ktx:1.0.1"
-    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.3.72"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.3.72"
     implementation 'androidx.appcompat:appcompat:1.1.0'
     testImplementation 'junit:junit:4.12'
     androidTestImplementation 'androidx.test:runner:1.2.0'
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -27,6 +27,7 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.3.72"
     implementation 'androidx.appcompat:appcompat:1.2.0'
     testImplementation 'junit:junit:4.12'
+    testImplementation "com.nhaarman.mockitokotlin2:mockito-kotlin:2.2.0"
     androidTestImplementation 'androidx.test:runner:1.2.0'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.2.0'
 }
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -1,4 +1,6 @@
 apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android-extensions'
+apply plugin: 'kotlin-android'
 
 android {
     compileSdkVersion 29
@@ -21,6 +23,8 @@ android {
 
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
+    implementation "androidx.core:core-ktx:1.0.1"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.3.72"
     implementation 'androidx.appcompat:appcompat:1.1.0'
     testImplementation 'junit:junit:4.12'
     androidTestImplementation 'androidx.test:runner:1.2.0'
--- a/core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
+++ b/core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
@@ -7,33 +7,15 @@ class KlaviyoMissingAPIKeyException: Exception("You must declare an API key for 
 
 class KlaviyoMissingContextException: Exception("You must add your application context to the Klaviyo SDK")
 
-class KlaviyoConfig private constructor(
-        apiKey: String,
-        applicationContext: Context?,
-        networkTimeout: Int,
-        networkFlushInterval: Int) {
-    companion object {
+class KlaviyoConfig {
+    internal companion object {
         private const val NETWORK_TIMEOUT_DEFAULT: Int = 500
         private const val NETWORK_FLUSH_INTERVAL_DEFAULT: Int = 60000
-    }
 
-    val apiKey: String
-    val applicationContext: Context?
-    val networkTimeout: Int
-    val networkFlushInterval: Int
-
-    init {
-        if (apiKey.isNullOrEmpty()) {
-            throw KlaviyoMissingAPIKeyException()
-        }
-        if (applicationContext == null) {
-            throw KlaviyoMissingContextException()
-        }
-
-        this.apiKey = apiKey
-        this.applicationContext = applicationContext
-        this.networkTimeout = networkTimeout
-        this.networkFlushInterval = networkFlushInterval
+        var apiKey: String = ""
+        var applicationContext: Context? = null
+        var networkTimeout = NETWORK_TIMEOUT_DEFAULT
+        var networkFlushInterval = NETWORK_FLUSH_INTERVAL_DEFAULT
     }
 
     class Builder {
@@ -67,11 +49,18 @@ class KlaviyoConfig private constructor(
             }
         }
 
-        fun build() = KlaviyoConfig(
-                apiKey,
-                applicationContext,
-                networkTimeout,
-                networkFlushInterval
-        )
+        fun build() {
+            if (apiKey.isEmpty()) {
+                throw KlaviyoMissingAPIKeyException()
+            }
+            if (applicationContext == null) {
+                throw KlaviyoMissingContextException()
+            }
+
+            KlaviyoConfig.apiKey = apiKey
+            KlaviyoConfig.applicationContext = applicationContext
+            KlaviyoConfig.networkTimeout = networkTimeout
+            KlaviyoConfig.networkFlushInterval = networkFlushInterval
+        }
     }
 }
--- a/core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
+++ b/core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
@@ -1,15 +1,39 @@
 package com.klaviyo.coresdk
 
-data class KlaviyoConfig (
-        val apiKey: String,
-        val networkTimeout: Int,
-        val networkFlushInterval: Int
-    ) {
+import java.lang.Exception
+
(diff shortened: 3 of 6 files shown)


core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
```diff
@@ -0,0 +1,34 @@
+package com.klaviyo.coresdk
+
+data class KlaviyoConfig (
```
jordan-griffin, line 3: Originally when speccing out the project I felt this was the way to go for allowing the user to define the configuration of our SDK. Now I'm feeling less certain that this is the case. The alternative is to apply configuration in some supporting files, perhaps XML config, or to create a Kotlin object class (basically a singleton with public vars we can write to) to store this informtion.

jordan-griffin: I thought about turning this into a singleton but it feels like an anti-pattern to be doing that in a builder

core/src/test/java/com/klaviyo/androidsdk/KlaviyoConfigTest.kt
```diff
@@ -0,0 +1,38 @@
+package com.klaviyo.androidsdk
+
+import com.klaviyo.coresdk.KlaviyoConfig
+import org.junit.Test
+
+class KlaviyoConfigTest {
+    @Test
+    fun `Build Configuration works properly!`() {
```
jordan-griffin, line 8: Can use this to specify any string we want as the function name for a unit test in Kotlin. Pretty cool but wonder what people think?

core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
```diff
@@ -0,0 +1,34 @@
...
+        val apiKey: String,
+        val networkTimeout: Int,
+        val networkFlushInterval: Int
+    ) {
+    class Builder {
+        private val NETWORK_TIMEOUT_DEFAULT: Int = 500
+        private val NETWORK_FLUSH_INTERVAL_DEFAULT: Int = 60000
+
+        private lateinit var apiKey: String
```
jordan-griffin, line 12: By making this a `lateinit` var without a default value like the others I'm forcing the builder to have this variable. Any `KlaviyoConfig` that is built out without a supplied `apiKey` will through an `UninitializedProperyAccessException`. Maybe not the perfect way to do this but I thought it was a start to forcing some key pieces of the configuration that we can't provide defaults for

build.gradle
```diff
@@ -1,14 +1,15 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
+    ext.kotlin_version = '1.3.72'
```
kennyklaviyo, line 4: might want to consider using kotlin gradle files, makes dealing with variables so much easier but minor considering how small we expect there gradle files to be

core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
```diff
@@ -0,0 +1,34 @@
...
+        private var networkTimeout: Int = NETWORK_TIMEOUT_DEFAULT
+        private var networkFlushInterval: Int = NETWORK_FLUSH_INTERVAL_DEFAULT
```
kennyklaviyo, line 14: should we validate inputs?

core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
kennyklaviyo: i think this is fine, in terms of getting the apikey, its easier to add a variable to the build config thats backed by an env var and do BuildConfig.KLAVIYO_API_KEY than it is to add a metadata property to the xml

kennyklaviyo: this pattern is fine but i think we need to consider blocking out the KlaviyoConfig constructor. do we want to force devs to use the builder?

core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
kennyklaviyo: hmm, i like the idea and its a super clever way of doing validation but i think we should provide a more specific exception

core/src/test/java/com/klaviyo/androidsdk/KlaviyoConfigTest.kt
kennyklaviyo: love it, much easier to read test names, maybe we can decide on a naming scheme for tests?

JenRen-tech: Seconded on Kenny's comment. Would also want to know

core/build.gradle
```diff
@@ -21,6 +23,8 @@ android {
 
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
+    implementation "androidx.core:core-ktx:1.0.1"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.3.72"
```
kennyklaviyo, line 27: should this be jdk8?

core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
kennyklaviyo: another option could be to require the apiKey in the Builder constructor

kennyklaviyo: maybe all non-defaultable fields can live in the constructor? thoughts?

core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
kennyklaviyo: also i think the data keyword isn't necessary since we dont need the benefits of data like `.copy` and autogenerated comparison functions

build.gradle
jordan-griffin: I'll look into it, it might be worth doing but I haven't used Kotlin gradle files before, always stuck with basic Gradle.

jordan-griffin: But it's probably something I'd convert to in its own PR

core/build.gradle
jordan-griffin: Yes that's a typo, oops!

core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
jordan-griffin: Those are fair points. Since technically this class exists to store data, I blindly attached the data keyword, but there's not going to be any need for comparisons or string generation in this class so its definitely unnecessary overhead.

I do think its best to force the users to use the Builder so I will try to block out the constructor.

core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
jordan-griffin: Had the same thought on adding it to the Builder constructor but I wondered what other non-defaultables we might have in the future, and then destroying the purpose of even having the Builder in the first place if many of the params are just being passed into the constructor. I'm probably overthinking it though/

core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
jordan-griffin: What kind of validation are you thinking? Ensuring numbers are always positive? Throwing exceptions or just reverting to defaults on invalid types?
Given that these values will have healthy defaults in place, I'm thinking of just using the default if the received value is invalid and logging a warning for the user rather than outright throwing an exception.

core/src/test/java/com/klaviyo/androidsdk/KlaviyoConfigTest.kt
jordan-griffin: Ok I'll think about this tomorrow and see if we can decide on a common test naming scheme!

core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
jordan-griffin: Also thanks this gives me a bit more confidence in this approach again

core/src/test/java/com/klaviyo/androidsdk/KlaviyoConfigTest.kt
kennyklaviyo: Nice, are we going to have a readme to document stylistic stuff and other things? or will there be a private wiki for that since this is public?

core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
kennyklaviyo: yeah validation would really just be: are the numbers > 0, i like the idea of defaulting if its not the case and logging a warning to the dev. wondering how we log things, if a dev does use a bad value, how do we feel about this logging getting onto the customer's device (if they ship it anyway with the bad value)?

core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
kennyklaviyo: yeah thats true, agreed not a good pattern to follow if there are many non-defaults, found this article thats interesting https://dev.to/siy/when-builder-is-anti-pattern-3j92

core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
jordan-griffin: Ah that's a good callout, luckily something I already documented in the RFCs but forgot to stroymap a ticket out for so I'll do that soon. But basically we can configure Timber to plant a debug tree so it only logs on debug builds, not release builds.

core/src/test/java/com/klaviyo/androidsdk/KlaviyoConfigTest.kt
jordan-griffin: Still thinking about that. I want to do that but I need to decide on the best place to put these things. I don't really want to use Github Wikis cause even though documenting a topic like this is harmless, there will be things to document like how we configure our locals down the road which we can't put there.

jordan-griffin: I updated tests to follow a standard like this: `Unit Being Worked + State Under Test + Expected Behavior`. Which is pretty standard for unit tests, just a bit more vague to follow with human readable names.
So using these unit tests as an example it breaks down like this:
- **Unit Being Worked**: KlaviyoConfig Builder
- **State Under Test**: missing API key
- **Expected Behaviour**: throws expected exception

core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
```diff
@@ -0,0 +1,58 @@
...
+    val apiKey: String
+    val networkTimeout: Int
+    val networkFlushInterval: Int
+
+    init {
+        if (apiKey.isNullOrEmpty()) {
+            throw KlaviyoMissingAPIKeyException()
+        }
+        if (networkTimeout < 0) {
+            // TODO: When Timber is installed, log warning here
```
kennyklaviyo, line 21: should we default here?

jordan-griffin: Defaults are defined by the builder, so this is just a catch for if the user overrides the default with their own invalid value

kennyklaviyo: right but the building isnt doing any validation besides the api key, so the network values will still be negative

core/build.gradle
```diff
@@ -21,6 +23,8 @@ android {
 
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
+    implementation "androidx.core:core-ktx:1.0.1"
```
kennyklaviyo, line 26: latest core version is 1.3.1
```suggestion
    implementation "androidx.core:core-ktx:1.3.1"
```

core/build.gradle
```diff
@@ -21,6 +23,8 @@ android {
...
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.3.72"
     implementation 'androidx.appcompat:appcompat:1.1.0'
```
kennyklaviyo, line 28:
```suggestion
    implementation 'androidx.appcompat:appcompat:1.2.0'
```

core/build.gradle
jordan-griffin: I had these updated in the following PR to this one but I'll commit these cause why the hell not

core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
jordan-griffin: Oh of course, totally overlooked what you were getting at. Fixing now. And I'll add a test case to verify

core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
```diff
@@ -0,0 +1,66 @@
...
+
+class KlaviyoConfig {
+    internal companion object {
+        private const val NETWORK_TIMEOUT_DEFAULT: Int = 500
+        private const val NETWORK_FLUSH_INTERVAL_DEFAULT: Int = 60000
+
+        lateinit var apiKey: String
+        lateinit var applicationContext: Context
+        var networkTimeout = NETWORK_TIMEOUT_DEFAULT
+        var networkFlushInterval = NETWORK_FLUSH_INTERVAL_DEFAULT
```
kennyklaviyo, line 18: oh so we're going with this approach for config now?

jordan-griffin: Ya, as I worked on following tickets I hit limitations on what I had done before, so I'm turning this into an object class so that it's a singleton and just using the builder as a convenience tool to populating this singleton.
I know builder + singleton can equal code smell. The encapsulation isn't perfect here but its also pretty harmless and for the sake of at least getting some easy-to-configure vars built out fast, I think it's adequate for the time being

jordan-griffin: I think in the future we will probably end up with a `Klaviyo` core class that the user calls `init` on and passes in a built out `KlaviyoConfig` instance, for storing config and kicking off internal threads and the like. But I'm tackling that piece of architecture in a later ticket

core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
```diff
@@ -0,0 +1,81 @@
...
+            if (networkFlushInterval < 0) {
+                // TODO: When Timber is installed, log warning here
+
+            } else {
+                this.networkFlushInterval = networkFlushInterval
+            }
+        }
+
+        fun networkFlushDepth(networkFlushDepth: Int) = apply {
+            if (networkFlushDepth < 0) {
```
kennyklaviyo, line 59:
```suggestion
            if (networkFlushDepth <= 0) {
```

PR #13, opened 2020-08-21
Add basic networking logic to send get and post requests via the SDK

jordan-griffin:
# Purpose
Adds the structure needed to make GET and POST requests via the Android SDK.

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

Adds the structure needed to make GET and POST requests via the Android SDK. These requests and their properties are stored in instances which will later be consumed by the batching queue once implemented. All code is unit tested.

## Changelog / Code Overview

Added multiple classes around networking. All networking objects will inherit from the base `NetworkRequest` class which contains all the necessary logic to make requests and process a basic response. `KlaviyoRequest` is an abstract implementation of this which adds another level of logic specific to our Klaviyo requests, and from there we create `TrackRequest` and `IdentifyRequest` as our two analytics requests. The idea is that this structure is flexible to include more requests in the future, rather than being inflexibility set to just track/identify. It could allow us to add `KlaviyoBatchRequest` later or something

## Test Plan

All main network request classes are unit tested. This contains some pretty heavy mocking to get around all the networking though.

## Related Issues/Tickets

https://klaviyo.tpondemand.com/entity/49005-add-abstract-networking-features-to-the

### DIFF
--- a/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoRequest.kt
+++ b/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoRequest.kt
@@ -5,13 +5,13 @@ abstract class KlaviyoRequest: NetworkRequest() {
         internal const val BASE_URL = "https://a.klaviyo.com"
     }
 
-    override var headerData: String? = null
+    override var queryData: String? = null
     override var payload: String? = null
 
-    internal abstract fun buildKlaviyoJsonHeader(): String
+    internal abstract fun buildKlaviyoJsonQuery(): String
 
     override fun sendNetworkRequest(): String? {
-        headerData = buildKlaviyoJsonHeader()
+        queryData = buildKlaviyoJsonQuery()
         return super.sendNetworkRequest()
     }
 
--- a/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoRequest.kt
+++ b/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoRequest.kt
@@ -15,6 +15,8 @@ abstract class KlaviyoRequest: NetworkRequest() {
         return super.sendNetworkRequest()
     }
 
+    // TODO: Potentially remove this later
+    //  This is a function for testing individual requests made instantly
     fun process(): Boolean {
         val response = sendNetworkRequest()
         return response == "1"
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoRequest.kt
```diff
@@ -0,0 +1,22 @@
+package com.klaviyo.coresdk.networking
+
+abstract class KlaviyoRequest: NetworkRequest() {
+    companion object {
+        internal const val BASE_URL = "https://a.klaviyo.com"
```
jordan-griffin, line 5: I wanna move this into some config value somewhere so that different build flavours can point to different endpoints (like a localhost) but I won't tackle that in this initial PR.

core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoRequest.kt
```diff
@@ -0,0 +1,22 @@
...
+    override var payload: String? = null
+
+    internal abstract fun buildKlaviyoJsonHeader(): String
+
+    override fun sendNetworkRequest(): String? {
+        headerData = buildKlaviyoJsonHeader()
+        return super.sendNetworkRequest()
+    }
+
+    fun process(): Boolean {
```
jordan-griffin, line 18: Function I put in for testing singular network calls for testing purposes (I have a super rudimentary test app I was kicking off requests from and wanted to do some testing before the batching queue is PRed).
I think there's value to keep this here for now

core/src/main/java/com/klaviyo/coresdk/networking/NetworkRequest.kt
```diff
@@ -0,0 +1,95 @@
...
+    internal abstract var requestMethod: RequestMethod
+    internal abstract var headerData: String?
+    internal abstract var payload: String?
+
+    internal fun isInternetConnected(context: Context): Boolean {
+        val connectivityManager = context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
+        // TODO: We may want to replace this later. Deprecated as of Android API 29.
+        //  But the alternative solution is an asynchronous task that requires the user to register
+        //  a network callback listener, which isn't ideal just for a network connectivity check in an SDK
+        val networkInfo = connectivityManager.activeNetworkInfo ?: return false
```
jordan-griffin, line 26: As noted in the comment above, this `NetworkInfo` class that we are fetching here is deprecated as of Android 29 sadly. And the alternative workaround is not as easy from an SDK-development standpoint. I vote we leave this in as a connectivity check for now but cut a ticket to consider moving to a network listener approach in the future.

PR #14, opened 2020-09-02
Add rudimentary network batching thread

jordan-griffin:
# Purpose
Adds a very basic network batcher which will use the existing information in the `KlaviyoConfig` class to determine the triggers for emptying the queue. This does not hit any special batch endpoints in Klaviyo yet, instead it just sends off all network requests in the queue one by one so that we only activate the Android network radio once and conserve on the device's battery power.

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

Adds a network batcher and accompanying tests. This batcher creates a HandlerThread which executes a runnable for monitoring the batch queue. The batch queue is a ConcurrentLinkedQueue of network requests for thread safety. The threading employed here is really basic stuff so we shouldn't need to worry much about synchronization, we just kick it off whenever the batch queue is populated and end it whenever the queue is empty.
This also includes a `KlaviyoLifecycleCallbackListener` class which can be implemented by the user's Application class to monitor the lifecycle of their application so that we can empty out the batch queue before they terminate the application.

## Changelog / Code Overview

- Added `KlaviyoLifecycleCallbackListener` to monitor the lifecycle of the parent application and empty out our batch queue when the application is stopping.
- Added `NetworkBatcher` to handle batching of our analytics calls
- Added `NetworkBatcherTest` to unit test the batch queue

## Test Plan

Unit tested and run against the test application to verify that the queue triggers on both timeouts and when reaching max capacity. Also tested the custom lifecycle callbacks in the test application to verify that the batch queue will be emptied when the application is stoped but before the process is terminated.

## What to look for

Make sure there won't be an synchronization issues here. How would we handle the case where the batching thread is prematurely terminated. We'd lose those customer metrics, can we back them up somewhere? Use a different threading API?

## Related Issues/Tickets

https://klaviyo.tpondemand.com/entity/49095-create-basic-batching-queue

### DIFF
--- a/core/src/main/java/com/klaviyo/coresdk/KlaviyoLifecycleCallbackListener.kt
+++ b/core/src/main/java/com/klaviyo/coresdk/KlaviyoLifecycleCallbackListener.kt
@@ -1,0 +1,37 @@
+package com.klaviyo.coresdk
+
+import android.app.Activity
+import android.app.Application
+import android.os.Bundle
+import com.klaviyo.coresdk.networking.NetworkBatcher
+
+class KlaviyoLifecycleCallbackListener: Application.ActivityLifecycleCallbacks {
+    private var activitiesActive = 0
+
+    override fun onActivityPaused(p0: Activity) {
+    }
+
+    override fun onActivityStarted(p0: Activity) {
+    }
+
+    override fun onActivityDestroyed(p0: Activity) {
+    }
+
+    override fun onActivitySaveInstanceState(p0: Activity, p1: Bundle) {
+    }
+
+    override fun onActivityStopped(p0: Activity) {
+        activitiesActive--
+        if (activitiesActive == 0) {
+            NetworkBatcher.forceEmptyQueue()
+        }
+    }
+
+    override fun onActivityCreated(p0: Activity, p1: Bundle?) {
+        activitiesActive++
+    }
+
+    override fun onActivityResumed(p0: Activity) {
+    }
+
+}
--- a/core/src/main/java/com/klaviyo/coresdk/KlaviyoLifecycleCallbackListener.kt
+++ b/core/src/main/java/com/klaviyo/coresdk/KlaviyoLifecycleCallbackListener.kt
@@ -8,30 +8,29 @@ import com.klaviyo.coresdk.networking.NetworkBatcher
 class KlaviyoLifecycleCallbackListener: Application.ActivityLifecycleCallbacks {
     private var activitiesActive = 0
 
-    override fun onActivityPaused(p0: Activity) {
+    override fun onActivityCreated(activity: Activity, bundle: Bundle?) {
     }
 
-    override fun onActivityStarted(p0: Activity) {
+    override fun onActivityStarted(activity: Activity) {
+        activitiesActive++
     }
 
-    override fun onActivityDestroyed(p0: Activity) {
+    override fun onActivityResumed(activity: Activity) {
     }
 
-    override fun onActivitySaveInstanceState(p0: Activity, p1: Bundle) {
+    override fun onActivitySaveInstanceState(activity: Activity, bundle: Bundle) {
     }
 
-    override fun onActivityStopped(p0: Activity) {
+    override fun onActivityPaused(activity: Activity) {
+    }
+
+    override fun onActivityStopped(activity: Activity) {
         activitiesActive--
         if (activitiesActive == 0) {
             NetworkBatcher.forceEmptyQueue()
         }
     }
 
-    override fun onActivityCreated(p0: Activity, p1: Bundle?) {
-        activitiesActive++
+    override fun onActivityDestroyed(activity: Activity) {
     }
-
-    override fun onActivityResumed(p0: Activity) {
-    }
-
 }
--- a/core/src/main/java/com/klaviyo/coresdk/networking/NetworkBatcher.kt
+++ b/core/src/main/java/com/klaviyo/coresdk/networking/NetworkBatcher.kt
@@ -32,7 +32,6 @@ object NetworkBatcher {
         if (batchQueue.isEmpty()) {
             initBatcher()
             queueInitTime = System.currentTimeMillis()
-
         }
 
         for (request in requests) {
--- a/core/src/main/java/com/klaviyo/coresdk/networking/NetworkBatcher.kt
+++ b/core/src/main/java/com/klaviyo/coresdk/networking/NetworkBatcher.kt
@@ -1,0 +1,67 @@
+package com.klaviyo.coresdk.networking
+
+import android.os.Handler
+import android.os.HandlerThread
+import com.klaviyo.coresdk.KlaviyoConfig
+import java.util.concurrent.ConcurrentLinkedQueue
+
+object NetworkBatcher {
+    private val handlerThread = HandlerThread("Klaviyo Network Batcher")
+    private var handler: Handler? = null
+    private var batchQueue = ConcurrentLinkedQueue<NetworkRequest>()
+    private var queueInitTime = 0L
+
+    internal fun initBatcher() {
+        if (!handlerThread.isAlive) {
+            handlerThread.start()
+            handler = Handler(handlerThread.looper)
+        }
+        handler?.post(NetworkRunnable())
+    }
+
+    internal fun forceEmptyQueue() {
+        handler?.removeCallbacksAndMessages(null)
+        handler?.post(NetworkRunnable(true))
+    }
+
+    internal fun getBatchQueueSize(): Int {
+        return batchQueue.size
+    }
+
+    internal fun batchRequests(vararg requests: NetworkRequest) {
+        if (batchQueue.size == 0) {
+            initBatcher()
+            queueInitTime = System.currentTimeMillis()
+
+        }
+
+        for (request in requests) {
+            if (request is TrackRequest) {
+                request.generateUnixTimestamp()
+            }
+            batchQueue.add(request)
+        }
+    }
+
+    internal class NetworkRunnable(private val forceEmpty: Boolean = false): Runnable {
+        override fun run() {
+            val emptied = emptyRequestQueue(forceEmpty)
+            if (!emptied) {
+                handler?.postDelayed(this, 2000)
+            }
+        }
+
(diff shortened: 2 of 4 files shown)


core/src/main/java/com/klaviyo/coresdk/KlaviyoLifecycleCallbackListener.kt
```diff
@@ -0,0 +1,37 @@
+package com.klaviyo.coresdk
+
+import android.app.Activity
+import android.app.Application
+import android.os.Bundle
+import com.klaviyo.coresdk.networking.NetworkBatcher
+
+class KlaviyoLifecycleCallbackListener: Application.ActivityLifecycleCallbacks {
```
jordan-griffin, line 8: This is not a necessary part of the application but outline the benefits of having a custom application lifecycle callback implemented by the parent application for us in monitoring user sessions: https://docs.google.com/document/d/10fseikYPosBoYg0daQqr6zGyfeLrn-DiIWUSdzvCmCw/edit#heading=h.tenapcbrmzfp

In this instance I am using the callbacks to empty out the batch thread before the application gets terminated. Threads will die when their parent app dies and we don't want to lose requests that may have been previously batched up.

jordan-griffin: Why not just put this in a threaded background service? Cause services need to be registered in the manifest of the parent application. It is totally possible to make that a requirement of user setup but I think it would a question for later improvements.

jordan-griffin: Another alternative is to use the WorkManager API to sync up our requests on the OS-level for some system-wide batching. Bonus: It can double as a push scheduler with Firebase Cloud Messaging too. This is an external dependency though so we'd like to avoid its inclusion if possible.
Detailed here: https://docs.google.com/document/d/1Hft90xwKJITFajUsLcXnhVMjfNXLXhbbhjzs9TwvUwI/edit#heading=h.tenapcbrmzfp

core/src/main/java/com/klaviyo/coresdk/KlaviyoLifecycleCallbackListener.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    override fun onActivityStopped(p0: Activity) {
+        activitiesActive--
+        if (activitiesActive == 0) {
+            NetworkBatcher.forceEmptyQueue()
+        }
+    }
+
+    override fun onActivityCreated(p0: Activity, p1: Bundle?) {
+        activitiesActive++
```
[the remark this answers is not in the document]
kennyklaviyo: might also want to rearrange the methods so its in the order of execution just so its easier to read but small nit

core/src/main/java/com/klaviyo/coresdk/networking/NetworkBatcher.kt
```diff
@@ -0,0 +1,67 @@
...
+            if (!emptied) {
+                handler?.postDelayed(this, 2000)
+            }
+        }
+
+        private fun emptyRequestQueue(forceEmpty: Boolean = false): Boolean {
+            val queueTimePassed = System.currentTimeMillis() - queueInitTime
+            val readyToEmpty = batchQueue.size >= KlaviyoConfig.networkFlushDepth || queueTimePassed >= KlaviyoConfig.networkFlushInterval
+            if (forceEmpty || readyToEmpty) {
+                while(batchQueue.size > 0) {
```
kennyklaviyo, line 58: is there a method on the queue like `batchQueue.isNotEmpty()`?

kennyklaviyo: also, looking at the source code `ConcurrentLinkedQueue` the size method is not constant time since its backed by a linked list. maybe its a little over the top but might be easier to use `.poll()` instead like:
```
var request: NetworkRequest? = null
while (batchQueue.poll().also { request = it } != null) {
    request?.sendNetworkRequest()
}
```

kennyklaviyo: this is because `.poll` doesnt throw an exception when the queue is empty like `.remove` does

core/src/main/java/com/klaviyo/coresdk/networking/NetworkBatcher.kt
```diff
@@ -0,0 +1,67 @@
...
+            initBatcher()
+            queueInitTime = System.currentTimeMillis()
+
+        }
+
+        for (request in requests) {
+            if (request is TrackRequest) {
+                request.generateUnixTimestamp()
+            }
+            batchQueue.add(request)
```
kennyklaviyo, line 42: if this is a queue, should be be using the queue interface over this? ie use `.offer` instead of `.add`, i think the source code for `ConcurrentLinkedQueue` also suggests using offer instead of add

core/src/main/java/com/klaviyo/coresdk/networking/NetworkBatcher.kt
```diff
@@ -0,0 +1,67 @@
...
+        handler?.removeCallbacksAndMessages(null)
+        handler?.post(NetworkRunnable(true))
+    }
+
+    internal fun getBatchQueueSize(): Int {
+        return batchQueue.size
+    }
+
+    internal fun batchRequests(vararg requests: NetworkRequest) {
+        if (batchQueue.size == 0) {
```
kennyklaviyo, line 32: you should use `.isEmpty()` since `.size` is an O(n) operation

core/src/main/java/com/klaviyo/coresdk/networking/NetworkBatcher.kt
```diff
@@ -0,0 +1,67 @@
...
+        handler?.post(NetworkRunnable())
+    }
+
+    internal fun forceEmptyQueue() {
+        handler?.removeCallbacksAndMessages(null)
+        handler?.post(NetworkRunnable(true))
+    }
+
+    internal fun getBatchQueueSize(): Int {
+        return batchQueue.size
```
kennyklaviyo, line 28: might want to be careful using this here too, maybe just worth noting in a comment that this is O(n)

core/src/test/java/com/klaviyo/coresdk/networking/NetworkBatcherTest.kt
```diff
@@ -0,0 +1,72 @@
...
+                .networkFlushInterval(1000)
+                .networkFlushDepth(10)
+                .build()
+    }
+
+    @Test
+    fun `Network Batcher empties the queue when full`() {
+        val batcherSpy = spy<NetworkBatcher>()
+
+        doNothing().`when`(batcherSpy).initBatcher()
```
kennyklaviyo, line 27: can we use `.whenever` instead of `.'when'`?

kennyklaviyo (changes requested):

core/src/main/java/com/klaviyo/coresdk/KlaviyoLifecycleCallbackListener.kt
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    override fun onActivityDestroyed(p0: Activity) {
+    }
+
+    override fun onActivitySaveInstanceState(p0: Activity, p1: Bundle) {
+    }
+
+    override fun onActivityStopped(p0: Activity) {
+        activitiesActive--
```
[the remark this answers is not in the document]
jordan-griffin: Pretty sure everything remains on the UI thread for these calls

core/src/main/java/com/klaviyo/coresdk/KlaviyoLifecycleCallbackListener.kt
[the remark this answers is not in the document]
jordan-griffin: Oh nice find ya, originally i had it in Created <-> Destroyed but realised that the app will terminate before asynchronous threads can finish, so I moved it to stopped and didn't change created

core/src/main/java/com/klaviyo/coresdk/networking/NetworkBatcher.kt
jordan-griffin: I added this cause I wanted to expose the batch queue size easily for testing. I don't see an issue in having this exposed as is, even if its an O(n) operation

core/src/main/java/com/klaviyo/coresdk/networking/NetworkBatcher.kt
jordan-griffin: I'll update sure cause I think its harmless to do so. The difference is super minimal though to my understanding. `.offer` will give us some exception handling for the cases where we can't add an element to the queue, but since our queue is unbounded (we have a threshold for checking if we should empty the queue of requests but not a bounded size constraint cause we still want to accept any requests in the interim) it's not actually something that should realistically happen to us.

core/src/main/java/com/klaviyo/coresdk/networking/NetworkBatcher.kt
jordan-griffin: Fair point! I think this is a nice optimization!

core/src/test/java/com/klaviyo/coresdk/networking/NetworkBatcherTest.kt
jordan-griffin: Good point, much clearner to keep using `whenever`

core/src/main/java/com/klaviyo/coresdk/networking/NetworkBatcher.kt
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    internal fun batchRequests(vararg requests: NetworkRequest) {
+        if (batchQueue.isEmpty()) {
+            initBatcher()
+            queueInitTime = System.currentTimeMillis()
+
```
kennyklaviyo, line 35:
```suggestion
```

kennyklaviyo (approved):

PR #15, opened 2020-09-04
Improve analytics API

jordan-griffin:
# Purpose
Makes some general improves to hide more of the implementation and design a more straightforward API for the end user.

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

Changes the API by adding in a `KlaviyoNetworking` class which has a bunch of top-level functions for the user to take advantage of. The internals still work the same, building up a request object, but this object just has most of its implementation details hidden from the user instead of the original design where the object was exposed as part of the API.
Other general improvements were made.

## Changelog / Code Overview

- Added `KlaviyoNetworking` as our exposed API class for networking calls
- Internalized the `TrackRequest` and `IdentifyRequest` classes (as well as the classes they inherit from
- Added some new values to the `KlaviyoConfig` class
- Updated tests to match changes
- Moved network request files to subdirectory for cleanup.


core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
```diff
@@ -12,7 +11,8 @@ object KlaviyoConfig {
     private const val NETWORK_TIMEOUT_DEFAULT: Int = 500
     private const val NETWORK_FLUSH_INTERVAL_DEFAULT: Int = 60000
     private const val NETWORK_FLUSH_DEPTH_DEFAULT: Int = 20
-    private const val NETWORK_FLUSH_CHECK_INTERVAL: Long = 2000
+    private const val NETWORK_FLUSH_CHECK_INTERVAL: Int = 2000
+    private const val NETWORK_USE_ANALYTICS_BATCH_QUEUE: Boolean = true
```
jordan-griffin, line 15: Adding activation of the batch queue to the configuration. Will be useful for early day testing

core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoEvent.kt
```diff
@@ -0,0 +1,7 @@
+package com.klaviyo.coresdk.networking
+
+sealed class KlaviyoEvent(val name: String) {
```
jordan-griffin, line 3: Note to self: Fill this out with more Klaviyo events

core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoNetworking.kt
```diff
@@ -0,0 +1,26 @@
+package com.klaviyo.coresdk.networking
+
+import com.klaviyo.coresdk.KlaviyoConfig
+import com.klaviyo.coresdk.networking.requests.IdentifyRequest
+import com.klaviyo.coresdk.networking.requests.KlaviyoRequest
+import com.klaviyo.coresdk.networking.requests.TrackRequest
+
+fun track(event: KlaviyoEvent, customerProperties: Map<String, String>, properties: Map<String, String>? = null) {
```
jordan-griffin, line 8: New top-level functions that help hide more of the SDK's implementation

core/src/main/java/com/klaviyo/coresdk/networking/requests/IdentifyRequest.kt
```diff
@@ -1,13 +1,14 @@
-package com.klaviyo.coresdk.networking
+package com.klaviyo.coresdk.networking.requests
 
 import com.klaviyo.coresdk.KlaviyoConfig
+import com.klaviyo.coresdk.networking.RequestMethod
 import org.json.JSONObject
 
-class IdentifyRequest (
+internal class IdentifyRequest (
```
jordan-griffin, line 7: There will be a few instances of this internal tag in this PR. Just internalizing some areas that I missed before to hide more of the SDK's implementation from the user.

core/src/main/java/com/klaviyo/coresdk/networking/requests/NetworkRequest.kt
```diff
@@ -1,18 +1,19 @@
-package com.klaviyo.coresdk.networking
+package com.klaviyo.coresdk.networking.requests
```
jordan-griffin, line 1: I moved all request types and abstract classes into a sub-folder since the networking code was expanding more and more. Hence the imports updated

core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoNetworking.kt
kennyklaviyo: i think it would be nicer to wrap this in a class of some sort like:
```
object KlaviyoEvents {
    fun track(...
}
```
this way, developers implementing will do KlaviyoEvents.track(...), of course, not sure how this would work in a Java based project...

kennyklaviyo (approved):

core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoEvent.kt
chrisconlon-klaviyo: I'm a little curious whether adding non-custom events here could be confusing, for example say the customer had a shopify store integrated which tracks a "Checkout Started" event, and then they start tracking one here (which would show up as a custom event). I just want to make sure that product is on board with potentially confusing scenarios like that.

core/src/main/java/com/klaviyo/coresdk/networking/NetworkBatcher.kt
```diff
@@ -46,7 +48,7 @@ object NetworkBatcher {
         override fun run() {
             val emptied = emptyRequestQueue(forceEmpty)
             if (!emptied) {
-                handler?.postDelayed(this, KlaviyoConfig.networkFlushCheckInterval)
+                handler?.postDelayed(this, KlaviyoConfig.networkFlushCheckInterval.toLong())
```
chrisconlon-klaviyo, line 51: This was changed to an `int` above, should we have `toInt()` here? Caveat, I have no idea what I'm talking about since I've never worked with kotlin, or much with java either.

core/src/main/java/com/klaviyo/coresdk/networking/requests/KlaviyoRequest.kt
```diff
@@ -1,6 +1,8 @@
-package com.klaviyo.coresdk.networking
+package com.klaviyo.coresdk.networking.requests
 
-abstract class KlaviyoRequest: NetworkRequest() {
+import com.klaviyo.coresdk.networking.NetworkBatcher
+
+internal abstract class KlaviyoRequest: NetworkRequest() {
     companion object {
         internal const val BASE_URL = "https://a.klaviyo.com"
```
chrisconlon-klaviyo, line 7: Not sure if it matters, but is this something we could make configurable? I could see that being very useful for testing, like if we wanted to stick an ngrok url in here.

chrisconlon-klaviyo (approved):
Looks solid!

noidea

core/src/main/java/com/klaviyo/coresdk/networking/NetworkBatcher.kt
kennyklaviyo: `.toInt()` wouldnt be needed since networkFlushCheckInterval is already an `Int` type but this doesnt matter since the method signature for `postDelayed` has this second argument as a `Long` thus the need to cast the `Int` with `.toLong()`

core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoNetworking.kt
jordan-griffin: I kinda thought the opposite would be better. I was fond of the idea of having top-level function calls for track and identify as it's a cleaner API.
I spose the drawback is that the naming is generic enough that there could be overlap with other dependencies leading to some developer confusion.

jordan-griffin: Also calling Kotlin object methods in Java is just the same as you would in Kotlin. top-level functions would need to invoke the Kotlin file of origin to work through Java, and cleaning that up to be more Java-esque would have to be included in the Java wrapper for this SDK

core/src/main/java/com/klaviyo/coresdk/networking/requests/KlaviyoRequest.kt
jordan-griffin: Absolutely yes. I  was planning specify local URLs in the build variables and then inject those into the manifest file originally, but I think there were issues with doing this sort of manifest work with a lib.
There are multiple ways of achieving this though, we could also utilize BuildConfig, or write to a properties file on the system. It's definitely planned anyway

PR #17, opened 2020-09-08
Add basic anonymous IDs to the SDK

jordan-griffin:
# Purpose
Adds basic anonymous IDs to the Android SDK.

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

This is a parallel of the anonymous ID logic performed in the existing iOS SDK but altered to suit the Android ecosystem. It attempts to read UUIDs from app files at runtime, generating them if they are not found. We use file storage so that the UUIDs will not change unless the parent app is uninstalled or its app data is cleared. 

## Changelog / Code Overview

- Added a `ConfigFileUtils` class for reading the new config file which stores static values for the SDK (currently just the UUID). 
- Updated the track and identify requests to attach UUIDs as an anonymous ID in the relevant part of the query with each request
- Added unit tests

## Test Plan

Tested locally using the provided unit tests and using the test application to send requests with logs to print state.

## Related Issues/Tickets

https://klaviyo.tpondemand.com/entity/49006-implement-basic-anonymous-ids


core/src/main/java/com/klaviyo/coresdk/ConfigFileUtils.kt
```diff
@@ -0,0 +1,45 @@
...
+import java.io.File
+import java.util.*
+
+internal object ConfigFileUtils {
+    private const val CONFIG_FILE_NAME = "klaviyo-sdk-config"
+
+    private const val UUID_KEY = "uuid"
+
+    fun readOrCreateUUID(): String {
+        val file = File("${KlaviyoConfig.applicationContext.filesDir}/$CONFIG_FILE_NAME")
```
jordan-griffin, line 12: I went with internal storage for the UUID since it is then stored alongside the parent application. When the parent application is uninstalled, or when app data is cleared, this file will be deleted. This gives us pretty much consistent anonymous IDs for each install of the application, which works as a basic initial solution to our problem.

jordan-griffin: I went for internal storage over shared preferences since its more private to the parent application and I'm not particularly fond of the idea of this just being publicly accessible and changeable information. With internal storage, it is at least protected from access via other apps.
To avoid this logic, the alternative is to utilize Firebase IIDs, which is a lib that will make generating the UUIDs pretty straightforward for this use case. Since we plan to change how we identify anonymous IDs in the future I decided not to utilize this dependency and instead do a manual solution that is more generic.

core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -6,13 +6,13 @@ import com.klaviyo.coresdk.networking.requests.KlaviyoRequest
 import com.klaviyo.coresdk.networking.requests.TrackRequest
 
 object Klaviyo {
-    fun track(event: KlaviyoEvent, customerProperties: Map<String, String>, properties: Map<String, String>? = null) {
+    fun track(event: KlaviyoEvent, customerProperties: MutableMap<String, String>, properties: Map<String, String>? = null) {
```
jordan-griffin, line 9: Needs to be a `MutableMap` now to support the anonymous IDs that we will tack on

core/src/main/java/com/klaviyo/coresdk/networking/requests/KlaviyoRequest.kt
```diff
@@ -5,14 +5,19 @@ import com.klaviyo.coresdk.networking.NetworkBatcher
...
     companion object {
         internal const val BASE_URL = "https://a.klaviyo.com"
+
+        internal const val ANON_KEY = "\$anonymous"
     }
 
     override var queryData: String? = null
     override var payload: String? = null
 
+    internal abstract fun addAnonymousIdToProps()
```
jordan-griffin, line 15: Track and identify requests insert anonymous IDs into different property maps, so I abstracted this function for each

core/src/main/java/com/klaviyo/coresdk/networking/NetworkBatcher.kt
```diff
@@ -37,9 +37,6 @@ object NetworkBatcher {
         }
 
         for (request in requests) {
-            if (request is TrackRequest) {
```
jordan-griffin, line 40: Unrelated to the rest of the PR, but just noticed that this code here was totally not necessary and this timestamp generation was moved to an earlier point in a previous PR.

core/src/main/java/com/klaviyo/coresdk/ConfigFileUtils.kt
```diff
@@ -0,0 +1,45 @@
+package com.klaviyo.coresdk
+
+import java.io.File
+import java.util.*
```
ndurell, line 4: Do you want to import star here?

core/src/main/java/com/klaviyo/coresdk/ConfigFileUtils.kt
ndurell: Nice sounds like a solid choice!

core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
ndurell, line 9: That feels weird assuming this is a public api. It's ok if you have to copy the map to new one to modify it.

core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -6,13 +6,13 @@ import com.klaviyo.coresdk.networking.requests.KlaviyoRequest
...
         val request = TrackRequest(event.name, customerProperties, properties)
         request.generateUnixTimestamp()
         processRequest(request)
     }
 
-    fun identify(properties: Map<String, String>) {
+    fun identify(properties: MutableMap<String, String>) {
```
ndurell, line 15: Same here nix the MutableMap.

core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
kennyklaviyo: +1

core/src/main/java/com/klaviyo/coresdk/networking/requests/KlaviyoRequest.kt
kennyklaviyo: you can sorta flip this dependence, this class knows the key and the value for anonymous is always the same no matter the type of request, so you can do
```
internal fun addAnonymous(map: Map<String, String>): Map<String, String> {
    return map.plus(ANON_KEY to "Android:${ConfigFileUtils.readOrCreateUUID()}")
}
```

and in the `buildKlaviyoJsonQuery` use this to modify it

jordan-griffin: I had a similar idea but didn't do that cause I didn't want to bloat the single-purpose of `buildKlaviyoJsonQuery`. I wanted that function to just take the existing data and simply convert it to a `JSONObject` and not worry about what other data gets filled in later.
This is nicer for adding requests in the future though. And its not a massive refactor for me to come back later and redefine some of this logic if I feel I'm bloating an internal function or two

jordan-griffin: Also just a very minor point, I didn't use `plus` to add to maps cause it rebuilds the entire map each time and I felt it may be more effective to use `put` or `set` in a mutable map instead.

core/src/main/java/com/klaviyo/coresdk/networking/requests/KlaviyoRequest.kt
```diff
@@ -1,15 +1,22 @@
...
         internal const val BASE_URL = "https://a.klaviyo.com"
+
+        internal const val ANON_KEY = "\$anonymous"
     }
 
     override var queryData: String? = null
     override var payload: String? = null
 
+    internal fun addAnonymousIdToProps(map: MutableMap<String, String>) {
+        map[ANON_KEY] = "Android:${ConfigFileUtils.readOrCreateUUID()}"
```
chrisconlon-klaviyo, line 17: QUESTION: Is there a way to namespace the context this is created within? My concern is that for apps that want to allow users to login/logout that they would appear as the same anonymous ID until we discover their phone number or email. Its not critical to address now, but we should probably at least consider it as an option now.

chrisconlon-klaviyo (approved):
One question, looks good though!

core/src/main/java/com/klaviyo/coresdk/networking/requests/KlaviyoRequest.kt
jordan-griffin: So, when I RFC'd my proposed solution for this issue there was some contentious aspects to it that we couldn't resolve there and then. What we decided to do in the meantime was mimic what the iOS SDK does for anonymous IDs for the early stages (which is what brings us to this drawback with logins/logouts) and to revisit the anonymous IDs later

jordan-griffin: And just to keep you in the loop on the details, my proposed solution was to use lifecycle listeners to generate new anonymous IDs for each "user session" (I defined user sessions as ending after 30-60 seconds of being away from the parent application).
The worry was the number of anonymous profiles I could end up generating in Klaviyo if users had applications that don't collect much user info.

chrisconlon-klaviyo: Ah thats right, I forgot about that. Carry on!

core/src/main/java/com/klaviyo/coresdk/ConfigFileUtils.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    private const val UUID_KEY = "uuid"
+
+    fun readOrCreateUUID(): String {
+        val file = File("${KlaviyoConfig.applicationContext.filesDir}/$CONFIG_FILE_NAME")
+
+        var uuid = findConfigValue(file, UUID_KEY)
+
+        if (uuid.isNullOrEmpty()) {
+            uuid = UUID.randomUUID().toString()
```
ndurell, line 17: This'll be unique across devices yah?

ndurell: Also toString gives you right value? Did you verify what it looks like?

core/src/main/java/com/klaviyo/coresdk/ConfigFileUtils.kt
```diff
@@ -0,0 +1,45 @@
...
+                if (line.startsWith("$key:")) {
+                    value = line.substring("$key:".length)
+
+                }
+            }
+        }
+        return value
+    }
+
+    fun writeConfigValue(file: File, key:String, value: String) {
```
ndurell, line 42: should this be private?

core/src/main/java/com/klaviyo/coresdk/ConfigFileUtils.kt
```diff
@@ -0,0 +1,45 @@
...
+
+        if (uuid.isNullOrEmpty()) {
+            uuid = UUID.randomUUID().toString()
+            writeConfigValue(file, UUID_KEY, uuid)
+        }
+
+        return uuid
+    }
+
+    fun findConfigValue(file: File, key: String): String? {
```
ndurell, line 24: same private?

core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
```diff
@@ -46,15 +46,15 @@ object KlaviyoConfig {
             this.applicationContext = context
         }
 
-        fun networkTimeout(networkTimeout: Int)  = apply {
+        fun networkTimeout(networkTimeout: Int) = apply {
             if (networkTimeout < 0) {
                 // TODO: When Timber is installed, log warning here
```
ndurell, line 51: Can we use Timber if the app might use it?

core/src/main/java/com/klaviyo/coresdk/networking/requests/IdentifyRequest.kt
```diff
@@ -28,10 +28,11 @@ internal class IdentifyRequest (
         }
      */
     override fun buildKlaviyoJsonQuery(): String {
+        addAnonymousIdToProps(properties)
```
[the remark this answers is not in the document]
ndurell: For example you don't want this called on the main thread.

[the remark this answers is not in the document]
ndurell: Also is this code thread safe? Does it need to be synchronized?

core/src/main/java/com/klaviyo/coresdk/networking/requests/TrackRequest.kt
```diff
@@ -32,11 +32,12 @@ internal class TrackRequest (
         }
      */
     override fun buildKlaviyoJsonQuery(): String {
+        addAnonymousIdToProps(customerProperties)
         val json = JSONObject(
             mapOf(
                 "token" to KlaviyoConfig.apiKey,
                 "event" to event,
-                "customer_properties" to JSONObject(customerProperties),
+                "customer_properties" to JSONObject(customerProperties.toMap()),
```
ndurell, line 40: Always wondered this: Do these properties get set on the customers why do we have two different sets of properties? I'm not necessarily expecting an answer to this just was curious.

core/src/main/java/com/klaviyo/coresdk/ConfigFileUtils.kt
jordan-griffin: This is one of Google's recommended ways to generate and store instance IDs in Android. They should be globally unique

https://developer.android.com/training/articles/user-data-ids.html#instance-ids-guids

Also this article on alternative approaches:
https://medium.com/@ssaurel/how-to-retrieve-an-unique-id-to-identify-android-devices-6f99fd5369eb

jordan-griffin: Also if we were to use device IDs to further identify our instance IDs, then we need to include more runtime permissions so I'd rather now. 
UUIDs are considered unique across space and time according to Google's loooong RFC on the topic: https://www.ietf.org/rfc/rfc4122.txt

core/src/main/java/com/klaviyo/coresdk/ConfigFileUtils.kt
jordan-griffin: The encompassing object is internal so there's no worry about it showing up as a function in the API. Left the function calls like this for any future file writing that the SDK needs. Acting as a utility class here

jordan-griffin: Could totally privatize too though. Just makes testing harder

core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
jordan-griffin: Ya if we make Timber a debug dependency only it won't be included in release builds

jordan-griffin: Pretty sure Timber has its own utilities to further ensure logs are only written on debug trees too

ndurell: Ok sounds good.

core/src/main/java/com/klaviyo/coresdk/networking/requests/IdentifyRequest.kt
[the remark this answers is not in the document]
jordan-griffin: True true. With the current setup this is only called on the batcher thread, so its already off the main thread. For the current setup this is totally safe, but if we expand the threading solution in the future then it would be good to have this synchronized

core/src/main/java/com/klaviyo/coresdk/networking/requests/TrackRequest.kt
jordan-griffin: When it comes to track requests, `properties` is where you define properties about the event you are tracking, and `customer_properties` is where you define the properties that identity who triggered said event.

But in identify requests `properties` is where you define the properties about the person you are identifying, so the Klaviyo API is kind of misleading to work with

jordan-griffin: Maybe that answers your question a bit?

ndurell (approved):

PR #20, opened 2020-09-09
Add modularized builds to the Android SDK and introduce the push module

jordan-griffin:
# Purpose
Improves the Gradle build scripts so that they better support a modularized build system

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

The redone Gradle scripts are now modularized for building of individual modules. The SDK modules are now stored in a sub-directory called `sdk` which will contain the existing `core` and new `push` modules. This is so that future work (to add an `example app` or `java wrappers` can be split off into their own sub-directories without interfering with the builds for the SDK modules.

## Changelog / Code Overview

- Added a sdk directory with its own build scripts which will branch off into our individual SDK modules
- Added the push SDK module as a placeholder for when we start adding push notifications and to demonstrate the modular builds
- Updated all build scripts from top level to the bottom level to support modularization
Note: Most of the changes in this PR are simply files being moved to their new location.

## Related Issues/Tickets

https://klaviyo.tpondemand.com/entity/49009-add-basic-gradle-scripts-for-modularized


build.gradle
```diff
@@ -1,13 +1,15 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
+    apply from: 'versions.gradle'
```
jordan-griffin, line 4: For consistency across modules I will be placing build and dependency versions in this new `versions.gradle` file and applying it to the build process at the top level

sdk/build.gradle
```diff
@@ -0,0 +1,48 @@
+subprojects {
```
jordan-griffin, line 1: This is the SDK build script. It makes a `subprojects` call so that all the contained build logic is applied to all its subprojects. Any build logic common to all SDK modules should be declared here.
Nothing is applied on this project level since this level is not designed to contain anything but the individual SDK modules that it builds out

sdk/build.gradle
```diff
@@ -0,0 +1,48 @@
...
+    apply plugin: 'com.android.library'
+    apply plugin: 'kotlin-android'
+    apply plugin: 'kotlin-android-extensions'
+
+    def ext = rootProject.ext
```
jordan-griffin, line 6: Allows us to fetch our dependency and build versions from the `versions.gradle` we declared at the top level build script. We call the extension of the root project to get this information

sdk/build.gradle
```diff
@@ -0,0 +1,48 @@
...
+
+        buildTypes {
+            release {
+                minifyEnabled false
+                proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+            }
+        }
+    }
+
+    dependencies {
```
jordan-griffin, line 28: Other than adding the dependency version variables nothing here has changed. Its just been moved from the core SDK build script to this level. All these dependencies are common to all submodules so it's high level stuff like setting up Kotlin, JUnit, Mockito, etc

sdk/core/build.gradle
```diff
@@ -0,0 +1,2 @@
+description = 'Core featureset of the Klaviyo SDK including SDK configuration, session tracking and analytics'
+evaluationDependsOn(':sdk')
```
jordan-griffin, line 2: This declares that buildouts of this module depend on the SDK module (which is the higher level)

sdk/push/build.gradle
```diff
@@ -0,0 +1,6 @@
+project.description = 'Push functionality for the Klaviyo SDK suite'
+evaluationDependsOn(':sdk')
+
+dependencies {
+    implementation project(':sdk:core')
```
jordan-griffin, line 5: Our push module will have a dependency on our core SDK so I'm gonna declare that here

chrisconlon-klaviyo (approved):

PR #21, opened 2020-09-23
Create and store push tokens on the client device

jordan-griffin:
# Purpose
Creates and stores push tokens on the client by integrating FCM dependencies with the core SDK.

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

Integrates FCM into the core SDK and uses an overridden FCM service to listen for the creation of push tokens which are then written to the existing config file and stored in the app files. This file will be correctly deleted and regenerated on any install of the application, and any time the push token is refreshed by FCM, the value in the config file will be written to file. An existing push token can be read from file and included in our track or identify calls to send them to the Klaviyo backend.

## Changelog / Code Overview

- Added FCM dependencies
- Added the `KlaviyoPushService` implementation of the base `FirebaseMessagingService` to track token generation
- Updated track and identify calls to attach push tokens

## Test Plan

Tested locally, pushing tokens to both localhost and the Mobile Dev accounts in prod. Using a corresponding PR for the Android test app to act as the push client.


sdk/core/src/main/java/com/klaviyo/coresdk/push/KlaviyoPushService.kt
```diff
@@ -0,0 +1,18 @@
+package com.klaviyo.coresdk.push
+
+import com.google.firebase.messaging.FirebaseMessagingService
+import com.klaviyo.coresdk.utils.ConfigFileUtils
+import com.klaviyo.coresdk.utils.ConfigKeys
+
+class KlaviyoPushService: FirebaseMessagingService() {
```
jordan-griffin, line 7: By overriding the FirebaseMessagingService we can listen for changes to the push tokens or for incoming push messages, allowing us to determine what happens on each event. In this case, we need this so we can react to push tokens whenever they are generated (typically if the app is run after a fresh install or if app data was cleared out).

This is Firebase's recommended way of listening for new push tokens.

sdk/core/src/main/java/com/klaviyo/coresdk/push/KlaviyoPushService.kt
```diff
@@ -0,0 +1,18 @@
...
+    companion object {
+        fun getPushToken(): String {
+            return ConfigFileUtils.readValue(ConfigKeys.PUSH_TOKEN_KEY)
+        }
+    }
+    override fun onNewToken(newToken: String) {
+        super.onNewToken(newToken)
+
+        ConfigFileUtils.writeConfigValue(ConfigKeys.PUSH_TOKEN_KEY, newToken)
```
jordan-griffin, line 16: We store push tokens to file. This file is stored in the app's directory so it is cleared out whenever app data is, allowing us. This is in line with how Firebase push tokens are generated and stored too

sdk/core/src/main/java/com/klaviyo/coresdk/utils/ConfigFileUtils.kt
```diff
@@ -0,0 +1,56 @@
+package com.klaviyo.coresdk.utils
```
jordan-griffin, line 1: Not a new file, just moved to a new utils dir

sdk/core/src/main/java/com/klaviyo/coresdk/utils/ConfigFileUtils.kt
```diff
@@ -0,0 +1,56 @@
...
+        }
+        return value
+    }
+
+    fun writeConfigValue(key: String, value: String) {
+        writeConfigValue(openConfigFile(), key, value)
+    }
+
+    fun writeConfigValue(file: File, key: String, value: String) {
+        file.appendText("$key:$value\n")
```
jordan-griffin, line 54: I was using `writeText` here before which is incorrect, I want to use `appendText` so that the file can contain multiple values (UUIDs and push tokens right now). Whenever either of these need to be generated, the file will be deleted anyway since they are only generated on initial install

sdk/core/src/main/java/com/klaviyo/coresdk/utils/ConfigKeys.kt
```diff
@@ -0,0 +1,6 @@
+package com.klaviyo.coresdk.utils
+
+internal object ConfigKeys {
```
jordan-griffin, line 3: just some keys for identifying information in the config file

settings.gradle
```diff
@@ -1,3 +1,3 @@
 // sdk modules
 include ':sdk'
-include ':sdk:core', ':sdk:push'
```
jordan-griffin, line 3: So I actually did away with the push module for now. Initially I thought having a push module separate was the way to go but getting deeper into implementation it was just too messy to try and decouple push logic entirely from the core SDK logic since they are coupled pretty fighting in the Klaviyo APIs thanks to push tokens being passed through in analytics calls

sdk/core/src/main/java/com/klaviyo/coresdk/push/KlaviyoPushService.kt
ndurell: Is this for the app or the sdk? Can you have multiple overrides of this? Customers may use many push services potentially and may need to send to each of them.

ndurell: I'm ok with this in theory but we still might need some API for customers to set tokens. Also we may want to provide a listener here so they can get the to pass to others if necessary.

ndurell (approved):
Had a few questions to think about. I think we can move forward here but may need to see how this comes out in the final version.

sdk/core/src/main/java/com/klaviyo/coresdk/push/KlaviyoPushService.kt
ndurell: actually don't we want to enqueue a request here?

ndurell: Also are we making this public now? I don't want developers to think they should access this class direclty.

sdk/core/src/main/java/com/klaviyo/coresdk/push/KlaviyoPushService.kt
jordan-griffin: Ya this seems to be the normal way to do things in SDKs with push integration. I did notice that a lot of SDKs have a "manual" mode that disables automatic token generation like this in favour of manual calls to the `FirebaseInstanceId` classes, which will generate tokens async when called.

jordan-griffin: One alternative I was thinking of was to create a `generatePushToken()` call that would check for a stored push token on file, and in the case where none exist, reaching out to `FirebaseInstanceId` to generate and store a fresh one.
We would just want to make sure that we aren't trying to generate push tokens in a getter since the process is async and we're not going to return a token straight away.

sdk/core/src/main/java/com/klaviyo/coresdk/push/KlaviyoPushService.kt
jordan-griffin: The config stuff is still internal but I guess you are worried about if they are overriding this service and seeing this logic in place that it's misleading? We should make sure to store the token somewhere though, or else make an async call to fetch it each time the app is opened and store it in memory.

jordan-griffin: I'll enqueue an identify call here though ya. You are right, didn't realise the iOS SDK was doing that straight away whenever a new token is set.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -1,23 +1,27 @@
...
+    }
+
+    fun track(event: KlaviyoEvent, customerProperties: Map<String, Any>, properties: Map<String, Any>? = null) {
         val request = TrackRequest(event.name, customerProperties.toMutableMap(), properties)
         request.generateUnixTimestamp()
         processRequest(request)
     }
 
-    fun identify(properties: Map<String, String>) {
+    fun identify(properties: Map<String, Any>) {
```
jordan-griffin, line 20: This needed to be updated to `Any` to hold onto nested maps (needed for adding push tokens in the append props).  We'll be changing this API a bit later so its less free-form but for the time being we just need the types to comply

PR #22, opened 2020-09-23
Add build variants

jordan-griffin:
# Purpose
Adds build variants to the SDK that allow us to use build configurations to change our server URLs so that we can connect to either localhost or production addresses.

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Changelog / Code Overview

- Adds build variants to the app Gradle script
- Reads the Klaviyo server URL from build configs
- Allow connections to be read over HTTP as well as HTTPS, needed for local testing.
- Some cleanup with network connections

## Test Plan

Tested locally

### DIFF
--- a/sdk/build.gradle
+++ b/sdk/build.gradle
@@ -21,6 +21,13 @@ subprojects {
             release {
                 minifyEnabled false
                 proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+                buildConfigField "String", "KLAVIYO_SERVER_URL", '"https://a.klaviyo.com"'
+            }
+            debug {
+                buildConfigField "String", "KLAVIYO_SERVER_URL", '"https://a.klaviyo.com"'
+            }
+            localDebug {
+                buildConfigField "String", "KLAVIYO_SERVER_URL", '"http://10.0.2.2:8080"'
             }
         }
     }
--- a/sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/NetworkRequest.kt
+++ b/sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/NetworkRequest.kt
@@ -5,9 +5,11 @@ import android.net.ConnectivityManager
 import android.net.NetworkCapabilities
 import android.os.Build
 import android.util.Base64
+import android.webkit.URLUtil
 import com.klaviyo.coresdk.KlaviyoConfig
 import com.klaviyo.coresdk.networking.RequestMethod
 import java.io.*
+import java.lang.Exception
 import java.net.HttpURLConnection
 import java.net.URL
 import javax.net.ssl.HttpsURLConnection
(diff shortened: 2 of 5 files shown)


sdk/build.gradle
```diff
@@ -21,6 +21,13 @@ subprojects {
             release {
                 minifyEnabled false
                 proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+                buildConfigField "String", "KLAVIYO_SERVER_URL", '"https://a.klaviyo.com"'
+            }
+            debug {
+                buildConfigField "String", "KLAVIYO_SERVER_URL", '"https://a.klaviyo.com"'
+            }
+            localDebug {
+                buildConfigField "String", "KLAVIYO_SERVER_URL", '"http://10.0.2.2:8080"'
```
jordan-griffin, line 30: 10.0.2.2 is the address on android emulators that will allow us to call in to the localhost address of our desktops. So local-klaviyo running on 127.0.0.1 can be reached within the emulator via this address

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/NetworkRequest.kt
```diff
@@ -54,7 +56,11 @@ internal abstract class NetworkRequest {
         }
 
         val url = buildURL()
-        val connection: HttpsURLConnection = url.openConnection() as HttpsURLConnection
+        val connection = if (URLUtil.isHttpsUrl(url.toString())) {
```
jordan-griffin, line 59: We need to correctly determine our protocol for the connection. The SDK is not going to connect to a local klaviyo instance over HTTPS so I'm adding in HTTP connections

sdk/core/src/test/java/com/klaviyo/coresdk/networking/requests/IdentifyRequestTest.kt
```diff
@@ -1,16 +1,13 @@
 package com.klaviyo.coresdk.networking.requests
 
 import android.content.Context
-import com.klaviyo.coresdk.ConfigFileUtils
```
jordan-griffin, line 4: Just cleaning up some unused imports from a previous push.

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/NetworkRequest.kt
```diff
@@ -67,17 +77,19 @@ internal abstract class NetworkRequest {
...
             }
         }
-
-        connection.connect()
-
-        return readResponse(connection)
+        return try {
+            connection.connect()
+            readResponse(connection)
+        } catch (ex: Exception) {
```
ndurell, line 87: Uhh this feels a little too broad IMO.

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/NetworkRequest.kt
```diff
@@ -67,17 +77,19 @@ internal abstract class NetworkRequest {
                 val outputStream = connection.outputStream
                 val writer = BufferedWriter(OutputStreamWriter(outputStream, "UTF-8"))
                 writer.use {
-                    it.write(payload)
+                    it.write(payload!!)
```
ndurell, line 80: Why is this necessary?

jordan-griffin: Suppresses a nullable warning. The if statement above should not allow this to ever be null or empty here

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/NetworkRequest.kt
jordan-griffin: Ya true, meant to revisit this

ndurell (approved):

PR #23, opened 2020-09-28
Add dokka

jordan-griffin:
# Purpose
Adds Dokka to the Android SDK and adds basic KDoc comment blocks to much of the SDK

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

Adds Dokka for documentation purposes. Dokka will compile comment blocks into pages of documentation in markdown. Very similar to JavaDocs. In this PR we add Dokka as a dependency, add the task to call it, and add KDoc comments to many of the functions and classes in the core SDK

## Changelog / Code Overview

- Most files had KDoc comment blocks added
- The base level Gradle script had Dokka added as a dependency and a task added to kick off the Dokka documentation process.

## Test Plan

Locally compiled the KDocs via Dokka and viewed the results in the build directory.

## What to look for

Ensure that comments are using the right annotations.

## Related Issues/Tickets

https://klaviyo.tpondemand.com/entity/45797-add-dokka-to-the-android-sdk


build.gradle
```diff
@@ -1,16 +1,20 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
+apply plugin: 'org.jetbrains.dokka'
+
 buildscript {
     apply from: 'versions.gradle'
 
-    ext.kotlin_version = '1.3.72'
```
jordan-griffin, line 6: Moved this into `versions.gradle`. Should have been in there before

build.gradle
```diff
@@ -27,3 +31,22 @@ allprojects {
 task clean(type: Delete) {
     delete rootProject.buildDir
 }
+
+dokka() {
```
jordan-griffin, line 35: This dokka block is what configures the Gradle task internal to the Dokka dependency. In the future I might expand this to have different tasks, one that documents only public API members, one that documents all code, etc

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -5,19 +5,43 @@ import com.klaviyo.coresdk.networking.requests.IdentifyRequest
...
+     * Additional properties associated to the event that are not for identifying the customer
+     */
     fun track(event: KlaviyoEvent, customerProperties: Map<String, String>, properties: Map<String, String>? = null) {
         val request = TrackRequest(event.name, customerProperties.toMutableMap(), properties)
         request.generateUnixTimestamp()
         processRequest(request)
     }
 
+    /**
+     * Creates an identify request and sends it off for processing.
```
chrisconlon-klaviyo, line 29: Might be useful to clarify where it sends it off to (Klaviyo's asynchronous APIs).

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoEvent.kt
```diff
@@ -1,5 +1,10 @@
 package com.klaviyo.coresdk.networking
 
+/**
+ * Events recognized by Klaviyo
```
chrisconlon-klaviyo, line 4: This might be misleading since we can have custom events. Maybe we could call these common events tracked in Klaviyo?

sdk/core/src/main/java/com/klaviyo/coresdk/networking/NetworkBatcher.kt
```diff
@@ -49,6 +74,14 @@ object NetworkBatcher {
             }
         }
 
+        /**
+         * Empties the batcher of all queued requests by instantly sending each request to Klaviyo
+         * The queue will empty whenever the triggers specified by the [KlaviyoConfig] are met
+         *
+         * @param forceEmpty Overrides the typical queue emptying triggers to force the queue to empty itself
+         *
+         * @return Whether the request queue was emptyed or not
```
chrisconlon-klaviyo, line 83: `emptyed` -> `emptied`

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/KlaviyoRequest.kt
```diff
@@ -14,24 +20,40 @@ internal abstract class KlaviyoRequest: NetworkRequest() {
     override var queryData: String? = null
     override var payload: String? = null
 
+    /**
+     * Adds an anonymous ID to the given property map
```
chrisconlon-klaviyo, line 24: Should we expand here to say that this is the identifier that the device will be identified by in Klaviyo if an email or phone number is not provided?

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/TrackRequest.kt
```diff
@@ -46,6 +52,9 @@ internal class TrackRequest (
         return json.toString()
     }
 
+    /**
+     * Generates a Unix timestamp to store  as the [timestamp] on this object
```
chrisconlon-klaviyo, line 56: Extra space in there.

chrisconlon-klaviyo (approved):
Some small comments, looks good otherwise!

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoEvent.kt
jordan-griffin: True. I actually have a value in this sealed class for specifying a custom event but the comment isn't highlighting its existence so I'll upload to include a note on that

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/TrackRequest.kt
kennyklaviyo: curious when [] brackets are used in documentation, seems like its generally used for class types but looks like here is for a property?

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/KlaviyoRequest.kt
jordan-griffin: Good idea!

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/TrackRequest.kt
jordan-griffin: Ya classes or properties can be linked to. I don't intend to just use it for classes even though it may be weighted that way. I'd like to provide linking to property mentions where applicable for easy navigation between the doc pages.

PR #24, opened 2020-09-29
API improvements

jordan-griffin:
# Purpose
This is work intended to improve the user-facing API of the Android SDK around making analytics requests.

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

Makes improvements to the front-facing API so that analytic requests cannot be made without a new `KlaviyoProperties` object implementation. These objects provide helper methods that control the inputs that are considered valid within a Klaviyo request, so that property maps are built up in a way that is less error-prone and more controlled.

How it would look for users to call this new API when building out a track request:
```
val customerProperties = KlaviyoCustomerProperties()
customerProperties.addEmail("test@test.com")
customerPropterties.addProperty(KlaviyoCustomerPropKeys.FIRST_NAME, "Jordan")

val eventProperties = KlaviyoEventProperties()
eventProperties.addCustomProp("my_key", "my_value")

Klaviyo.track(KlaviyoEvent.PLACED_ORDER, customerProperties, eventProperties)
```

## Changelog / Code Overview

- Added new `KlaviyoPropertyKeys` sealed classes to hold onto all the valid property keys Klaviyo accepts for customer and event property maps
- Added a new `KlaviyoProperties` class to control the accepted content into the customer and event property maps that we pass in to our analytics requests. These classes obscure a standard hashmap, publicizing a set of helper functions that accept our defined `KlaviyoPropertyKeys` as allowed map keys.
- Updated unit tests and added a new test to ensure that property maps containing custom nested hashmaps will still serialize into JSON property
- Updated our APIs to match the changes
- Moved the appending of the anonymous IDs in to the new `KlaviyoProperties` classes

## Test Plan

Tested locally

## What to look for

We should consider if this is the style of API we want. This provides a less user-error-prone API, but it could be seen as more obscure than necessary to a user. We should consider all angles here to make sure that this is the sort of API that we want.

## Related Issues/Tickets

https://klaviyo.tpondemand.com/entity/51775-add-a-property-map-builder-so


sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoProperties.kt
```diff
@@ -0,0 +1,74 @@
+package com.klaviyo.coresdk.networking
+
+import com.klaviyo.coresdk.ConfigFileUtils
+
+// Didn't inherit from a mutablemap class cause I don't want to expose the standard put options and force this API to be the only access the user has to the map
+/**
+ * Abstract class that wraps around a map to control access to its contents.
+ * Provides helper functions to add properties to the map while controlling the keys available for entry
+ */
+abstract class KlaviyoProperties {
```
jordan-griffin, line 10: I didn't inherit from a mutablemap class cause I don't want to expose any of the standard map functions and this way I can force this API to be the only access the user has to the contained map

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoProperties.kt
```diff
@@ -0,0 +1,74 @@
...
+ */
+class KlaviyoCustomerProperties: KlaviyoProperties() {
+    /**
+     * Adds an anonymous ID to the customer properties. This is fetched from existing configuration
+     */
+    internal fun addAnonymousId() {
+        propertyMap[KlaviyoCustomerPropKeys.ANONYMOUS_ID.name] = "Android:${ConfigFileUtils.readOrCreateUUID()}"
+    }
+
+    fun addId(value: String) {
```
jordan-griffin, line 46: I added some helper functions for appending common data types to the property maps (the identifiers in this case). Maybe this is a little more confusing from the point of view of an end-user with no internal knowledge of the API rather than just using the `addProperty` function for everything?

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoProperties.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    fun addEmail(value: String) {
+        propertyMap[KlaviyoCustomerPropKeys.EMAIL.name] = value
+    }
+
+    fun addPhoneNumber(value: String) {
+        propertyMap[KlaviyoCustomerPropKeys.PHONE_NUMBER.name] = value
+    }
+
+    internal fun addAppendedProp(key: String, value: Any) {
```
jordan-griffin, line 57: I'm internalizing append props cause I don't think there's a use case for these outside of push tokens (which we will append to a request upon generation for the user anyway)? Correct me if I am wrong.

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoProperties.kt
```diff
@@ -0,0 +1,82 @@
...
+    override fun addCustomProp(key: String, value: Any) {
+        propertyMap[KlaviyoCustomerPropKeys.CUSTOM(key).name] = value
+    }
+}
+
+/**
+ * Controls the data that can be input into a map of event properties recognised by Klaviyo
+ */
+class KlaviyoEventProperties: KlaviyoProperties() {
+    fun addValue(value: String) {
```
ndurell, line 75: this one is strange? is the key always the same?

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoProperties.kt
```diff
@@ -0,0 +1,82 @@
...
+    internal fun addAppendedProp(key: String, value: Any) {
+        if (propertyMap.containsKey(KlaviyoCustomerPropKeys.APPENDED.name)) {
+            val appendMap = propertyMap[KlaviyoCustomerPropKeys.APPENDED.name] as MutableMap<String, Any>
+            appendMap[key] = value
+        } else {
+            propertyMap[KlaviyoCustomerPropKeys.APPENDED.name] = mutableMapOf(key to value)
+        }
+    }
+
+    override fun addCustomProp(key: String, value: Any) {
```
ndurell, line 66: Do you need this if custom exists?

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoProperties.kt
jordan-griffin: Oh ya I was going to remove this and ask a question on it actually.
I was unsure if this was an actual assigned key that Klaviyo recognised since I see `$value` thrown around a lot but I don't think its actually assigned to anything in the app code after a very quick peak

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoProperties.kt
jordan-griffin: No, not needed. Similar to what I posted above about having `addEmail` and `addPhoneNumber` calls I just thought it would be a helper for defining custom props, but I understand if it seems a bit confusing from the outside looking in

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoProperties.kt
jordan-griffin: Nevermind I was wrong, it is an assigned field used to define values for Placed Order events

ndurell (commented):
We should take this for a spin with some common example events maybe from a real customer.

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoPropertyKeys.kt
```diff
@@ -0,0 +1,46 @@
...
+    object COUNTRY: KlaviyoCustomerPropKeys("\$country")
+    object ZIP_CODE: KlaviyoCustomerPropKeys("\$zip")
+    object IMAGE: KlaviyoCustomerPropKeys("\$image")
+    object CONSENT: KlaviyoCustomerPropKeys("\$consent")
+
+    // Custom properties
+    class CUSTOM(propertyName: String): KlaviyoCustomerPropKeys(propertyName)
+
+    // Other
+    internal object APPENDED: KlaviyoCustomerPropKeys("\$append")
```
kennyklaviyo, line 35: rename to append

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoProperties.kt
```diff
@@ -0,0 +1,82 @@
...
+        propertyMap[KlaviyoCustomerPropKeys.CUSTOM(key).name] = value
```
kennyklaviyo, line 67: do we want to do anything to prevent overwriting? if a developer creates a "custom" property and sets the key to be "$email" or something?

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoProperties.kt
jordan-griffin: Will update the function name to `setIdentifier` so. Also pretty sure its always a string ya

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoProperties.kt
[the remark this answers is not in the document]
jordan-griffin: I can swap out the `Any` type for `Serializable` which is an interface implemented by many types in Java, like HashMaps. It just means I need to make sure I declare my maps as hashmaps and not mutablemaps. HashMaps are a serializable implementation of mutablemaps, shouldn't be any issues with using them in place.

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/TrackRequest.kt
```diff
@@ -32,13 +34,13 @@ internal class TrackRequest (
         }
      */
     override fun buildKlaviyoJsonQuery(): String {
-        addAnonymousIdToProps(customerProperties)
+        customerProperties.addAnonymousId()
```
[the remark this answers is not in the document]
jordan-griffin: Good call, I hadn't realised those are no longer in the right place. Gonna move to a constructor

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoProperties.kt
jordan-griffin: Hmmm, I'll have to think about that. Implementing would be straightforward, just iterate over the subclasses in my sealed class. Maybe some way to tell the user about what they are doing is overwriting a reserved key, but I'm not 100% convinced that this is the route I gotta be taking here

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -26,6 +28,7 @@ object Klaviyo {
         UserInfo.email = email
     }
 
+
```
kennyklaviyo, line 31: extra line

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoProperties.kt
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/**
+ * Controls the data that can be input into a map of customer properties recognised by Klaviyo
+ */
+class KlaviyoCustomerProperties: KlaviyoProperties() {
+
+    init {
+        setAnonymousId()
+        setDefaultEmail()
```
ndurell, line 42: you have to have an email?

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoProperties.kt
```diff
@@ -0,0 +1,112 @@
...
+    }
+}
+
+/**
+ * Controls the data that can be input into a map of customer properties recognised by Klaviyo
+ */
+class KlaviyoCustomerProperties: KlaviyoProperties() {
+
+    init {
+        setAnonymousId()
```
ndurell, line 41: Crap sorry maybe this is bad too...this could block the main thread. Probably it was better before since that method is always access on the main thread.

ndurell (approved):
I think it's a great improvement!

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoProperties.kt
jordan-griffin: Ya, actually an idea I stole from the iOS SDK. Its cause of the automatic token generation, it could happen before the user assigns any email address, so the token will upload to Klaviyo as part of an anonymous profile. Putting this call here will see if we have an email address cached already from a previous request and add it to the request so that it associates to a real profile and reduces the creation of anonymous profiles

jordan-griffin: As an alternative to this, I can later add the manual token generation path, which will put the time the token is generated and sent off to Klaviyo more in the user's hands with a single API call

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoProperties.kt
jordan-griffin: Ah damn you are right I didn't even consider this stupidly enough.

Although, since its a shared preference access, and our shared preferences file remains very very small, there's actually no need to keep this off the UI thread. It will not actually block execution cause its so simple

jordan-griffin: But if you build up a large shared preferences file then it can block. So I still think this is smell that should change

kennyklaviyo (approved):

PR #28, opened 2022-12-06
Add test for requesting a new token

jordan-griffin:
# Purpose
Just adds one more test to boost coverage a bit. Test covers the case where a new push token has been generated by FCM and is sent as part of an identify call before being popped into the shared preferences for the SDK.

**Type of PR**
* [ ] Bug Fix
* [ ] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [x] Other (tests)


evan-masseau (approved):
works for me!

PR #31, opened 2023-01-03
Ecm/sdk api changes

evan-masseau:
# Purpose

**Type of PR**
* [ ] Bug Fix
* [ ] Feature Work
* [ ] Revert
* [x] Refactor / Code Cleanup
* [x] Other - Documentation/drafting readme

## Short Description

Updating API and documentation. This is really just a stepping stone, but it seemed like a good smaller unit of work for code review before I get into bigger stuff. Reminder [this](https://docs.google.com/document/d/1smKMAl9H6Vm0WiaG3H6nM-5kv9izTSQPQo9a7ZYpz3Y/edit) is the google doc with API discussion 

FYI This will wind up getting superseded by #32, in which I continued off of this branch into a larger refactor project.

## Changelog / Code Overview

Renamed public API methods
Added phone as UserInfo identifier
Added `Klaviyo.configure` for setting apiKey and context, removed `KlaviyoConfig` from public view (seems simpler that way)
Updated to latest dokka and experimented with generating docs with it.


sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -9,54 +10,126 @@ import com.klaviyo.coresdk.networking.requests.KlaviyoRequest
...
+     * @param networkFlushDepth
+     * @param networkFlushCheckInterval
+     * @param networkUseAnalyticsBatchQueue
+     * @return
      */
-    fun setUserEmail(email: String) {
+    fun configure(
+        apiKey: String,
+        applicationContext: Context,
+        networkTimeout: Int = KlaviyoConfig.NETWORK_TIMEOUT_DEFAULT,
```
ndurell, line 37: Interesting we are allow these be configured.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -9,54 +10,126 @@ import com.klaviyo.coresdk.networking.requests.KlaviyoRequest
...
+     * If customer properties are not set, this will fallback on the current profile identifiers
+     *
+     * @param customerProperties A map of customer property information.
+     * Defines the customer that triggered this event
+     * @param properties A map of event property information.
+     * Additional properties associated to the event that are not for identifying the customer
+     */
+    fun createEvent(
+        event: KlaviyoEvent,
+        customerProperties: KlaviyoCustomerProperties? = null,
```
ndurell, line 123: nit: might move event first...

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: IDK if this will survive, it was publicly exposed before, so I kept it public for now.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: Had the same thought while I was writing this readme section!

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -9,54 +10,126 @@ import com.klaviyo.coresdk.networking.requests.KlaviyoRequest
...
-    fun setUserEmail(email: String) {
+    fun configure(
+        apiKey: String,
+        applicationContext: Context,
+        networkTimeout: Int = KlaviyoConfig.NETWORK_TIMEOUT_DEFAULT,
+        networkFlushInterval: Int = KlaviyoConfig.NETWORK_FLUSH_INTERVAL_DEFAULT,
+        networkFlushDepth: Int = KlaviyoConfig.NETWORK_FLUSH_DEPTH_DEFAULT,
+        networkFlushCheckInterval: Int = KlaviyoConfig.NETWORK_FLUSH_CHECK_INTERVAL,
+        networkUseAnalyticsBatchQueue: Boolean = KlaviyoConfig.NETWORK_USE_ANALYTICS_BATCH_QUEUE,
+    ) = apply {
```
kennyklaviyo, line 42: the ` = apply` shouldn't be necessary

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -9,54 +10,126 @@ import com.klaviyo.coresdk.networking.requests.KlaviyoRequest
...
+     * Defines the customer that triggered this event
+     * @param properties A map of event property information.
+     * Additional properties associated to the event that are not for identifying the customer
+     */
+    fun createEvent(
+        event: KlaviyoEvent,
+        customerProperties: KlaviyoCustomerProperties? = null,
+        properties: KlaviyoEventProperties? = null
+    ) {
+        val profile = customerProperties ?: KlaviyoCustomerProperties().also { setEmail(UserInfo.email) }
```
ndurell, line 126: Should you set other identifiers here too? phone number and external id?

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -9,54 +10,126 @@ import com.klaviyo.coresdk.networking.requests.KlaviyoRequest
...
+     * @param phone Phone number for active user
      */
-    fun track(
-        event: KlaviyoEvent,
-        customerProperties: KlaviyoCustomerProperties,
-        properties: KlaviyoEventProperties? = null
-    ) {
-        val request = TrackRequest(KlaviyoConfig.apiKey, event, customerProperties, properties)
-        processRequest(request)
+    fun setPhone(phone: String) = apply {
```
ndurell, line 82: Maybe this:
```suggestion
    fun setPhoneNumber(phoneNumber: String) = apply {
```
I also noticed the v3 api seems to check the format of this.

README.md
```diff
@@ -1,17 +1,74 @@
...
+
+## Installation
+*TODO*
 
 ## Core SDK
 
-Android SDK allows developers to incorporate Klaviyo event and profile tracking functionality within native Android applications.
+### Configuration
+The SDK must be configured with the public API key for your Klaviyo account.
+It also requires application context so the SDK can be responsive to
```
jordan-griffin, line 21: `It also requires the application context`?

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -9,54 +10,126 @@ import com.klaviyo.coresdk.networking.requests.KlaviyoRequest
...
+            .networkUseAnalyticsBatchQueue(networkUseAnalyticsBatchQueue)
+            .build()
+    }
+
+    //region Fluent setters
+
+    /**
+     * Assigns an email address to the current UserInfo
+     *
+     * UserInfo is saved to keep track of current profile details and
```
ndurell, line 59: Wonder if we should mention UserInfo if it's not exposed.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -9,54 +10,126 @@ import com.klaviyo.coresdk.networking.requests.KlaviyoRequest
...
     }
 
     /**
-     * Creates an identify request for the Klaviyo APIs and begins processing.
+     * Clears all stored UserInfo identifiers (e.g. email or phone)
+     *
+     * This should be called whenever an active user in your app is removed
+     * (e.g. after a logout)
+     */
+    fun reset() = apply {
```
kennyklaviyo, line 92: maybe name this closer to reseting a the profile

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -9,54 +10,126 @@ import com.klaviyo.coresdk.networking.requests.KlaviyoRequest
...
+    //region Analytics API
+
+    /**
+     * Queues a request to identify profile properties to the Klaviyo API
      * Identify requests track specific properties about a user without triggering an event
      *
      * @param properties A map of properties that define the user
      */
-    fun identify(properties: KlaviyoCustomerProperties) {
+    fun createProfile(properties: KlaviyoCustomerProperties) {
```
kennyklaviyo, line 106: why `create`?

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: Yes what I'd like to do here long term is just have a method in `UserInfo` that creates a `KlaviyoCustomerProperties` object from the current internal UserInfo

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
ndurell: That's the name of the v3 api.

ndurell: But agreed I wrestled with this name.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: I thought of using `apply` here and on other methods to make them chainable, such as 
```Klaviyo.configure()
.setEmail()
.setPhone()```
but i'm not wedded to it.

File: sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt (hunk 1 above)
Comment (jordan-griffin) [reply sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt:37]: I vaguely remember some ancient conversation around this years ago. There was some reason we had them exposed at the time but it's likely overkill and something we can just make private before we launch

File: README.md (hunk 6 above)
Comment (evan-masseau) [reply README.md:21]: I don't love this sentence but I was just trying to explain why a required argument of `Klaviyo.configure` is the application's `Context` object.

Comment (evan-masseau) [reply README.md:21]: Maybe better if I at least put that in backticks like
> We require access to the application `Context` so....

File: sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt [hunk 10]
```diff
@@ -9,54 +10,126 @@ import com.klaviyo.coresdk.networking.requests.KlaviyoRequest
...
+            .applicationContext(applicationContext)
+            .networkTimeout(networkTimeout)
+            .networkFlushInterval(networkFlushInterval)
+            .networkFlushDepth(networkFlushDepth)
+            .networkFlushCheckInterval(networkFlushCheckInterval)
+            .networkUseAnalyticsBatchQueue(networkUseAnalyticsBatchQueue)
+            .build()
+    }
+
+    //region Fluent setters
```
jordan-griffin, line 54: Whats this comment for?

jordan-griffin: Oh sorry, I see now

evan-masseau: Region/endregion makes a collapsible section, its just a code-organization/IDE thing

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
ndurell: Yah I don't have it yet...also not sure we'd want folks mucking with it.

ndurell: Or making it easy to muck with it.

README.md
```diff
@@ -1,17 +1,74 @@
...
+## DISCLAIMER
+*This project is still in pre-alpha,
+breaking changes are still being made to the API. 
+This is not yet intended for public use*
+
+Android SDK allows developers to incorporate Klaviyo event and profile tracking functionality
+within native Android applications.
+The SDK assists in identifying users and tracking user events.
+Once integrated, your marketing team will be able to better understand your app users' needs and
+send them timely messages via FCM.
```
ajaysubra, line 12: Should this also mention about push notifications?

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: Something I realized while writing the readme is this is the only way of setting any profile properties other than email/phone, so `create` does feel like a misnomer.

README.md
jordan-griffin: Ya I like the sound of that. But maybe `application Context` in backticks rather than just `context`?

evan-masseau: `applicationContext` is the property's name, `Context` is the type name... but yeah `applicationContext` reads best.

README.md
```diff
@@ -71,4 +128,7 @@ class YourPushService: KlaviyoPushService() {
         KlaviyoPushService.handlePush(message.data, KlaviyoCustomerProperties())
     }
 }
-```
\ No newline at end of file
+```
+
+## Code Documentation
+Browse complete code documentation autogenerated with dokka [here](https://klaviyo.github.io/klaviyo-android-sdk/)
```
ajaysubra, line 134: The generated docs look great. @ndurell we should do something similar with docc. Once docc is setup may be github actions can generate them whenever something is merged to main and host it.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: BTW these parameters were the reason the tester app only sends any API calls after pressing the button twice...  `networkFlushDepth` was set to 2, when I changed it to 1 it was fine

ndurell: We might want to adjust that higher for production.

evan-masseau: For sure. But that was an override specified from the test-app code.

ajaysubra: If an app developer knows that they are building and distributing their app in regions where network latency is a concern would it make sense for them to configure a large time out?

sdk/core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
```diff
@@ -17,13 +17,13 @@ class KlaviyoMissingContextException : Exception("You must add your application
...
  * Stores all configuration related to the Klaviyo Android SDK.
  */
-object KlaviyoConfig {
-    private const val NETWORK_TIMEOUT_DEFAULT: Int = 500
-    private const val NETWORK_FLUSH_INTERVAL_DEFAULT: Int = 60000
-    private const val NETWORK_FLUSH_DEPTH_DEFAULT: Int = 20
-    private const val NETWORK_FLUSH_CHECK_INTERVAL: Int = 2000
-    private const val NETWORK_USE_ANALYTICS_BATCH_QUEUE: Boolean = true
-    private val SYSTEM_CLOCK: Clock = SystemClock
+internal object KlaviyoConfig {
```
ajaysubra, line 20: what is the default access modifier in Kotlin? In Swift it's `internal` by default.

sdk/core/src/test/java/com/klaviyo/coresdk/KlaviyoTest.kt
```diff
@@ -1,14 +1,56 @@
...
-        Klaviyo.setUserEmail(email)
+        Klaviyo.setEmail(email)
 
         assert(UserInfo.email == email)
     }
-}
\ No newline at end of file
+
+    @Test
+    fun `Sets user phone into info`() {
```
ajaysubra, line 39: like how these tests read.

ajaysubra (commented):
non blocking comments.

README.md
ndurell: Yup!

sdk/core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
evan-masseau: seems to be public

sdk/core/src/test/java/com/klaviyo/coresdk/KlaviyoTest.kt
evan-masseau: neat right? (i can't take credit)

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: Could be an argument for keeping it configurable. On the other hand, I don't think I've ever encountered an SDK of this sort that allowed this level of customization.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: Yes I was thinking UserInfo should have format validators on set.

README.md
evan-masseau: yeaaa it should have the word push in there thanks

evan-masseau: copied that from the ios sdk its noah's fault :p

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
ndurell: Yah might not be a bad idea however I'm not sure I want to stuff all that in the sdks. More just mentioning in case you get confused during testing.

sdk/core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
ndurell: I think that is package private potentially. Assuming the test app is under the same package.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: you're thinking better to keep the client SDK dumb, let the klaviyo api backend be responsible for validation?

sdk/core/src/main/java/com/klaviyo/coresdk/KlaviyoConfig.kt
evan-masseau: test app is its own repo altogether, changing this to internal took access away from the test app, which was why I put the arguments in `Klaviyo.configure`

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: how about `setProfile`? see update

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: I'm marking this thread resolved but making a note of this convo in the android arch document

sdk/core/src/main/java/com/klaviyo/coresdk/networking/UserInfo.kt
```diff
@@ -6,6 +6,7 @@ package com.klaviyo.coresdk.networking
  * Stores information on the currently active user
  */
 internal object UserInfo {
+    var external_id: String = ""
```
kennyklaviyo, line 9: camelcase

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
kennyklaviyo: maybe... its nice to have the chaining ability however a developer implementing this probably wont even have an email/phone to set when they should be calling the configure so its not always going to be useful

evan-masseau: Yeah true, in my use case in the test app I had a hardcoded email, but that is hardly typical. I'll take it out from here.

kennyklaviyo: that said, it doesnt hurt but if you want to keep it, we should be explicit about the return value of these functions:
```
fun setEmail(email: String): KlaviyoConfig = apply {
    ...
}
```

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -18,112 +18,127 @@ object Klaviyo {
...
      *
      * This should be called whenever the active user in your app changes
      * (e.g. after a fresh login)
      *
      * @param email Email address for active user
+     * @return
      */
-    fun setEmail(email: String) = apply {
+    fun setEmail(email: String): Klaviyo = apply {
+        // TODO validate email format?
```
jordan-griffin, line 56: What would we do with email addresses we can't validate though?

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -18,112 +18,127 @@ object Klaviyo {
...
-        UserInfo.phone = phone
+    fun setPhoneNumber(phone: String): Klaviyo = apply {
+        // TODO validate phone format?
+        UserInfo.phoneNumber = phone
+        this.setProfile(UserInfo.getAsCustomerProperties())
     }
 
     /**
-     * Clears all stored UserInfo identifiers (e.g. email or phone)
+     * Assigns an external to the current internally tracked profile
```
jordan-griffin, line 79: `an external ID` I think you meant?

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: Throw I guess, Noah pointed out this might be too much clientside intelligence.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -18,112 +18,127 @@ object Klaviyo {
...
      * Queues a request to identify profile properties to the Klaviyo API
      * Identify requests track specific properties about a user without triggering an event
      *
      * @param properties A map of properties that define the user
+     * @return
      */
-    fun createProfile(properties: KlaviyoCustomerProperties) {
+    fun setProfile(properties: KlaviyoCustomerProperties): Klaviyo = apply {
+        // TODO Extract phone/email and save in state?
+        //  I'm still a bit confused about the exact purpose of this method
```
jordan-griffin, line 104: This was originally just a call-through to issue an identity request to Klaviyo. I know it got renamed to match the iOS SDK a bit closer, but is there anything in particular you are confused about here?

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
jordan-griffin: Was thinking the same. I'm not sure if I'm just overlooking some use case that would be handy, but I feel like we're pretty agnostic here and will just pass up to the API whatever we have been given and let the `CustomerProcessor` within Klaviyo figure it out

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: Oh yeah thanks for highlighting that. 
What is unclear to me is how the profile properties passed in as the argument should relate to the properties in internal state. E.g. if email is set in this argument's properties, should that email replace the current email in state? If email is _not_ specified, should we add the current email from state? Should it be a pure pass through that doesn't touch internal state or read from it?

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: Thanks, I wanted to hear at least one other person's feedback but I figured I'd remove this comment.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: Thanks, made updates to remove chaining from initialize and specify return type explicitly on the others

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
jordan-griffin: Oo I see, good point. Since we weren't really dealing with state before, we just exposed the function with a map of customer properties as an input but if we start doing proper persistence in the SDK this makes less sense.

It could be confusing for internal state to overwrite what a customer may pass in here (and they likely will want to pass in some information here since this takes in properties beyond the basic identifiers we store). I wonder if we should take the route where we overwrite whats stored in our persistence layer with any identifier values that have been attached to the customer properties map.

I could see this going multiple ways based on different opinions

evan-masseau: Yeah, and I think it depends on naming too. `createProfile` sounds a little more like a straight up pass-through that ignores state. `setProfile` sounds more like its going to overwrite internal state (and subsequently make the API call). 
And by the way thinking about this is what prompted me to propose a `setProfileAttribute` method where one could update any properties without having to specify the a whole profile object. Just realizing now I didn't put that method in this PR, but it was in the google doc.

evan-masseau: I'll push another commit with a proposal for merging this incoming argument with the internal state

evan-masseau: @ndurell your input would be valuable too. I _think_ what Jordan and I are talking about is similar to how `trackPersonWithInfo` works, which would be the iOS analog until you rename things.

sdk/core/src/main/java/com/klaviyo/coresdk/networking/UserInfo.kt
```diff
@@ -6,9 +6,70 @@ package com.klaviyo.coresdk.networking
...
+    }
+
+    /**
+     * Two-way merge of a [KlaviyoCustomerProperties] object with UserInfo
+     * Identifiers present on the incoming object will be applied to UserInfo
+     * Any other identifiers will be added to the properties object from UserInfo
+     *
+     * @param properties
+     */
+    fun mergeCustomerProperties(properties: KlaviyoCustomerProperties) {
```
evan-masseau, line 69: @jordan-griffin following on our thread about `setProfile`, here's an idea to merge the incoming profile properties _into_ state and update it with other identifiers _from_ state

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: Updated, [see here](https://github.com/klaviyo/klaviyo-android-sdk/pull/31/files#r1066335425) added a two-way merge for this situation.

evan-masseau:
Closing: generated useful discussion but my work on #32 has made this branch irrelevant at this point.

PR #32, opened 2023-01-09
Architecture rework, new API and implementation 

evan-masseau:
# Purpose
Major restructure of the SDK to break functionality up into distinct services defined by interfaces. Incorporates the existing code into those broken out services, but also adds some major new functionality like persistence of state/queues

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [x] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Changelog / Code Overview

- Created a very basic "service registry" object that centralizes services so they can be shared only via interfaces.
- Updated all our unit tests to be much more isolated. Mocking/stubbing is much easier with the service registry and interfaces. 
- Renamed API methods and data classes per team discussion doc, implemented missing API methods
- `config` package with interface for internal SDK configuration fields
- `lifecycle` package for hooking into the application lifecycle
- `models` package for data representation and persistence
  - Moved existing data classes here, and did some minor refactoring/renaming
  - Added an interface for persistent store and made an example implementation with `SharedPreferences`
- Persistent User state (and added support for email, phone, external ID)
- `networking` package which contains the "coordinator" and "engine" functionality, but only exposes a basic interface for enqueueing API requests.
  -  Persistent network queue
- 

## Test Plan

We started with decent unit test coverage, and I kept it as close to 100% as I could. The interfaces/registry pattern makes it easy to write more isolated unit tests. The models package appears to have a lot of missing coverage, but it is just the property keys which are defined as objects like a pseudo enum. That can be fixed when we revisit which keys we pre-define in the SDK. Also there's some places where I extracted some android utilities into helper classes, things that are just too hard to mock. But as a result those little helpers are holes in coverage.

I think the next step is to merge this and work on the tester app, which will inevitably include some SDK additions and fixes. 

## What to look for

This is now ready for full review, I'm sorry it is so big I started by trying to keep it in small chunks but it just turned into so much work to keep things in separate branches. I think at least one team member should pull it to their local and review in android studio, its a lot to look at on GH. 

I am _more than happy_ to review 1:1 too. 

## Related Issues/Tickets

[Discussion doc](https://klaviyo.atlassian.net/l/cp/M49s08mR)
[TP story on Architecture Rework](https://klaviyo.tpondemand.com/entity/159927-android-sdk-architecture-rework-rfc)


sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -1,73 +1,171 @@
...
+     * This should be called whenever the active user in your app changes
+     * (e.g. after a fresh login)
+     *
+     * @param email Email address for active user
+     * @return
      */
-    fun setUserEmail(email: String) {
-        UserInfo.email = email
+    fun setEmail(email: String): Klaviyo = apply {
+        this.setProfile(Profile().setEmail(email))
```
jordan-griffin, line 77: Possibly dumb question but this looks to me that it initializes a new profile and assigns an email to that. Which would wipe out other profile settings to initialize a new instance (aka: during a login). Does this match up properly with the expectations set by the first line of the docstring for this function though? Its not really the _current_ since if you tried to add an email to an existing profile you'd be resetting it to a new one right?

Am I following that logic correctly?

sdk/build.gradle
```diff
@@ -70,4 +71,18 @@ subprojects {
             include("**/java/**")
         }
     }
+
+    dokkaHtmlPartial {
+        dokkaSourceSets {
+            configureEach {
+                includeNonPublic.set(false)
+                reportUndocumented.set(true)
+                skipDeprecated.set(true)
```
jordan-griffin, line 80: nit: I feel like its typical to display documentation for deprecated functions as long as they are appropriately called out as such?

Honestly I say that but this is probably better since if its not in the docs at all, there's less of a chance someone will try to use it.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
jordan-griffin: Oh actually sorry I just read how `setProfile` works, but merging info. Makes sense

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -1,73 +1,171 @@
...
      *
-     * @param properties A map of properties that define the user
+     * @param profile A map of properties that define the user
+     * @return
      */
-    fun identify(properties: KlaviyoCustomerProperties) {
-        val request = IdentifyRequest(KlaviyoConfig.apiKey, properties)
-        processRequest(request)
+    fun setProfile(profile: Profile): Klaviyo = apply {
+        // TODO debounce so fluent setters don't have to create 1 request per call
```
jordan-griffin, line 137: We need this tracked anywhere? Just to make sure we follow up on it

sdk/core/src/main/java/com/klaviyo/coresdk/config/KlaviyoConfig.kt
```diff
@@ -14,111 +17,111 @@ class KlaviyoMissingAPIKeyException : Exception("You must declare an API key for
...
             if (applicationContext == null) {
                 throw KlaviyoMissingContextException()
             }
 
+            val permissions = applicationContext!!.packageManager.getPackageInfoCompat(
+                applicationContext!!.packageName, PackageManager.GET_PERMISSIONS
+            ).requestedPermissions ?: arrayOf()
+
+            if (Manifest.permission.ACCESS_NETWORK_STATE !in permissions) {
+                throw KlaviyoMissingPermissionException(Manifest.permission.ACCESS_NETWORK_STATE)
```
jordan-griffin, line 107: nice call on checking for this

sdk/core/src/main/java/com/klaviyo/coresdk/model/SharedPreferencesDataStore.kt
```diff
@@ -0,0 +1,54 @@
...
+     *
+     * The write operation is performed async, but has no callback
+     *
+     * @param key The identifying key that the value being written will go by
+     * @param value The value that we are writing to the shared preferences
+     */
+    override fun store(key: String, value: String) {
+        openSharedPreferences()
+            .edit()
+            .putString(key, value)
```
jordan-griffin, line 33: Any need to mention that this obviously is only for pushing string values into sharedpreferences? Are there any cases where we may be pushing non-string data into our storage?

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: Ah thanks, I'd say this is a blocker to even merging this PR, it should have been part of the overall "api updates" work. I will go through my TODOs in android studio.

sdk/build.gradle
evan-masseau: Fair! I think I just copied this from SO. At the moment we don't have deprecated things anyway, but I think we should document them if we ever do.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: Yeah `Profile` is just a data structure, `UserInfo` is the source of truth for current profile internal state. I chose to make all these setters boil down to using the same method `setProfile`

sdk/core/src/main/java/com/klaviyo/coresdk/model/SharedPreferencesDataStore.kt
evan-masseau: Could do! The workaround is just to json encode of course, which is what I'm doing with the network request queue.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -1,73 +1,140 @@
...
+     *
+     * This should be called when you collect additional data about your user
+     * (e.g. first and last name, or an address)
+     *
+     * @param propertyKey
+     * @param value
+     * @return
+     */
+    fun setProfileAttribute(propertyKey: KlaviyoProfileAttributeKey, value: String): Klaviyo =
+        setProfile(Profile().also { it[propertyKey] = value })
```
kennyklaviyo, line 90: should this also use the `apply` pattern

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -1,73 +1,140 @@
...
-    private fun processRequest(request: KlaviyoRequest) {
-        if (KlaviyoConfig.networkUseAnalyticsBatchQueue) {
-            request.batch()
-        } else {
-            request.process()
-        }
+    fun createEvent(event: KlaviyoEventType, properties: Event? = null): Klaviyo = apply {
+        Registry.apiClient.enqueueEvent(
+            event,
+            properties ?: Event(),
```
kennyklaviyo, line 136: should this default just be in the params instead? `, properties: Event = Event()`

sdk/core/src/main/java/com/klaviyo/coresdk/config/KlaviyoConfig.kt
```diff
@@ -14,111 +17,111 @@ class KlaviyoMissingAPIKeyException : Exception("You must declare an API key for
...
         }
     }
 }
+
+internal fun PackageManager.getPackageInfoCompat(packageName: String, flags: Int = 0): PackageInfo =
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+        getPackageInfo(packageName, PackageManager.PackageInfoFlags.of(flags.toLong()))
+    } else {
+        @Suppress("DEPRECATION") getPackageInfo(packageName, flags)
+    }
```
kennyklaviyo, line 127: https://developer.android.com/reference/androidx/core/content/PackageManagerCompat
i think this exists?

sdk/core/src/main/java/com/klaviyo/coresdk/config/KlaviyoConfig.kt
```diff
@@ -14,111 +17,111 @@ class KlaviyoMissingAPIKeyException : Exception("You must declare an API key for
...
             if (apiKey.isEmpty()) {
                 throw KlaviyoMissingAPIKeyException()
             }
             if (applicationContext == null) {
                 throw KlaviyoMissingContextException()
             }
 
+            val permissions = applicationContext!!.packageManager.getPackageInfoCompat(
+                applicationContext!!.packageName, PackageManager.GET_PERMISSIONS
+            ).requestedPermissions ?: arrayOf()
```
kennyklaviyo, line 104: clearer to do `emptyArray()`

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -1,73 +1,140 @@
...
+     * @return
      */
-    fun identify(properties: KlaviyoCustomerProperties) {
-        val request = IdentifyRequest(KlaviyoConfig.apiKey, properties)
-        processRequest(request)
+    fun setProfile(profile: Profile): Klaviyo = apply {
+        // TODO debounce so fluent setters don't have to create 1 request per call
+        // Note - is there value to tighter coupling the UserInfo update with the API call
+        UserInfo.mergeProfile(profile)
+        Registry.apiClient.enqueueProfile(profile)
```
kennyklaviyo, line 108: shouldnt this enqueue `UserInfo.getAsProfile()`?

sdk/core/src/main/java/com/klaviyo/coresdk/config/KlaviyoConfig.kt
evan-masseau: I did find that, but I couldn't figure out how to access it properly in this context. I think the getter is only available as an extension function of Activity

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: `UserInfo` is only concerned with identifiers, so `getAsProfile` would return a `Profile` object with *just* the identifiers. 

`mergeProfile` adds identifiers from internal state to the attributes of the `Profile` object before we enqueue.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: Yup.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: doesn't need it, because setProfile already returns this. Actually I meant to take the `= apply` off all the other setters but `setProfile`.

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -0,0 +1,230 @@
...
+        fun equals(request1: KlaviyoApiRequest, request2: KlaviyoApiRequest): Boolean =
+            request1.uuid == request2.uuid
+    }
+
+    /**
+     * Compiles the base url, path and query data into a [URL] object
+     */
+    val url: URL
+        get() {
+            val baseUrl = Klaviyo.Registry.config.baseUrl
```
jordan-griffin: So with the original structure, one thought I had was that I was building an abstract `NetworkRequest` class separate from whatever concrete implementation decides that it's a Klaviyo request. So I had a few layers of inheritance but thought if there were use cases in the future, maybe we'd have separate request objects that handled authentication/constants/specific-payload-building for non-Klaviyo entities.

jordan-griffin: I may have just over-prepped though. I don't think we have any reason to be making network requests to any other services through this SDK.

But thats the reasons around why it was so layered originally.

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/TrackApiRequest.kt
```diff
@@ -0,0 +1,50 @@
...
+internal class TrackApiRequest(
+    eventType: KlaviyoEventType,
+    profile: Profile,
+    eventAttributes: Event? = null
+) : KlaviyoApiRequest(
+    "client/events",
+    RequestMethod.POST
+) {
+    override var headers: Map<String, String> = mapOf(
+        "Content-Type" to "application/json",
```
jordan-griffin, line 26: Header fields could be constants maybe?

sdk/core/src/test/java/com/klaviyo/coresdk/BaseTest.kt
```diff
@@ -0,0 +1,81 @@
...
+        every { Klaviyo.Registry.networkMonitor } returns networkMonitorMock
+        every { Klaviyo.Registry.dataStore } returns dataStoreSpy
+        every { Klaviyo.Registry.apiClient } returns apiClientMock
+
+        // Mock using latest SDK
+        setFinalStatic(Build.VERSION::class.java.getField("SDK_INT"), 33)
+    }
+
+    /**
+     * Gross way to modify a final static field through reflection
```
jordan-griffin, line 63: `Gross way`. How could you

sdk/push/src/main/java/com/klaviyo/push/KlaviyoPushService.kt
```diff
@@ -18,12 +17,11 @@ import com.klaviyo.coresdk.utils.KlaviyoPreferenceUtils
...
     companion object {
-        internal const val PUSH_TOKEN_PREFERENCE_KEY = "PUSH_TOKEN"
-
-        private const val REQUEST_PUSH_KEY = "\$android_tokens"
+        internal const val PUSH_TOKEN_KEY = "push_token"
+        private const val PUSH_TOKEN_APPEND_KEY = "\$android_tokens"
 
         /**
-         * Registers FCM push token to the current Klaviyo profile
+         * Save the device FCM push token and register to the current profile
```
jordan-griffin, line 25: `Saves` and `registers` for better tense?

sdk/push/src/main/java/com/klaviyo/push/KlaviyoPushService.kt
```diff
@@ -34,70 +32,68 @@ class KlaviyoPushService : FirebaseMessagingService() {
...
+            Klaviyo.createEvent(KlaviyoEventType.OPENED_PUSH, event)
+        }
+
+        /**
+         * Handles a push received while app is in the foreground
          *
-         * @param pushPayload The data attributes of the push notification payload
-         * @param customerProperties Profile with which to associate the event
-         * @param eventProperties Optional additional properties for the event
+         * @param message
```
jordan-griffin, line 71: Description for this param? Seems like you did the others so just noticed this one was missing

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/KlaviyoApiRequest.kt
evan-masseau: I figured, but also figured I didn't see us communicating with other services so subtracted a bit of the complication that way, its pretty easy to add it back too.
The main reason that I took out the abstraction layer was that this structure made it easy to deserialize a JSON representation of any API request object and have it be _functionally_ equivalent even though I don't deserialize into the subclass.

sdk/push/src/main/java/com/klaviyo/push/KlaviyoPushService.kt
```diff
@@ -34,70 +32,68 @@ class KlaviyoPushService : FirebaseMessagingService() {
...
-            customerProperties: KlaviyoCustomerProperties,
-            eventProperties: KlaviyoEventProperties? = null
-        ) {
-            pushPayload["_k"] ?: return // Track pushes originating from klaviyo
-            val properties = eventProperties ?: KlaviyoEventProperties()
-            properties.addCustomProperty("push_token", getCurrentPushToken())
-            pushPayload.forEach { (k, v) -> properties.addCustomProperty(k, v) }
-            Klaviyo.track(KlaviyoEvent.OPENED_PUSH, customerProperties, properties)
+        fun handlePush(message: RemoteMessage): Boolean {
+            message.data["_k"] ?: return false
```
jordan-griffin, line 74: nit: Wonder if we should have a `isKlaviyoPush` function or something similar. Takes in a map or whatever. Just cause you seem to be parsing for this `_k` var from a few places.

sdk/push/src/main/java/com/klaviyo/push/KlaviyoPushService.kt
```diff
@@ -34,70 +32,68 @@ class KlaviyoPushService : FirebaseMessagingService() {
...
+            // TODO could we display the notification in system tray?
```
jordan-griffin, line 75: Sorry not sure I'm following what are you referring to here with this TODO

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/KlaviyoApiRequest.kt
jordan-griffin: Yup thats all fine with me! Just thought I'd specify what was going through my head at the time!

sdk/push/src/main/java/com/klaviyo/push/KlaviyoPushService.kt
evan-masseau: I'm proposing our SDK could automatically display the notification to the app user when received in the foreground. Basically, that functionality that I had to add to the _tester app_ so it could show a foreground notification, the SDK would do that for you.

jordan-griffin: Oh I see, thats what had me confused but I had a feeling it was foreground push related. Makes sense!

sdk/core/src/main/java/com/klaviyo/coresdk/lifecycle/LifecycleMonitor.kt
```diff
@@ -0,0 +1,18 @@
...
+ * Provides methods to react to changes in the application environment
+ */
+interface LifecycleMonitor {
+
+    /**
+     * Register an observer to be notified when all application activities stopped
+     *
+     * @param observer
+     */
+    fun whenStopped(observer: ActivityObserver)
```
kennyklaviyo, line 17: rename to `onAllActivitiesStopped` going off of android convention for observer names

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -1,73 +1,143 @@
...
+     * @param properties Additional properties associated to the event that are not for identifying the profile
+     * @return
      */
-    private fun processRequest(request: KlaviyoRequest) {
-        if (KlaviyoConfig.networkUseAnalyticsBatchQueue) {
-            request.batch()
-        } else {
-            request.process()
-        }
+    fun createEvent(event: KlaviyoEventType, properties: Event = Event()): Klaviyo = apply {
```
evan-masseau, line 136: TODO - Reviewing Noah's PR and I forgot that we decided in the RFC doc to make Event type part of the `Event` data structure

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
kennyklaviyo: but if there are identifiers that are added because of the merge, API request should include both right?

kennyklaviyo: oh wait so youre saing that `mergeProfile` changes the state of the provided `profile` input?

evan-masseau: Yes, and thats what `mergeProfile` does, it mutates the `profile` arg.
The naming may be causing confusion, let me think if there's a better name for `mergeProfile`

sdk/core/src/main/java/com/klaviyo/coresdk/config/KlaviyoConfig.kt
```diff
@@ -0,0 +1,132 @@
...
+
+        override fun applicationContext(context: Context) = apply {
+            this.applicationContext = context
+        }
+
+        override fun debounceInterval(debounceInterval: Int) = apply {
+            if (debounceInterval >= 0) {
+                this.debounceInterval = debounceInterval
+            } else {
+                // TODO Logging - TP-164154
```
kennyklaviyo, line 71: hmm, philosophical question, should we be including ticket numbers in a public repo?

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
kennyklaviyo: is there a way we can avoid mutation?

kennyklaviyo: generally thats ripe for bug infestations

sdk/core/src/main/java/com/klaviyo/coresdk/config/KlaviyoConfig.kt
evan-masseau: Oh yeah didn't think about that!

sdk/push/src/main/java/com/klaviyo/push/KlaviyoPushService.kt
```diff
@@ -18,12 +18,11 @@ import com.klaviyo.coresdk.utils.KlaviyoPreferenceUtils
  */
 class KlaviyoPushService : FirebaseMessagingService() {
     companion object {
-        internal const val PUSH_TOKEN_PREFERENCE_KEY = "PUSH_TOKEN"
-
-        private const val REQUEST_PUSH_KEY = "\$android_tokens"
+        internal const val PUSH_TOKEN_KEY = "push_token"
+        private const val PUSH_TOKEN_APPEND_KEY = "\$android_tokens"
```
kennyklaviyo, line 22: later, we might want to consolidate these event keys into some enum somewhere else maybe, future tech debt to clean up

sdk/push/src/main/java/com/klaviyo/push/KlaviyoPushService.kt
```diff
@@ -18,12 +18,11 @@ import com.klaviyo.coresdk.utils.KlaviyoPreferenceUtils
  */
 class KlaviyoPushService : FirebaseMessagingService() {
```
kennyklaviyo, line 19: did we want to make this open?

sdk/push/src/main/java/com/klaviyo/push/KlaviyoPushService.kt
```diff
@@ -34,70 +33,69 @@ class KlaviyoPushService : FirebaseMessagingService() {
...
 
-            KlaviyoPreferenceUtils.writeStringPreference(PUSH_TOKEN_PREFERENCE_KEY, pushToken)
+            Registry.dataStore.store(PUSH_TOKEN_KEY, pushToken)
         }
 
         /**
-         * Returns the current push token that we have stored on this device
+         * Retrieves the device FCM push token stored on this device
          *
          * @return The push token we read from the shared preferences
```
kennyklaviyo, line 46: update string, `read from the datastore`

sdk/core/src/main/java/com/klaviyo/coresdk/lifecycle/LifecycleMonitor.kt
evan-masseau: I was a little torn on that, I had started with `onStopped`. This class contains several `onXyz` methods that are invoked by the system after this object is with `ActivityLifecycleCallbacks`, whereas this method is for a _different_ service to register its closure as a callback to an event broadcast from _this_ object... Maybe I'm just overthinking it and the use case is clear from the arguments though.

sdk/push/src/main/java/com/klaviyo/push/KlaviyoPushService.kt
```diff
@@ -34,70 +33,69 @@ class KlaviyoPushService : FirebaseMessagingService() {
...
          */
-        fun getCurrentPushToken(): String {
-            return KlaviyoPreferenceUtils.readStringPreference(PUSH_TOKEN_PREFERENCE_KEY) ?: ""
+        internal fun getPushToken(): String {
+            return Registry.dataStore.fetch(PUSH_TOKEN_KEY) ?: ""
```
kennyklaviyo, line 49: should we handle null more explicitly downstream?

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: Perhaps that would be best... I could make mergeProfile return `getAsProfile` instead, so it is a new profile object.

sdk/push/src/main/java/com/klaviyo/push/KlaviyoPushService.kt
kennyklaviyo: looks like we only using it in `openedPush` to include push tokens, to slim the payload, shouldnt we no include the `PUSH_TOKEN_KEY` if there is no token?

evan-masseau: True, and I think this will go back to being a public function again too, and returning null is a clearer empty state than just empty string.

sdk/push/src/main/java/com/klaviyo/push/KlaviyoPushService.kt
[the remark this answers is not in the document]
evan-masseau: `PUSH_TOKEN_KEY` is the key for persistent store, `PUSH_TOKEN_APPEND_KEY` is the key for the API payload in the identify call.

sdk/core/src/main/java/com/klaviyo/coresdk/config/SystemClock.kt
```diff
@@ -0,0 +1,29 @@
...
+    override fun currentTimeAsString(): String {
+        return format.format(Date(currentTimeMillis()))
+    }
+
+    override fun schedule(delay: Long, task: () -> Unit): Clock.Cancellable = Timer()
+        .schedule(delay) { task() }.let {
+            object : Clock.Cancellable {
+                override fun cancel() = it.cancel()
+            }
+        }
```
kennyklaviyo, line 28: since task is a lambda, you can just do `.schedule(delay, task)` no need for the extra `{ task() }`

sdk/core/src/main/java/com/klaviyo/coresdk/lifecycle/KlaviyoLifecycleMonitor.kt
```diff
@@ -0,0 +1,44 @@
...
+ */
+internal object KlaviyoLifecycleMonitor : LifecycleMonitor, Application.ActivityLifecycleCallbacks {
+
+    private var activitiesActive = 0
+
+    private var whenStopped = mutableListOf<ActivityObserver>()
+
+    override fun whenStopped(observer: ActivityObserver) {
+        whenStopped += observer
+    }
```
kennyklaviyo, line 18: should there be a way to remove an observer? when will this list get cleared?

sdk/core/src/main/java/com/klaviyo/coresdk/lifecycle/KlaviyoLifecycleMonitor.kt
```diff
@@ -0,0 +1,44 @@
...
+import android.app.Activity
+import android.app.Application
+import android.os.Bundle
+
+/**
+ * Service for monitoring the application lifecycle and network connectivity
+ */
+internal object KlaviyoLifecycleMonitor : LifecycleMonitor, Application.ActivityLifecycleCallbacks {
+
+    private var activitiesActive = 0
```
kennyklaviyo, line 12: nit: `activeActivities`

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -0,0 +1,234 @@
...
+        /**
+         * Compare two requests purely by their UUIDs
+         *
+         * @param request1
+         * @param request2
+         * @return
+         */
+        fun equals(request1: KlaviyoApiRequest, request2: KlaviyoApiRequest): Boolean =
+            request1.uuid == request2.uuid
+    }
```
kennyklaviyo, line 113: should this just live inside the overriden equals?

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -0,0 +1,234 @@
...
+        val connection = buildUrlConnection()
+
+        return try {
+            connection.connect()
+            return readResponse(connection)
+        } catch (ex: IOException) {
+            // TODO Logging - TP-164154
+            return null
+        } finally {
+            connection.disconnect()
```
kennyklaviyo, line 149: can an exception be thrown during `connection.connect()` and will that make disconnect throw an exception too?

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -0,0 +1,234 @@
...
+    /**
+     * Reads the response body from the open [HttpURLConnection]
+     * If the request was successful, extracts the response body
+     * If the request was unsuccessful, extracts the error response body
+     *
+     * @return The string extracted from the response body
+     */
+    private fun readResponse(connection: HttpURLConnection): String {
+        val statusCode = connection.responseCode
+        val stream = if (statusCode in HTTP_OK until HTTP_MULT_CHOICE) {
```
kennyklaviyo, line 185: does this perform well? or will it create a list from OK to MULT_CHOICE and iterate until statusCode matches an element in the list?

kennyklaviyo: can we do a comparison check instead?

evan-masseau: Hum I just copied this, I'll take a closer look

sdk/core/src/main/java/com/klaviyo/coresdk/networking/NetworkMonitor.kt
```diff
@@ -0,0 +1,111 @@
...
+ * whenever any internet connectivity change is detected
+ */
+internal object KlaviyoNetworkMonitor : NetworkMonitor {
+
+    private lateinit var networkRequest: NetworkRequest
+
+    private val connectivityManager: ConnectivityManager
+        get() = ContextCompat.getSystemService(
+            Registry.config.applicationContext, ConnectivityManager::class.java
+        ) as ConnectivityManager
```
kennyklaviyo, line 46: can do just do `getSystemService<ConnectivityManager>()` since we support at least 23, ContextCompat isnt necessary (impl of it checks for >= 23)

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -1,73 +1,143 @@
...
+     * populated from the current internally tracked profile info.
+     *
      * Identify requests track specific properties about a user without triggering an event
      *
-     * @param properties A map of properties that define the user
+     * @param profile A map of properties that define the user
+     * @return
+     */
+    fun setProfile(profile: Profile): Klaviyo = apply {
+        UserInfo.mergeProfile(profile)
```
evan-masseau, line 106: Oh crap this is definitely not right, its going to lose all but the most recent function call's profile attributes. 
It needs a pending profile object to merge into as multiple calls are made.

evan-masseau: Fixed and added tests of the resultant API call from the debouncing

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/KlaviyoApiRequest.kt
evan-masseau: Yeah i think you're right this is basically like saying `if (200 in [1,2,3,4,5,6,7...])` using greater than/less than makes more sense. Syntactic sugar is so tempting.

evan-masseau: Huh https://stackoverflow.com/questions/69171532/is-kotlins-range-checking-is-efficent 
The IDE tells me to use range check too... and who am I to disobey android studio

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/KlaviyoApiRequest.kt
evan-masseau: Great question, I'm going to put a TODO in, I wanted to test the exception handling with the tester app too.

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/KlaviyoApiRequest.kt
evan-masseau: Yep, I made it static for a unit test... but I can just test the overridden equals directly 🤦

sdk/push/src/main/java/com/klaviyo/push/KlaviyoPushService.kt
```diff
@@ -34,70 +33,69 @@ class KlaviyoPushService : FirebaseMessagingService() {
...
-         * TODO customer and event properties should both be optional, once SDK is tracking customer state internally
+         * @param payload The data attributes of the push notification payload
+         */
+        internal fun openedPush(payload: Map<String, String>) {
+            if (!isKlaviyoPush(payload)) return // Track only pushes originating from klaviyo
+
+            val event = Event().also {
+                payload.forEach { (k, v) -> it.setProperty(k, v) }
+                it.setProperty(PUSH_TOKEN_KEY, getPushToken())
+            }
```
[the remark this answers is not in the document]
evan-masseau: Sure, I'll do that when I update `Event` to encapsulate type

sdk/core/src/main/java/com/klaviyo/coresdk/lifecycle/KlaviyoLifecycleMonitor.kt
evan-masseau: Took me like 4 tries to even see the difference there. But yeah thats better.

sdk/core/src/main/java/com/klaviyo/coresdk/lifecycle/KlaviyoLifecycleMonitor.kt
evan-masseau: Yes... I skipped that because the observers we need will live for the life of the application, but yes I should at least write the remove method.

sdk/core/src/main/java/com/klaviyo/coresdk/config/SystemClock.kt
evan-masseau: I might be missing something, but that doesn't seem to work. `Timer.schedule` expects a `TimerTask.()->Unit` (what??) so my lambda doesn't convert without using the closure.

sdk/push/src/main/java/com/klaviyo/push/KlaviyoPushService.kt
evan-masseau: I thought we wound up on no, but I'll check the discussion docs again.

sdk/push/src/main/java/com/klaviyo/push/KlaviyoPushService.kt
evan-masseau: Agreed, I was also thinking this might get revisited if we try to make `dataStore` internal and instead have the `push` package use some other method.

sdk/push/src/main/java/com/klaviyo/push/KlaviyoPushService.kt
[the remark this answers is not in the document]
evan-masseau: Might be a nice constructor for Profile too, maybe I can make it part of the abstract class.

sdk/core/src/main/java/com/klaviyo/coresdk/lifecycle/LifecycleMonitor.kt
evan-masseau: Nvm I decided I am indeed overthinking it.

sdk/push/src/main/java/com/klaviyo/push/KlaviyoPushService.kt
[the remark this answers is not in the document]
evan-masseau: Updated the models with some new constructors that take a properties dict

sdk/core/src/main/java/com/klaviyo/coresdk/lifecycle/KlaviyoLifecycleMonitor.kt
evan-masseau: Come to think of it, maybe the listeners should be removed or paused when the activity sleeps... I'll need to read up on lifecycle a little more.

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/KlaviyoApiRequest.kt
kennyklaviyo: oooo very nice

sdk/core/src/main/java/com/klaviyo/coresdk/model/UserInfo.kt
```diff
@@ -46,31 +46,42 @@ internal object UserInfo {
...
         }
 
-    fun reset() {
+    /**
+     * Updates [UserInfo] identifiers in state from a given [Profile] object
+     *
+     * @param profile
+     */
+    fun updateFromProfile(profile: Profile) = apply {
+        externalId = profile.identifier ?: externalId
```
kennyklaviyo, line 63: i think we should rename `identifier` to `externalId`

evan-masseau: thought i did! I missed one, thanks.

evan-masseau: Oh i see its in the profile class's attribute yeah

sdk/push/src/main/java/com/klaviyo/push/KlaviyoPushService.kt
evan-masseau: Yeah we landed on it being unnecessary to make our class open, when all they'd need to do is call the static methods from their implementation, see readme
### Manual implementation of `FirebaseMessagingService` (Advanced)
If you'd prefer to implement `FirebaseMessagingService` yourself, follow the FCM 
setup docs including referencing your own service class in the manifest.
Then update your implementation of `onNewToken` and `onMessageReceived` as below to communicate 
push tokens and notifications received to the Klaviyo SDK. Note that the launcher activity 
code snippets above are still required.
```kotlin
import com.google.firebase.messaging.RemoteMessage
import com.klaviyo.push.KlaviyoPushService

class YourPushService: FirebaseMessagingService() {
    override fun onNewToken(newToken: String) {
        super.onNewToken(newToken)
        KlaviyoPushService.setPushToken(newToken)
    }

        KlaviyoPushService.handlePush(message)
    }
}
```

sdk/core/src/main/java/com/klaviyo/coresdk/config/SystemClock.kt
kennyklaviyo: interesting, yeah just played around locally and seems the only way is to change the type:
```
fun schedule(delay: Long, task: TimerTask.() -> Unit) =
    Timer().schedule(delay=delay, task)
```
but you can still schedule with a lambda as before:
```
schedule(10L) { // this context is a TimerTask
    something()
}
```
that said, not sure if we want the interface exposing the `TimerTask` to the lambda so this is fine

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoApiClient.kt
```diff
@@ -0,0 +1,180 @@
...
+import com.klaviyo.coresdk.networking.requests.KlaviyoApiRequest
+import com.klaviyo.coresdk.networking.requests.TrackApiRequest
+import java.util.concurrent.ConcurrentLinkedQueue
+import org.json.JSONArray
+import org.json.JSONObject
+
+/**
+ * Coordinator of API request traffic
+ */
+internal object KlaviyoApiClient : ApiClient {
```
ajaysubra, line 20: The comments are great to get a sense of what's going on! Thanks!

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoApiClient.kt
```diff
@@ -0,0 +1,180 @@
...
+    internal const val QUEUE_KEY = "klaviyo_api_request_queue"
+
+    private val handlerThread = HandlerUtil.getHandlerThread(KlaviyoApiClient::class.simpleName)
+    private var handler: Handler? = null
+    private var apiQueue = ConcurrentLinkedQueue<KlaviyoApiRequest>()
```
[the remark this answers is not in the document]
evan-masseau: My impression is no, but I haven't gone deep on threading yet. I copied this handler/looper stuff forward from existing code. If we refactor this in the future, I believe co-routines would be the modern way to handle this, but I don't think this would be considered deprecated or even outdated either.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -1,73 +1,146 @@
...
+     * Assigns an external ID to the current internally tracked profile
+     *
+     * The SDK keeps track of current profile details to
+     * build analytics requests with profile identifiers
+     *
+     * This should be called whenever the active user in your app changes
+     * (e.g. after a fresh login)
+     *
+     * @param id Phone number for active user
+     * @return
```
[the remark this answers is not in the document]
evan-masseau: Ah copy pasta error thx for the close read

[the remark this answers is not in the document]
evan-masseau: In general yes the comment is applicable -- the external ID should be the app developers' unique identifier of the user, so just like email or phone, if a user logs in, the external ID, if you're using it, would be set at that time.

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoApiClient.kt
```diff
@@ -0,0 +1,180 @@
...
+        }
+    }
+
+    /**
+     * Initialize the network queue batching with a thread to run on
+     */
+    private fun initBatch() {
+        if (!handlerThread.isAlive) {
+            handlerThread.start()
+            handler = HandlerUtil.getHandler(handlerThread.looper)
```
ndurell, line 100: hmm...surprised a little by this. I thought maybe there were newer concurrency mechanisms in kotlin.

ndurell: Also does this support cancellation?

evan-masseau: There is, coroutines, it could be on our later roadmap to update to use coroutines, but I didn't think this approach was outdated or problematic so I kept it as is (this was in the 2020 code)

sdk/core/src/test/java/com/klaviyo/coresdk/networking/requests/IdentifyApiRequestTest.kt
```diff
@@ -0,0 +1,146 @@
...
+
+    override fun setup() {
+        super.setup()
+
+        // Start all tests with an empty profile and base64 mock for encoding it
+        profile.setAnonymousId(ANON_ID)
+        withMockBase64("{\"properties\":{\"\$anonymous\":\"$ANON_ID\"},\"token\":\"$API_KEY\"}")
+    }
+
+    private fun withMockBase64(expectedString: String) {
```
ndurell, line 28: I don't think you need this if you use the post request.

ndurell: Sorry meant to comment on the actual request no the test.

evan-masseau: I do have a ticket to update to the V3 API, is that what you mean?

evan-masseau: Oh christ you mean there's a v2 `identify` endpoint that just takes POST with a body?? I wasted like half a day figuring out how to write those stupid base64 mocked tests.

evan-masseau: Sorry, frustration directed at myself for not looking at the docs closer, not you.

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/IdentifyApiRequest.kt
```diff
@@ -0,0 +1,30 @@
...
+    override var query: Map<String, String> = profile.let { properties ->
+        val data = JSONObject(
+            mapOf(
+                "token" to Registry.config.apiKey,
+                "properties" to JSONObject(properties.toMap())
+            )
+        ).toString()
+
+        mapOf(
+            "data" to Base64.encodeToString(data.toByteArray(), Base64.NO_WRAP)
```
ndurell, line 27: Yah if you make it a post you don't this base64 stuff anymore.

ndurell: Or I guess just use v3.

evan-masseau: Thanks, I'll put a comment in the V3 ticket to both update profile call to v3 but also change the legacy call to use POST.

build.gradle
```diff
@@ -38,21 +38,7 @@ task clean(type: Delete) {
...
-        reportUndocumented = true
-        skipEmptyPackages = true
-
-        includes = ["README.md"]
-
-        jdkVersion = 11
-    }
+dokkaHtmlMultiModule {
+    outputDirectory = file("${buildDir}/../docs")
+    includes.from("README.md")
```
ndurell, line 43: Just curious what the includes does ?

evan-masseau: I think its like `include` in php, it just appends that file's content into the home page of the output? something like that.

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoApiClient.kt
evan-masseau: Pretty sure it supports cancellation yes, this is the main reason i reopened the network layer ticket though, I hadn't fully finished making this ApiClient class respond to lifecycle events properly.

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoApiClient.kt
```diff
@@ -0,0 +1,180 @@
...
+            if (force || getQueueSize() >= flushDepth || queueTimePassed >= flushInterval) {
+                while (!apiQueue.isEmpty()) {
+                    val apiRequest = apiQueue.poll()
+                    apiRequest?.apply {
+                        send()
+                        Registry.dataStore.clear(uuid)
+                    }
+                }
+
+                Registry.dataStore.clear(QUEUE_KEY)
```
ndurell, line 164: What thread is this work performed on ? Can something be added to the queue after you exit the while loop but before this is called.

evan-masseau: As I understand it, everything in `NetworkRunnable` will be running in our BG thread. Concurrency of the persisted API queue is definitely a concern worth investigating though

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoApiClient.kt
```diff
@@ -0,0 +1,180 @@
...
+         * @return Whether the request queue was emptied or not
+         */
+        private fun flushQueue(force: Boolean = false): Boolean {
+            val queueTimePassed = Registry.clock.currentTimeMillis() - queueInitTime
+
+            if (force || getQueueSize() >= flushDepth || queueTimePassed >= flushInterval) {
+                while (!apiQueue.isEmpty()) {
+                    val apiRequest = apiQueue.poll()
+                    apiRequest?.apply {
+                        send()
```
ndurell, line 159: Do you need to handle the return value here? Seems like sometimes you return null like when the network is disconnected. Maybe this is for a future iteration?

evan-masseau: ooh shoot yeah good call this will clear the API queue even if the requests don't send!

sdk/core/src/main/java/com/klaviyo/coresdk/model/UserInfo.kt
```diff
@@ -0,0 +1,99 @@
...
+            field
+        }
+
+    var phoneNumber: String = ""
+        set(value) {
+            field = value
+            Registry.dataStore.store(PHONE_NUMBER.name, value)
+        }
+        get() = field.ifEmpty {
+            field = Registry.dataStore.fetch(PHONE_NUMBER.name) ?: ""
```
ndurell, line 40: Feels like a bit of duplicated code is there a way to maybe update this in one spot and then persist the whole thing or at least whatever changed since last time. Maybe though you are trying to optimize how often things are written to disk?

evan-masseau: Could you elaborate? Do you mean how each property is written to disk individually?

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoApiClient.kt
evan-masseau: If I just write the content of the concurrent queue here, instead of simply clearing, that should be safe...

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoApiClient.kt
evan-masseau: This is the other reason I'd re-opened the network layer ticket BTW, because I'm totally missing things like retrying rate limit errors or handling other error types. But I didn't even notice that I'm straight up throwing requests while offline 🤦

sdk/core/src/main/java/com/klaviyo/coresdk/model/UserInfo.kt
evan-masseau: @ndurell I made a change here to pull out some boilerplate and fix the fact that I'm storing an empty string instead of _clearing_ the key in the data store... did this address your comment at all too?

ndurell: I don't think so the main concern for me is it mixes couple the concerns of updating and getting state with persistence. I had thought in a previous chat you had mentioned something would observe changes to state and persist them. If we add a new field you will have to repeat this pattern again. I think though this class will probably not change very much over time so maybe it's not a big deal right now. Definitely not blocker!

evan-masseau: Ah I see what you're saying. Yeah when I realized there's only 4 things to persist about a profile, I didn't take on the added complication of a single state object that gets persisted altogether. I think if we later have more to persist in state, I'd consider changing this pattern.

sdk/core/src/main/java/com/klaviyo/coresdk/model/UserInfo.kt
```diff
@@ -0,0 +1,103 @@
...
+        email = profile.email ?: email
+        phoneNumber = profile.phoneNumber ?: phoneNumber
+    }
+
+    /**
+     * Reset all user identifiers to defaults
+     * which will cause a new anonymous ID to be generated
+     */
+    fun reset() = apply {
+        externalId = ""
```
ndurell, line 77: This will call the setters above and do persistence stuff yah?

evan-masseau: Yup!

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: Resolving, I made some changes here including to add the debouncing but also took out the merge/mutation thing.

sdk/push/src/main/java/com/klaviyo/push/KlaviyoPushService.kt
evan-masseau: This can get updated when I split the project up into more modules.

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/KlaviyoApiRequest.kt
evan-masseau: Most samples I see just put this in a finally... but I'm paranoid enough that I'm going to put an extra try/catch around `.disconnect()`

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoApiClient.kt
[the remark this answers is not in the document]
evan-masseau: Ah I was wrong, there are thread priorities 1-10. Default seems to be fine for us though.

sdk/core/src/main/java/com/klaviyo/coresdk/lifecycle/KlaviyoLifecycleMonitor.kt
evan-masseau: Added removal option.

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoApiClient.kt
ndurell: I had realizations like when I was working on the iOS SDK. Don't beat yourself up there's a lot to think about. :)

evan-masseau: Thanks for the reassurance, its definitely more complicated than it seems from the surface!

evan-masseau: Resolving, see PR #33

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoApiClient.kt
evan-masseau: #33

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoApiClient.kt
evan-masseau: Does support cancelling "messages" see #33

sdk/core/src/main/java/com/klaviyo/coresdk/model/Profile.kt
```diff
@@ -0,0 +1,92 @@
...
+    object EXTERNAL_ID : ProfileKey("\$external_id")
+    object EMAIL : ProfileKey("\$email")
+    object PHONE_NUMBER : ProfileKey("\$phone_number")
+    internal object ANONYMOUS : ProfileKey("\$anonymous")
+
+    // Personal information
+    object FIRST_NAME : ProfileKey("\$first_name")
+    object LAST_NAME : ProfileKey("\$last_name")
+    object TITLE : ProfileKey("\$title")
+    object ORGANIZATION : ProfileKey("\$organization")
```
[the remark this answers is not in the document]
evan-masseau: Will do, I'm gonna do that in another PR with the v3 API work.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -1,73 +1,146 @@
...
-     * Creates an identify request for the Klaviyo APIs and begins processing.
+     * Assigns an external ID to the current internally tracked profile
+     *
+     * The SDK keeps track of current profile details to
+     * build analytics requests with profile identifiers
+     *
+     * This should be called whenever the active user in your app changes
+     * (e.g. after a fresh login)
+     *
+     * @param id Phone number for active user
```
jordan-griffin, line 69: Think you still have a copy/paste issue here since its mentioning phone number as a description for the id param

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -1,73 +1,146 @@
...
+     * The SDK keeps track of current profile details to
+     * build analytics requests with profile identifiers
+     *
+     * This should be called whenever the active user in your app changes
+     * (e.g. after a fresh login)
+     *
+     * @param phoneNumber Phone number for active user
+     */
+    fun setPhoneNumber(phoneNumber: String): Klaviyo =
+        this.setProfile(Profile().setPhoneNumber(phoneNumber))
```
jordan-griffin, line 58: nit, but I still find this a little hard to read. When I set set profile taking in a fresh `Profile` object which just has a phone number set, I imagine that I'm overwriting all profile data in storage. I know thats not what the function does, but when I read this as code it just fails to click in my mind until I dig into the `UserInfo` class and see what the logic actually does and I worry a little that it would have the same effect on some users of our SDK.

jordan-griffin: I don't think it's a critical or blocking issue of any sort cause its something that could be pretty easily changed without affecting the contract between our SDK and the developer's app. It's all internal details.

But ya I just think maybe something like this might be more immediately consumable.
```kotlin
fun setPhoneNumber(phoneNumber: String): Klaviyo =
    UserInfo.phoneNumber = phoneNumber
    identify()
```

Where `identify` is just a call to enqueue the identify request after building the current profile from whats in `UserInfo` exactly as you currently do.

jordan-griffin: Honestly it really doesn't affect much but thought I'd put down my rambled thoughts on it

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -1,73 +1,146 @@
...
+     */
+    fun setProfile(profile: Profile): Klaviyo = apply {
+        // Update user identifiers in state
+        UserInfo.updateFromProfile(profile)
+
+        // Start or update a pending profile object for API call
+        pendingProfile = UserInfo.getAsProfile().merge(pendingProfile?.merge(profile) ?: profile)
+
+        timer?.cancel()
+        timer = Registry.clock.schedule(Registry.config.debounceInterval.toLong()) {
```
jordan-griffin, line 113: Another nit: but thoughts on ripping the actual logic to enqueue the indentify request and process the debouncing into a private function? It just seems like it could be kept decoupled from the function of `setProfile`.

Otherwise I think this debounce logic on the requests looks great!

sdk/core/src/main/java/com/klaviyo/coresdk/model/Profile.kt
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+    internal fun setAnonymousId(anonymousId: String) = apply { this.anonymousId = anonymousId }
+    internal var anonymousId: String?
+        get() = (this[ProfileKey.ANONYMOUS]) as String
+        set(value) {
+            this[ProfileKey.ANONYMOUS] = value
+        }
+
+    // TODO internal?
```
jordan-griffin, line 43: What are the cons of leaving this exposed?

evan-masseau: Just a potential confusion, it really shouldn't be used for anything but a push token. I've actually taken this out while working on the V3 profile API though because V3 api doesn't even recognize `$append`.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: In a related vein I actually worry a bit that `setProfile` as a public method is confusing, because like you said it sounds like it would overwrite all profile data, not just set the values you're passing in.

sdk/core/src/main/java/com/klaviyo/coresdk/model/Profile.kt
jordan-griffin: Cool ok, so sounds inconsequential to this PR!

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: Yeah I'm down for that.

sdk/core/src/main/java/com/klaviyo/coresdk/model/Profile.kt
evan-masseau: Yup, PR for V3 coming shortly and I think you'll see what I mean. I just put the handling of `$append` straight into the push token API Request class.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: Fixed in https://github.com/klaviyo/klaviyo-android-sdk/pull/35

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: Thanks for this feedback, I agree it wasn't very readable. Check out this PR https://github.com/klaviyo/klaviyo-android-sdk/pull/35 I think it makes it clearer.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: See https://github.com/klaviyo/klaviyo-android-sdk/pull/35 I extracted it which actually helped see how to make the other setters's code more readable

sdk/core/src/main/java/com/klaviyo/coresdk/model/Profile.kt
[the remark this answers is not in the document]
evan-masseau: See #34

sdk/core/src/main/java/com/klaviyo/coresdk/model/UserInfo.kt
evan-masseau: Resolving this per discussion on standup

jordan-griffin (approved):
Between this, #34. and #35 I think my concerns are all addressed!

If I have time I'll try to take one more sweep through to look at your test coverage properly

PR #33, opened 2023-02-15
Network layer improvement

evan-masseau:
# Purpose

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

Revisited network client because my earlier PR was missing error handling, retries and proper lifecycle behaviors.

## Changelog / Code Overview

- Fixed lifecycle event usage: Cancel network bg job on all activities stop, stop bg job on network offline, restart job and flush immediately when network restored.
- Added JSON parser error handling
- Added API request error handling including rate limit retries
- Added config for max number of retries
- Changed to ConcurrentDeque so I can re-queue requests in their original place... I considered using peek+remove instead, was a little worried of potential for bugs
- Test coverage of all the above

## Test Plan

Unit tests

## What to look for

Input wanted: sanity check of the default config values that I updated in here... here's my logic:
- Network timeout: 10s (was 500ms), this seems like a decent balance for slow/flaky network connections. It is also the default for popular lib OKHTTP
- Flush interval: 30s (was 60s) even under heavy usage, this shouldn't mean radios get time to sleep. Also means our retry with exp back off would result in 30s, 1min, 3min, 12min which seems good?
- Flush depth: 100 (was 20) I just figure this lets the timer be primarily in control unless a large number of requests pile up 
- Max retries: 4 (new variable) Total of 5 attempts, perhaps too high but conservative since the only thing we can do after that is throw the request away really. It is an extreme edge case. 

## Related Issues/Tickets

Revisited https://klaviyo.tpondemand.com/entity/156424-android-sdk-network-layer

### DIFF
--- a/sdk/core/src/main/java/com/klaviyo/coresdk/config/KlaviyoConfig.kt
+++ b/sdk/core/src/main/java/com/klaviyo/coresdk/config/KlaviyoConfig.kt
@@ -28,10 +28,31 @@ class MissingPermission(permission: String) : Exception("You must declare $permi
  * Stores all configuration related to the Klaviyo Android SDK.
  */
 object KlaviyoConfig : Config {
-    private const val DEBOUNCE_INTERVAL: Int = 500
-    private const val NETWORK_TIMEOUT_DEFAULT: Int = 500
-    private const val NETWORK_FLUSH_INTERVAL_DEFAULT: Int = 60000
-    private const val NETWORK_FLUSH_DEPTH_DEFAULT: Int = 20
+    /**
+     * Default value: Debounce time for fluent profile setter methods
+     */
+    private const val DEBOUNCE_INTERVAL: Int = 100
+
+    /**
+     * Default value: Network request timeout duration
+     */
+    private const val NETWORK_TIMEOUT_DEFAULT: Int = 10_000
+
+    /**
+     * Default value: Interval between flushing network queue.
+     * Also used as the basis for exponential backoff when retrying requests
+     */
+    private const val NETWORK_FLUSH_INTERVAL_DEFAULT: Int = 30_000
+
+    /**
+     * Default value: How many API requests can be enqueued before flush
+     */
+    private const val NETWORK_FLUSH_DEPTH_DEFAULT: Int = 100
+
+    /**
+     * Default value: How many retries to allow an API request before permanent failure
+     */
+    private const val NETWORK_MAX_RETRIES_DEFAULT: Int = 4
 
     override val baseUrl: String = BuildConfig.KLAVIYO_SERVER_URL
     override lateinit var apiKey: String private set
@@ -44,6 +65,8 @@ object KlaviyoConfig : Config {
         private set
     override var networkFlushDepth = NETWORK_FLUSH_DEPTH_DEFAULT
         private set
+    override var networkMaxRetries = NETWORK_MAX_RETRIES_DEFAULT
+        private set
 
     /**
      * Nested class to enable the builder pattern for easy declaration of custom configurations
@@ -55,6 +78,7 @@ object KlaviyoConfig : Config {
         private var networkTimeout: Int = NETWORK_TIMEOUT_DEFAULT
         private var networkFlushInterval: Int = NETWORK_FLUSH_INTERVAL_DEFAULT
         private var networkFlushDepth = NETWORK_FLUSH_DEPTH_DEFAULT
+        private var networkMaxRetries = NETWORK_MAX_RETRIES_DEFAULT
 
         override fun apiKey(apiKey: String) = apply {
             this.apiKey = apiKey
@@ -96,6 +120,14 @@ object KlaviyoConfig : Config {
             }
         }
 
+        override fun networkMaxRetries(networkMaxRetries: Int) = apply {
+            if (networkMaxRetries >= 0) {
+                this.networkMaxRetries = networkMaxRetries
+            } else {
+                // TODO Logging
+            }
+        }
+
         override fun build(): Config {
             if (apiKey.isEmpty()) {
                 throw MissingAPIKey()
@@ -118,6 +150,7 @@ object KlaviyoConfig : Config {
             KlaviyoConfig.networkTimeout = networkTimeout
             KlaviyoConfig.networkFlushInterval = networkFlushInterval
             KlaviyoConfig.networkFlushDepth = networkFlushDepth
+            KlaviyoConfig.networkMaxRetries = networkMaxRetries
 
             return KlaviyoConfig
         }
--- a/sdk/core/src/main/java/com/klaviyo/coresdk/config/KlaviyoConfig.kt
+++ b/sdk/core/src/main/java/com/klaviyo/coresdk/config/KlaviyoConfig.kt
@@ -29,28 +29,41 @@ class MissingPermission(permission: String) : Exception("You must declare $permi
  */
 object KlaviyoConfig : Config {
     /**
-     * Default value: Debounce time for fluent profile setter methods
+     * Debounce time for fluent profile setter methods
+     *
+     * Reasoning: The debounce is only intended to merge chained profile updates into one API call
      */
     private const val DEBOUNCE_INTERVAL: Int = 100
 
     /**
-     * Default value: Network request timeout duration
+     * Network request timeout duration
+     *
+     * Reasoning: Ten seconds accommodates a reasonable network latency
+     * On a perfect connection our API calls should take around 1 second
      */
     private const val NETWORK_TIMEOUT_DEFAULT: Int = 10_000
 
     /**
-     * Default value: Interval between flushing network queue.
-     * Also used as the basis for exponential backoff when retrying requests
+     * Interval between flushing network queue, and the basis for retry with exponential backoff
+     *
+     * Reasoning: A 30 second interval should give radios time to go back to sleep between batches,
+     * four retries with a typical backoff pattern would then be 30s, 60s, 3m, 12m.
      */
     private const val NETWORK_FLUSH_INTERVAL_DEFAULT: Int = 30_000
 
     /**
-     * Default value: How many API requests can be enqueued before flush
+     * How many API requests can be enqueued before flush
+     *
+     * Reasoning: The goal of depth control is to limit duration that radios are active
+     * if a typical request takes 1-3 seconds, this should ideally limit us to 30-90 seconds
      */
-    private const val NETWORK_FLUSH_DEPTH_DEFAULT: Int = 100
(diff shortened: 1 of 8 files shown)


sdk/core/src/main/java/com/klaviyo/coresdk/config/KlaviyoConfig.kt
```diff
@@ -28,10 +28,31 @@ class MissingPermission(permission: String) : Exception("You must declare $permi
...
-    private const val NETWORK_FLUSH_DEPTH_DEFAULT: Int = 20
+    /**
+     * Default value: Debounce time for fluent profile setter methods
+     */
+    private const val DEBOUNCE_INTERVAL: Int = 100
+
+    /**
+     * Default value: Network request timeout duration
+     */
+    private const val NETWORK_TIMEOUT_DEFAULT: Int = 10_000
```
jordan-griffin, line 39: Is it shameful that I never knew you could do this underscore style with integers in Kotlin? That's super nice for readability

evan-masseau: Right?? I had no idea I saw it when I Googled ok https default timeout

sdk/core/src/main/java/com/klaviyo/coresdk/config/KlaviyoConfig.kt
```diff
@@ -28,10 +28,31 @@ class MissingPermission(permission: String) : Exception("You must declare $permi
...
+
+    /**
+     * Default value: Interval between flushing network queue.
+     * Also used as the basis for exponential backoff when retrying requests
+     */
+    private const val NETWORK_FLUSH_INTERVAL_DEFAULT: Int = 30_000
```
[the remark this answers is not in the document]
evan-masseau: See PR description

sdk/core/src/main/java/com/klaviyo/coresdk/config/KlaviyoConfig.kt
```diff
@@ -28,10 +28,31 @@ class MissingPermission(permission: String) : Exception("You must declare $permi
...
+
+    /**
+     * Default value: How many API requests can be enqueued before flush
+     */
+    private const val NETWORK_FLUSH_DEPTH_DEFAULT: Int = 100
```
jordan-griffin, line 50: nit: having reviewed your PR description this is the only networking constant change I think seems a little strange. It doesn't really affect anything (as you said, your network flush interval will force the queue to flush every 30s anyway). But it just seems super unlikely to me that a single client would ever be building up a queue depth of 100 within 30s. Makes the setting seem very redundant to me, so I guess I'd choose a lower number

sdk/core/src/main/java/com/klaviyo/coresdk/config/KlaviyoConfig.kt
```diff
@@ -28,10 +28,31 @@ class MissingPermission(permission: String) : Exception("You must declare $permi
...
+
+    /**
+     * Default value: How many retries to allow an API request before permanent failure
+     */
+    private const val NETWORK_MAX_RETRIES_DEFAULT: Int = 4
```
jordan-griffin, line 55: Since we're just popping failed requests back on the queue, they'll only really retry every 30s. Seems maybe a little slow to me for a retry backoff since that means a failed request will be tried for 2.5mins total. 


I spose it doesn't matter since none of the requests our SDK makes are time critical. Just a thought anyway

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoApiClient.kt
```diff
@@ -68,55 +71,81 @@ internal object KlaviyoApiClient : ApiClient {
...
+            }
+        }?.forEach { uuid ->
             Registry.dataStore.fetch(uuid)?.let { json ->
-                val jsonObject = JSONObject(json)
-                val request = KlaviyoApiRequest.fromJson(jsonObject)
-                enqueueRequest(request)
+                try {
+                    val request = KlaviyoApiRequest.fromJson(JSONObject(json))
+                    if (!apiQueue.contains(request)) {
+                        apiQueue.offer(request)
```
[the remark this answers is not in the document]
jordan-griffin: Or is it that we always keep the request in store till send completes, rather than when placed back on the queue?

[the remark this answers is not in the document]
jordan-griffin: If the latter, it would explain why we are doing a `persistQueue` call below I guess

sdk/core/src/main/java/com/klaviyo/coresdk/config/KlaviyoConfig.kt
evan-masseau: Yeah good point that is way too high. For background my level of though here was: "lets see what noah did... 200, hm that sounds high I'll do half that"

Thinking it through more... The goal of depth control should be to prevent keeping the radios on for too _long_ (while the interval is to prevent turning on too _frequently_). If we figure a typical request takes ~1 second, maybe 25 is a good limit.

cc @ndurell

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoApiClient.kt
```diff
@@ -125,48 +154,56 @@ internal object KlaviyoApiClient : ApiClient {
...
-            val emptied = flushQueue(force)
-            if (!emptied) {
-                handler?.postDelayed(this, flushInterval)
-            }
-        }
+        private val flushDepth: Int = Registry.config.networkFlushDepth
 
         /**
-         * Empties all queued requests by instantly sending each request to Klaviyo
+         * Sends all requests serially in queued order until
```
jordan-griffin, line 165: until what?

evan-masseau: Always leave the audience wanting more

sdk/core/src/main/java/com/klaviyo/coresdk/config/KlaviyoConfig.kt
jordan-griffin: Very well put. I like the conclusion a lot!

sdk/core/src/main/java/com/klaviyo/coresdk/config/KlaviyoConfig.kt
jordan-griffin: Ah yes I see it now, nice!

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoApiClient.kt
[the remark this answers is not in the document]
evan-masseau: Basically the latter. The reason for calling `persistQueue` here is just in case the process of restoring the queue into memory causes any change, such as parsing errors.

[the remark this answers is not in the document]
evan-masseau: I figured this way we're far less likely to lose API requests due to unlucky timing if for example the app quit/crashed and we didn't get the opportunity to write our in memory queue into persistent store before shutdown.

sdk/core/src/main/java/com/klaviyo/coresdk/config/KlaviyoConfig.kt
ndurell: Sorry not sure I have this setting. What does this do exactly?

evan-masseau: I believe this is equivalent of `let MAX_QUEUE_SIZE = 200`

evan-masseau: Oh sorry no it isn't... `MAX_QUEUE_SIZE` is like a guard against memory overflow on iOS. You're right you don't have this setting that I can see. 
This setting triggers the queue to be flushed once it reaches a certain _size_.

evan-masseau: Gee maybe I need a max queue size too...

evan-masseau: This is the iOS code: 
```    
    mutating func enqueueRequest(request: KlaviyoAPI.KlaviyoRequest) {
        guard queue.count + 1 < MAX_QUEUE_SIZE else {
            return
        }
        queue.append(request)
    }
```
So, appears to just drop the new request completely. We're definitely getting deep into edge case territory here but I wonder if the way to go would be to persist an overflow queue straight to disk, you can put a lot more on disk than you can just hold in memory.

evan-masseau: I'm thinking this might be another post-GA refinement item to circle back to, same as adding some additional special error handling.

jordan-griffin (approved):
I think this looks good and addresses my concerns. Would definitely encourage a second review though! Bet I missed some things

sdk/core/src/main/java/com/klaviyo/coresdk/config/KlaviyoConfig.kt
kennyklaviyo: python too

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoApiClient.kt
```diff
@@ -68,55 +71,81 @@ internal object KlaviyoApiClient : ApiClient {
...
+                    }
+                } catch (exception: JSONException) {
+                    Registry.dataStore.clear(uuid)
+                }
             }
         }
+
+        if (apiQueue.isNotEmpty()) initBatch()
+
+        persistQueue()
```
kennyklaviyo, line 114: is it necessary to write to disk right after reading from disk?

kennyklaviyo (approved):

sdk/core/src/main/java/com/klaviyo/coresdk/networking/KlaviyoApiClient.kt
evan-masseau: I figured it just persists any mutations that had to be made in case of a JSON parse error above. But I guess it should be conditional, only do this if we know something was changed, otherwise it is a waste.

PR #34, opened 2023-02-16
V3 Profile endpoint and re-purpose V2 Profile to be for Push Token only

evan-masseau:
# Purpose

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

Updated our Profile API call to V3, repurposed the V2 endpoint to be just for push token registration.

This wound up touching more than I expected because of new and renamed keys. I also noticed some mistakes in the V3 Event api request that I corrected in the process. Cleaned up API unit tests a little more too.

Just once I want to be able to say "gee this affected fewer files than I expected!"

## Test Plan

Considerable unit tests, comparison against API docs and Postman, and most importantly tested against local-klaviyo using the tester app. All 3 endpoints are confirmed working now. 

## What to look for

Let me know what you think about the ProfileKey handling, its really funky how V2 Profile and V3 Event use $-prefixed keys for profile, but V3 profile does not. 
Also input welcome on how I did data-structure translation in the API objects.

## Related Issues/Tickets

https://klaviyo.tpondemand.com/entity/160252-android-sdk-should-use-v3-apis


sdk/core/src/main/java/com/klaviyo/coresdk/model/Profile.kt
```diff
@@ -35,17 +33,11 @@ class Profile(properties: Map<ProfileKey, Serializable>?) :
...
     internal var anonymousId: String?
-        get() = (this[ProfileKey.ANONYMOUS]) as String
+        get() = (this[ProfileKey.ANONYMOUS_ID]) as String
         set(value) {
-            this[ProfileKey.ANONYMOUS] = value
+            this[ProfileKey.ANONYMOUS_ID] = value
         }
 
-    // TODO internal?
-    fun addAppendProperty(key: String, value: Serializable) = apply {
```
jordan-griffin, line 44: R.I.P.

sdk/core/src/main/java/com/klaviyo/coresdk/model/Profile.kt
```diff
@@ -54,39 +46,60 @@ class Profile(properties: Map<ProfileKey, Serializable>?) :
...
 
-    // Other
-    internal object APPEND : ProfileKey("\$append")
+    /**
+     * Helper method to translate certain keys to their dollar-prefixed key
+     * This only applies to the identifier keys under certain circumstances
+     *
+     * @return
+     */
+    internal fun specialKey(): String = when (this) {
```
jordan-griffin, line 100: Sorry what's this for? Is it because events API still uses the old $ keys but V3 profiles API does not? So you need to do some simple conversion?

sdk/core/src/main/java/com/klaviyo/coresdk/model/UserInfo.kt
```diff
@@ -10,7 +10,7 @@ import java.util.UUID
 /**
  * Stores information on the currently active user
  */
-internal object UserInfo {
+object UserInfo {
```
jordan-griffin, line 13: Remind me (cause I am sure you told me before), but why are we exposing `UserInfo` now?

sdk/core/src/main/java/com/klaviyo/coresdk/networking/requests/PushTokenApiRequest.kt
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Defines the content of an API request to append a push token to a [Profile]
+ *
+ * Using legacy V2 API until push tokens are supported by a V3 endpoint
+ *
+ * @param token The push token
+ * @param profile Profile identifiers that the token belongs to
+ */
+internal class PushTokenApiRequest(token: String, profile: Profile) : KlaviyoApiRequest(
```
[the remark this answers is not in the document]
evan-masseau: Answered in the `gradle` PR -- my intention is that the push module is just to separate out the FCM dependency, keeping all API logic in the analytics module such that if we add support for another push platform we just have to add another small module but the API code is already written.

sdk/core/src/main/java/com/klaviyo/coresdk/model/Profile.kt
evan-masseau: Exactly right: profile V2 and events V3 and use `$`, profile V3 does not use `$`. I need the conversion for these identifier fields so that the event and push token API calls can get associated with the correct profile. 

I'm going to copy that comment into this spot in the repo

sdk/core/src/main/java/com/klaviyo/coresdk/model/UserInfo.kt
evan-masseau: No worries, I don't think I did. I realized I needed access to `UserInfo.getAsProfile` from the `push` module in order to be able to call `enqueuePushToken(token: String, profile: Profile)` with the correct identifiers. Previously, I was just using `Klaviyo.setProfile`, so didn't need to access `UserInfo` directly. 

Alternatively I could add a method `Klaviyo.setPushToken` but I worried about having that be in the public API of the _analytics_ module and getting mixed up with the push module.

Ultimately, `UserInfo` will be private when we split the `core` and `analytics` into two modules though.

evan-masseau: Actually the more I think about it, it'd make sense to put `setPushToken` in the `analytics` module and call to it from `push-fcm`, as we'd want to use the same method from, say a `push-fireos` module in the future. Indeed if the app developer is implementing their firebase message handler entirely in their own code, they wouldn't need our `push-fcm` package at all.

evan-masseau: If you don't mind, I'm going to put a TODO on this and otherwise handle this question in the gradle organizing branch.

jordan-griffin: All good points, ya I think a TODO is the best call here! Just make sure we properly note down these thoughts into the SDK or a ticket and we'll be able to revisit this easily in the future. It's an abstraction detail so I don't foresee any issues with making changes post-release.

evan-masseau: Cool, added TODO and a comment to the gradle TP ticket about

sdk/core/src/main/java/com/klaviyo/coresdk/model/Profile.kt
evan-masseau: plus side, don't have to think about that TODO now

evan-masseau:
I have now confirmed all 3 endpoints are working by testing with the tester app against local-klaviyo.

evan-masseau:
Discussed with the team, this ones ready to go in

PR #35, opened 2023-02-17
Add profile getters and other minor cleanup of the public API

evan-masseau:
# Purpose

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

Added getter methods for email, phone and external ID to public API per spec (and tests)
Addressed, I think, some code-readability concerns. Extracted the debouncing logic into a private method. 
Made some fixes to the API comments for clarity or just fix out of date comments

## What to look for

Agreed with @jordan-griffin that the readability on some of the profile setter methods wasn't great. I think this makes it more self-evident how these methods behave. 

## Related Issues/Tickets

https://klaviyo.tpondemand.com/entity/164289-expose-profile-getters-from-android-sdk

### DIFF
--- a/sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
+++ b/sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
@@ -18,7 +18,7 @@ object Klaviyo {
      * Configure Klaviyo SDK with your account's public API Key and application context.
      * This must be called to before using any other SDK functionality
      *
-     * @param apiKey - Your Klaviyo account's public API Key
+     * @param apiKey Your Klaviyo account's public API Key
      * @param applicationContext
      * @return
      */
@@ -30,7 +30,7 @@ object Klaviyo {
     }
 
     /**
-     * Assigns an email address to the current internally tracked profile
+     * Assigns an email address to the currently tracked Klaviyo profile
      *
      * The SDK keeps track of current profile details to
      * build analytics requests with profile identifiers
@@ -41,10 +41,15 @@ object Klaviyo {
      * @param email Email address for active user
      * @return
      */
-    fun setEmail(email: String): Klaviyo = this.setProfile(Profile().setEmail(email))
+    fun setEmail(email: String): Klaviyo = this.setProfileAttribute(ProfileKey.EMAIL, email)
 
     /**
-     * Assigns a phone number to the current internally tracked profile
+     * @return The email of the currently tracked profile, if set
+     */
+    fun getEmail(): String? = UserInfo.email.ifEmpty { null }
+
+    /**
+     * Assigns a phone number to the currently tracked Klaviyo profile
      *
      * The SDK keeps track of current profile details to
      * build analytics requests with profile identifiers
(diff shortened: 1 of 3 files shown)


sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -41,10 +41,15 @@ object Klaviyo {
      * @param email Email address for active user
      * @return
      */
-    fun setEmail(email: String): Klaviyo = this.setProfile(Profile().setEmail(email))
+    fun setEmail(email: String): Klaviyo = this.setProfileAttribute(ProfileKey.EMAIL, email)
```
ndurell, line 44: Maybe I should do this too!

evan-masseau: You know what they say, use your own apis

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -53,62 +58,110 @@ object Klaviyo {
...
+     * @return
      */
     fun setPhoneNumber(phoneNumber: String): Klaviyo =
-        this.setProfile(Profile().setPhoneNumber(phoneNumber))
+        this.setProfileAttribute(ProfileKey.PHONE_NUMBER, phoneNumber)
+
+    /**
+     * @return The phone number of the currently tracked profile, if set
+     */
+    fun getPhoneNumber(): String? = UserInfo.phoneNumber.ifEmpty { null }
```
ndurell, line 69: I wonder if we should ignore empty?

ndurell, line 69: Is there a kotlin way to expose a var and just override the setter getter? Might be less code…

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
ndurell: I was waffling on even having the identifiers as keys. I think I kind of like the idea of recording every identifier that is set back the server. Maybe unlikely but I think if I do it this we might lose repeated setting of the same identifier type within the denounce interval.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
```diff
@@ -53,62 +58,110 @@ object Klaviyo {
      * (e.g. after a fresh login)
      *
      * @param phoneNumber Phone number for active user
+     * @return
```
jordan-griffin, line 61: nit: I mean doesn't really matter but add a return description here since you have it on the other setters?

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: Yeah this is kind of a side-effect of how `Profile` model is implemented as just a managed `map` object, so everything needed to be keyed.

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: I went back and forth, but decided that null vs empty string was potentially useful distinction for the app developer.

evan-masseau: Yes, but I feel that keeping setters as functions is more indicative that there is work going on to sync with the API. Also methods are chain-able, assignments would not be (and I'm just a big fan of chaining)
```
Klaviyo.email = "test@gmail.com" // Seems like a simple assignment, obscures the fact that it is going to produce a side effect
```

sdk/core/src/main/java/com/klaviyo/coresdk/Klaviyo.kt
evan-masseau: You know I almost just took those out earlier, they were just auto added by android studio. The return type here is `Klaviyo` (i.e. `this`) to allow chaining methods. Do you think I should say that in the `@return` description of all the setXYZ methods?

jordan-griffin: I don't really care outside of how it affects the auto-generated docs. If you generate with dokka and it looks ok, its fine IMO. Usually I tend to just go overly-verbose when I'm doing docstrings.

evan-masseau: I added @return comments, i agree, especially in this file which is like 99% of our API!

PR #36, opened 2023-02-17
Split up SDK into more gradle modules

evan-masseau:
[DO NOT MERGE]

# Purpose

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [x] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

Restructured the project into 4 gradle modules. Particularly to prevent exposing our core services in our published packages.

## Changelog / Code Overview

Re-arranges the repo into 4 modules
- Core: for internal services like DataStore (unpublished)
- Analytics: base package for Klaviyo client API (published)
- Push-FCM: extension for FCM push notification support (published)
- Core-Shared-Tests: test fixtures and other shared test functionality (unpublished) -- the other modules only depend upon this with `testImplementation`, so it is only used for test builds.

I still intend to look over how the modules share services a bit more, and I probably have some README updates to make. 

## Test Plan

All unit tests still passing

## Related Issues/Tickets

https://klaviyo.tpondemand.com/entity/165252-gradle-project-structure


sdk/analytics/src/main/java/com/klaviyo/analytics/model/BaseModel.kt
```diff
@@ -1,4 +1,4 @@
-package com.klaviyo.coresdk.model
+package com.klaviyo.analytics.model
```
jordan-griffin, line 1: Could be in `core` module maybe? Since its pretty abstract

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
```diff
@@ -0,0 +1,37 @@
+package com.klaviyo.analytics.networking.requests
```
jordan-griffin, line 1: Left this comment on another PR, but since you are actually moving things around here, thoughts on this going into the `push` module?

sdk/analytics/src/main/java/com/klaviyo/analytics/model/Keyword.kt
```diff
@@ -0,0 +1,10 @@
+package com.klaviyo.analytics.model
```
jordan-griffin, line 1: Also could be in `core` maybe since its a pretty abstract class?

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
evan-masseau: So my take here was we publish a `push-fcm` module with the intent that it isolates _just_ the FCM dependency in an otherwise very small package. `analytics` contains all Klaviyo API client code, even the push token endpoint. The main reason was that implementing another push platform (fire os I think? maybe baidu?) would just be another module like `push-fireos`

sdk/analytics/src/main/java/com/klaviyo/analytics/model/BaseModel.kt
evan-masseau: True, just gotta double check that it won't complain when we expose `Profile` publicly, but I don't think it will.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
jordan-griffin: Aah gotcha that makes a lot of sense! I'm ok with that approach!

sdk/analytics/src/main/java/com/klaviyo/analytics/model/BaseModel.kt
evan-masseau: Done

sdk/analytics/src/main/java/com/klaviyo/analytics/model/Keyword.kt
evan-masseau: Done

evan-masseau: Darn, actually can't move this. Weirdly, there's `infix` methods that map back to the base class and so if the base class isn't public, you can't do something like `mapOf(ProfileKey.EMAIL to "emailaddress") because the infix function `to` isn't public!!

jordan-griffin: Oof thats unexpected!

evan-masseau: Yeah I did not see that coming!

PR #40, opened 2023-02-24
Add different timers based on network connection type

jordan-griffin:
# Purpose
DRAFT
Just messing around with network timers. Trying it out as an array stored in config but probably won't stick to this pattern. Just thought it'd be easier than adding more ints

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)


sdk/core/src/main/java/com/klaviyo/core/config/Config.kt
```diff
@@ -20,7 +20,9 @@ interface Config {
         fun applicationContext(context: Context): Builder
         fun debounceInterval(debounceInterval: Int): Builder
         fun networkTimeout(networkTimeout: Int): Builder
-        fun networkFlushInterval(networkFlushInterval: Int): Builder
+        fun networkFlushIntervalWifi(networkFlushInterval: Int): Builder
+        fun networkFlushIntervalCell(networkFlushInterval: Int): Builder
+        fun networkFlushIntervalOffline(networkFlushInterval: Int): Builder
```
evan-masseau, line 25: FYI we _should_ be able to just turn off the background thread altogether when offline, and restart it when we get a network-restored event. This would be a good fall back though if a device doesn't properly report connection change events... that is an area I need to do a little real-world testing on. In theory the `NewtorkMonitor` class is providing that functionality, but since we are supporting back to API 23 we should thoroughly test older APIs.

jordan-griffin: Ya I had a similar thought about the offline state but just decided to follow what was written in the ticket as a fallback for the same reasons you are outlying here

sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
```diff
@@ -161,13 +179,25 @@ object KlaviyoConfig : Config {
...
-            KlaviyoConfig.networkFlushInterval = networkFlushInterval
+            KlaviyoConfig.networkFlushIntervals = networkFlushIntervals
             KlaviyoConfig.networkFlushDepth = networkFlushDepth
             KlaviyoConfig.networkMaxRetries = networkMaxRetries
 
             return KlaviyoConfig
         }
     }
+
+    enum class NetworkTypes(val position: Int) {
```
jordan-griffin, line 190: Not sure if this should live here (where configs store timers based on network types) or in the network monitor (where timers are read based on network types). Guess it depends on if we'd use network types for other operations in the future.

sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
```diff
@@ -161,13 +179,25 @@ object KlaviyoConfig : Config {
...
+        WIFI(0),
+        CELL(1),
+        OFFLINE(2);
+
+        companion object {
+            fun fromPosition(position: Int) = NetworkTypes.values().first() {
```
evan-masseau, line 196: Would it be simpler to just have this as a raw enum and store the intervals as a `map<NetworkType, Int>`

sdk/core/src/main/java/com/klaviyo/core/networking/KlaviyoNetworkMonitor.kt
```diff
@@ -53,6 +54,20 @@ internal object KlaviyoNetworkMonitor : NetworkMonitor {
         .getNetworkCapabilities(connectivityManager.activeNetwork)
         ?.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET) ?: false
 
+    /**
+     * Check what type of network connection is currently servicing the device
+     *
+     * @return Integer representing the current network type
+     */
+    override fun getNetworkType(): Int {
```
evan-masseau, line 62: You could broadcast this value as part of the `onNetworkChange` event, i.e. update `NetworkObserver` typealias to take a second argument, network type. Then network monitor obersvers can get live updates.

sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
jordan-griffin: Oh ya, I didnt even realize I pluralized that. Gross. Will fix and move to `NetworkMonitor`

sdk/core/src/main/java/com/klaviyo/core/networking/KlaviyoNetworkMonitor.kt
evan-masseau: Then again I guess that would only be necessary if we were "interrupting" a background job to change timers when network changes, which we said we didn't want to do.

sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
jordan-griffin: Also realized I could probably just use `ordinal()` instead of having this passed in `position` value to assign to each enum entry. Can go either way I spose

evan-masseau: probably the nitpickiest of nits. `ordinal` is cool, hadn't seen that before

sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
jordan-griffin: Can do. I considered both approaches. Just thought an array was simpler to implement but a map might be nicer overall.

I'm happy to make the swap if thats preferrable.

sdk/core/src/main/java/com/klaviyo/core/networking/KlaviyoNetworkMonitor.kt
jordan-griffin: Oh good point, I'll look into that!

sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
evan-masseau: TBH I'm happy either way, sometimes I'm just typing whatever comes into my head...

evan-masseau (approved):
I'm good with this implementation, its still on my todo list to test this background runnable thing more thoroughly on a real device anyways, so I'll test this at the same time.

sdk/core/src/main/java/com/klaviyo/core/networking/KlaviyoNetworkMonitor.kt
evan-masseau: Yeah, or it would just be unused. So agreed, no need to make that change now.

PR #41, opened 2023-02-27
Logging implementation and events exposure

evan-masseau:
# Purpose

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

How I intend to implement logging in the SDK as well as expose logged events to the tester app.
2 build types:
- debug = debuggable, has logging to console
- release = code signed, release build, no-op logging implementation

2 "flavors": 
- local = targets local-klaviyo.com and allows http 
- production = targets klaviyo.com and restricts to https 

Note, I didn't split the http/https thing up yet in this PR but you _can_ do merged manifests with this technique I'm using

## Changelog / Code Overview

Firstly, I needed to be able to implement different logger for debug vs production. So I changed from using 3 "build types" to just 2 types, but added 2 flavors. You can add different source code sets for just build types, just flavors, or a cross of both. With this new arrangement, I only need a `debug` and `release` implementation, whereas I would have needed a third `localDebug` implementation when we had just 3 build types -- annoying. 

Secondly, I added a `Log` service to the Registry which exposes the `Logger` implementation from either release or production source set. I'll build upon that interface but basically anywhere we want to log a string or broadcast an event to the logger, we'll call `Registry.log.debug(string)` or `Registry.log.onNetworkRequest(request)` for example. 

Lastly, the registry allows `Registry.log` to be overridden. One can register a new `Log` service with `Registry.register<Log>(myLogger)` to receive al the logging callbacks. That's what the test app will do.


sdk/core/src/debug/java/com/klaviyo/core/Logger.kt
```diff
@@ -0,0 +1,31 @@
+package com.klaviyo.core
+
+import com.klaviyo.core.config.Log
+import com.klaviyo.core.config.NetworkRequest
+import java.lang.Exception
+
+internal object Logger : Log {
+    override fun debug(output: String) {
+        println(output)
```
ndurell, line 9: Will these use android logging at a later date?

sdk/core/src/main/java/com/klaviyo/core/config/Log.kt
```diff
@@ -0,0 +1,98 @@
...
+interface NetworkRequest {
+    /**
+     * Unsent, PendingRetry, Complete or Failed
+     */
+    val state: String
+
+    /**
+     * Time the request was initiated
+     */
+    val time: String
```
ndurell, line 55: Maybe good to have time completed?

constants.gradle
```diff
@@ -0,0 +1,62 @@
...
+        if (rootProject.file("local.properties").canRead()) {
+            localProperties.load(new FileInputStream(rootProject.file("local.properties")))
+        }
+
+        flavorDimensions "server"
+        productFlavors {
+            production {
+                //select the dimension of flavor
+                dimension "server"
+                buildConfigField "String", "KLAVIYO_SERVER_URL", '"https://a.klaviyo.com"'
```
ndurell, line 54: Maybe extract prod server url as a constant above?

constants.gradle
```diff
@@ -0,0 +1,62 @@
+ext {
+    // build dependencies
+    gradleBuildToolsVersion = '7.3.1'
```
ndurell, line 3: Total nit but I feel like keeping versions separate was a sort of convention I saw in a lot of android projects.

constants.gradle
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * Extracted product flavor as a closure to be applied for all android projects
+     */
+    flavorConfig = {
+        def localProperties = new Properties()
+        if (rootProject.file("local.properties").canRead()) {
+            localProperties.load(new FileInputStream(rootProject.file("local.properties")))
+        }
+
+        flavorDimensions "server"
```
ndurell, line 49: Hmm I forget what these mean but does it make sense for flavor dimension and dimension to be the same?

ndurell (commented):
Looks good so far.

sdk/core/src/debug/java/com/klaviyo/core/Logger.kt
kennyklaviyo: what do you mean by android logging? there's Logger but thats just from `java.utils`

evan-masseau: I put `println` in as a placeholder, I'm going to flesh this out with a snippet we were going to copy from timber that includes stack trace, and I think it even had a way to color the output too?

constants.gradle
evan-masseau: Could you elaborate, because specifying versions as constants is exactly what this file is for.

constants.gradle
evan-masseau: I think this line is how you specify all the available dimensions for flavors, and below is how you specify which dimension  _this_ flavor is for. Just so happens I only have 1 flavor dimension so it looks repetitive.

evan-masseau: Like we could theoretically have build types of `debug` and `release`, flavors defined by `flavorDimensions "server", "ice_cream"` and then ultimately have build _variants_ like this: 
```
productionVanillaRelease
productionChocolateRelease
debugVanillaRelease
...etc
```

sdk/core/src/debug/java/com/klaviyo/core/Logger.kt
ndurell: Android has built in logging capabilities. I think timber was a wrapper around it. https://developer.android.com/reference/android/util/Log

constants.gradle
ndurell: As a convention version constants are kept in a separate file called versions I think typically. It looks like you moved them with other constants here.

evan-masseau: oh i see, ok, i could make a separate constants file for non-version things

flavors.gradle
```diff
@@ -0,0 +1,30 @@
+ext {
```
evan-masseau, line 1: Changed from 3 build types to 2 build types and 2 flavors. makes it simpler to have just 2 release and production logger implementations -- besides it just seems like the better structure for what we're doing with targeting local vs prod server.

gradle.properties
```diff
@@ -23,3 +23,8 @@ android.useAndroidX=true
 # Automatically convert third-party libraries to use AndroidX
 android.enableJetifier=true
 
+# URL to Klaviyo server
```
evan-masseau, line 26: Using gradle.properties to extract some constants.

sdk/analytics/build.gradle
```diff
@@ -14,8 +14,8 @@ afterEvaluate {
         publications {
             // Creates a Maven publication called "release".
             release(MavenPublication) {
-                from components.release
-                groupId = ext.groupId
+                from components[ext.publishBuildVariant]
```
evan-masseau, line 17: Might want to publish a debug variant too, but at least you can publish to local maven and use that from the test app. The _sample_ app will be able to use any build variant because depends on the projects instead of artifacts

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/ApiClient.kt
```diff
@@ -27,4 +33,18 @@ internal interface ApiClient {
      * @param profile
      */
     fun enqueueEvent(event: Event, profile: Profile)
+
+    /**
+     * Register an observer to be notified when an API request is enqueued or changes state
+     *
+     * @param observer
+     */
+    fun onApiRequest(observer: ApiObserver)
```
evan-masseau, line 42: I opted to just expose the more complicated "events" as part of the services they belong to, rather than in the Log interface. Some of my services already had this pattern anyway.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/ApiClient.kt
```diff
@@ -2,8 +2,14 @@ package com.klaviyo.analytics.networking
...
 import com.klaviyo.analytics.model.Profile
+import com.klaviyo.analytics.networking.requests.ApiRequest
 
-internal interface ApiClient {
+typealias ApiObserver = (request: ApiRequest) -> Unit
+
+/**
+ * Defines public API of the network coordinator service
+ */
+interface ApiClient {
```
evan-masseau, line 12: While the interface is public, the implementation will still be inaccessible like all other services except via `core`

sdk/core/src/debug/java/com/klaviyo/core/Console.kt
```diff
@@ -0,0 +1,19 @@
+package com.klaviyo.core
```
evan-masseau, line 1: Isolates dependency on android.util.log for testing

sdk/core-shared-tests/src/main/java/com/klaviyo/core_shared_tests/StaticClock.kt
```diff
@@ -1,4 +1,4 @@
-package com.klaviyo.core_shared_tests
+package com.klaviyo.core_shared_tests // ktlint-disable package-name
```
evan-masseau, line 1: ktlint is complaining about underscores in my package names. not sure how this got by pre-commit before, but i need to disable it for now and circle back to it.

sdk/core/src/debug/java/com/klaviyo/core/Logger.kt
```diff
@@ -0,0 +1,62 @@
...
+        )
+
+        /**
+         * Extract the tag which should be used for the message from the `element`. By default
+         * this will use the class name without any anonymous class suffixes (e.g., `Foo$1`
+         * becomes `Foo`).
+         *
+         * Note: This will not be called if a manual tag is specified
+         */
+        private fun makeTag(): String = Throwable().stackTrace
```
evan-masseau, line 46: we'll tag our output with the class it came from

sdk/core/src/main/java/com/klaviyo/core/KlaviyoException.kt
```diff
@@ -0,0 +1,14 @@
+package com.klaviyo.core
+
+/**
+ * Exceptions that automatically hook into our logger
+ *
+ * @property message
+ */
+abstract class KlaviyoException(final override val message: String) : Exception(message) {
```
evan-masseau, line 8: For any of our custom exception classes, this will give you logging for free

sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
```diff
@@ -108,55 +125,52 @@ object KlaviyoConfig : Config {
...
-                // TODO Logging
-            }
-        }
-
-        override fun networkFlushIntervalCell(networkFlushInterval: Int) = apply {
-            if (networkFlushInterval >= 0) {
-                this.networkFlushIntervals[NetworkMonitor.NetworkType.CELL.position] = networkFlushInterval
-            } else {
-                // TODO Logging
+                Registry.log.error(
```
evan-masseau, line 138: There is a separate PR about this, i didn't mean to get this intermingled, but I made a mistake in git. best laid plans...

versions.gradle
```diff
@@ -3,18 +3,17 @@ ext {
     gradleBuildToolsVersion = '7.3.1'
     gradleKotlinVersion = '1.7.20'
     dokkaVersion = '1.7.20'
-    gradleKtlintVersion = "11.0.0"
+    gradleKtlintVersion = "11.2.0"
```
evan-masseau, line 6: OH this is probably why that ktlint change happened! hah.

gradle.properties
ndurell: Unrelated to comment is there going to be directions on how to override?

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -37,8 +56,11 @@ internal object KlaviyoApiClient : ApiClient {
 
         // Stop the background batching job while offline
         Registry.networkMonitor.onNetworkChange { isOnline ->
-            if (isOnline) startBatch(true)
-            else stopBatch()
+            if (isOnline) {
```
ndurell, line 59: Did we change the formmatter?

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -96,24 +131,35 @@ internal object KlaviyoApiClient : ApiClient {
...
-                    Registry.dataStore.clear(uuid)
+                } else {
+                    try {
+                        val request = KlaviyoApiRequest.fromJson(JSONObject(json))
+                        if (!apiQueue.contains(request)) {
+                            apiQueue.offer(request)
+                        }
+                    } catch (exception: JSONException) {
+                        wasMutated = true
+                        Registry.log.wtf("Invalid request JSON $uuid", exception)
```
ndurell, line 159: Nice I like how you add wtf!

gradle.properties
evan-masseau: the override goes in local.properties and I believe the instructions are in the flavor config, as well is our wiki

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/ApiRequest.kt
```diff
@@ -0,0 +1,65 @@
+package com.klaviyo.analytics.networking.requests
+
+import java.net.URL
+
+/**
+ * Immutable representation of the data of a network request
+ */
+interface ApiRequest {
```
ndurell, line 8: I did something similar. I’m thinking it might be interesting though to more events like flushing the queue etc…though at the same time might be harder to find what you are looking for in the test app unless we add filter.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
evan-masseau: i mean how could i not once i saw it in android logger...

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
evan-masseau: I'm not sure why that happened, the pre-commit formatter is pretty lax, I might have done this myself absentmindedly.

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
```diff
@@ -100,14 +116,12 @@ internal class KlaviyoApiClientTest : BaseTest() {
 
     @Test
     fun `Enqueues a profile API call`() {
```
ndurell, line 118: Are you adding additional assertions to existing tests for logs. Kind of a nit but maybe worth isolating to different tests. Like if something changed with the logging and a test called might be confusing.

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
```diff
@@ -100,14 +116,12 @@ internal class KlaviyoApiClientTest : BaseTest() {
...
-        mockkStatic(Base64::class)
-        every { Base64.encodeToString(any(), any()) } returns "mock"
-
         assertEquals(0, KlaviyoApiClient.getQueueSize())
 
         KlaviyoApiClient.enqueueProfile(Profile().setAnonymousId(ANON_ID))
 
         assertEquals(1, KlaviyoApiClient.getQueueSize())
+        verify { logSpy.debug(any()) }
```
ndurell, line 124: Total nit: Possible to assert on something other than any.

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
evan-masseau: I did a bit of both

sdk/core/src/debug/java/com/klaviyo/core/Console.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import android.util.Log
+
+/**
+ * Android Log output wrapper
+ */
+object Console {
+    enum class Level {
```
ndurell, line 9: Neat!

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
evan-masseau: Again I did a bit of both, in this case the log output is pretty large, so i did the lazy thing of just using any

sdk/core/src/debug/java/com/klaviyo/core/Console.kt
ndurell: Oh btw swift has a way to capture file and line number not sure if that’s possible in kotlin but might be nice to have!

evan-masseau: I do capture classname as the "tag". I'm not sure if i could get line number, but if we're logging an exception it should log the whole stack trace. For info/debug i didn't think line number was especially necessary, you can pretty much tell where it came from just from the message string anyway

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/ApiRequest.kt
evan-masseau: Yeah I wouldn't be surprised if we have some edits to this once we start using it in the tester app.

ndurell (approved):
Looks like it’ll do the job. There’s a bunch of non blocking comments.

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
evan-masseau: But fair point, i can probably test all the logging i need in just 1 added test rather than adding to each of these

sdk/core/src/debug/java/com/klaviyo/core/Console.kt
```diff
@@ -0,0 +1,19 @@
...
+        Debug, Info, Error, Assert
+    }
+
+    fun log(msg: String, level: Level, tag: String, ex: Throwable? = null) = when (level) {
+        Level.Debug -> ex?.let { Log.d(tag, msg, ex) } ?: Log.d(tag, msg)
+        Level.Info -> ex?.let { Log.i(tag, msg, ex) } ?: Log.i(tag, msg)
+        Level.Error -> ex?.let { Log.e(tag, msg, ex) } ?: Log.e(tag, msg)
+        Level.Assert -> ex?.let { Log.wtf(tag, msg, ex) } ?: Log.wtf(tag, msg)
```
kennyklaviyo, line 17: instead of doing this `when` statement on each log, you can reference the func as a property of the enum:
```
object Console {
    enum class Level(val call: (String, String, Throwable?) -> Unit) {
        Debug(Log::d), Info(Log::i), Error(Log::e), Assert(Log::wtf)
    }

    fun log(msg: String, level: Level, tag: String, ex: Throwable? = null) {
        level.call(tag, msg, ex)
    }
}
```

kennyklaviyo: based on source code, looks like they all handle nullable throwables so it should be fine

evan-masseau: oh dang thats a good idea

sdk/core/src/debug/java/com/klaviyo/core/Console.kt
kennyklaviyo: looks like you do use the stacktrace object below so youll have the line number, file name, class name and everything

evan-masseau: ya, I adapted that from timber code like we talked about during grooming, just hadn't explored what other properties are in there. makes sense line number would be

evan-masseau: Copied over the line number thing from kenny's utils repo

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
evan-masseau: I put logging into its own tests and tested for the actual output.

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
evan-masseau: I put logging into its own tests and tested for the actual output.

PR #42, opened 2023-03-01
Updated behavior of setProfile and resetProfile

evan-masseau:
# Purpose

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

`setProfile` behavior felt confusing based on the method's name

## Changelog / Code Overview

Updated reset profile to handle disassociating the push token from the current profile. Since we don't have an endpoint for this, @ndurell and I talked about just setting it to another anonymous profile. 
Updated setProfile behavior to call reset profile first, i.e. replace the current profile with a new one. 
Added tests and documentation changes to go along with the above

## Test Plan

Unit tests

## Related Issues/Tickets

SDK API's spec was updated after codeshare/discussion


sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -174,8 +173,7 @@ object Klaviyo {
      * @return Returns [Klaviyo] for call chaining
      */
     fun setProfile(profile: Profile): Klaviyo = apply {
```
ndurell, line 175: Note to self: check if iOS implementation is chainable.

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -164,8 +164,7 @@ object Klaviyo {
     }
 
     /**
-     * Updates the currently tracked profile from a map-like [Profile] object
-     * Saves identifying attributes (externalId, email, phone) to the currently tracked profile.
+     * Replaces the currently tracked profile with a new [Profile] object
```
ndurell, line 167: Maybe re-iterating that this will reset the existing profile.

ndurell (approved):
Overall makes sense just left some comments.

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
evan-masseau: good thinking, added an @see for reset profile

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
evan-masseau: arguably this one doesn't really need it, but i could see maybe wanting to do 
```
Klaviyo.setProfile(profile)
   .setPushToken(token)
```

PR #43, opened 2023-03-01
Some things I didn't catch about the network types/timers code

evan-masseau:
# Purpose

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

I just noticed some more things about the network timers PR jordan and I merged yesterday once i actually started looking at the code in android studio.

## Changelog / Code Overview

Fixed a logical hole where transport types other than wifi or cell would be treated as offline 
Added Test coverage
Some other trivial things like enum casing and using 1 config method with an argument instead of 3 separate methods.

## Test Plan

Added coverage

### DIFF
--- a/sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
+++ b/sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
@@ -8,7 +8,6 @@ import com.klaviyo.analytics.model.EventType
 import com.klaviyo.analytics.model.Profile
 import com.klaviyo.analytics.networking.requests.KlaviyoApiRequest
 import com.klaviyo.core.Registry
-import com.klaviyo.core.config.KlaviyoConfig
 import com.klaviyo.core.lifecycle.ActivityEvent
 import com.klaviyo.core.lifecycle.ActivityObserver
 import com.klaviyo.core.networking.NetworkMonitor
@@ -52,10 +51,14 @@ internal class KlaviyoApiClientTest : BaseTest() {
         delayedRunner = null
 
         every { Registry.clock } returns staticClock
-        every { configMock.networkFlushIntervals } returns intArrayOf(flushIntervalWifi, flushIntervalCell, flushIntervalOffline)
+        every { configMock.networkFlushIntervals } returns intArrayOf(
+            flushIntervalWifi,
+            flushIntervalCell,
+            flushIntervalOffline
+        )
         every { configMock.networkFlushDepth } returns queueDepth
         every { networkMonitorMock.isNetworkConnected() } returns false
-        every { networkMonitorMock.getNetworkType() } returns NetworkMonitor.NetworkType.WIFI.position
+        every { networkMonitorMock.getNetworkType() } returns NetworkMonitor.NetworkType.Wifi
         every { lifecycleMonitorMock.onActivityEvent(capture(slotOnActivityEvent)) } returns Unit
         every { networkMonitorMock.onNetworkChange(capture(slotOnNetworkChange)) } returns Unit
 
@@ -84,7 +87,20 @@ internal class KlaviyoApiClientTest : BaseTest() {
         mockk<KlaviyoApiRequest>().also {
             every { it.uuid } returns uuid
             every { it.send() } returns status
-            every { it.toJson() } returns "{\"headers\":{\"headerKey\":\"headerValue\"},\"method\":\"GET\",\"query\":{\"queryKey\":\"queryValue\"},\"time\":\"time\",\"uuid\":\"$uuid\",\"url_path\":\"test\"}"
+            every { it.toJson() } returns """
+                {
+                  "headers": {
+                    "headerKey": "headerValue"
+                  },
+                  "method": "GET",
+                  "query": {
+                    "queryKey": "queryValue"
+                  },
+                  "time": "time",
+                  "uuid": "$uuid",
+                  "url_path": "test"
+                }
+            """.trimIndent()
             every { it.equals(any()) } answers { a ->
                 it.uuid == (a.invocation.args[0] as KlaviyoApiRequest).uuid
             }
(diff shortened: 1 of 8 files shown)


evan-masseau:
@jordan-griffin i'll tag you in case you want to see this later, but obvs not worth interrupting your PTO.

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
```diff
@@ -84,7 +87,20 @@ internal class KlaviyoApiClientTest : BaseTest() {
         mockk<KlaviyoApiRequest>().also {
             every { it.uuid } returns uuid
             every { it.send() } returns status
-            every { it.toJson() } returns "{\"headers\":{\"headerKey\":\"headerValue\"},\"method\":\"GET\",\"query\":{\"queryKey\":\"queryValue\"},\"time\":\"time\",\"uuid\":\"$uuid\",\"url_path\":\"test\"}"
+            every { it.toJson() } returns """
```
evan-masseau, line 90: this is unrelated just something i've been changing when i see it in tests, much easier to read multiline strings

sdk/core/src/main/java/com/klaviyo/core/networking/NetworkMonitor.kt
```diff
@@ -33,21 +33,15 @@ interface NetworkMonitor {
...
     /**
      * Enum class representing the different network connection types that may affect how our
      * SDK operates.
      */
     enum class NetworkType(val position: Int) {
-        WIFI(0),
-        CELL(1),
-        OFFLINE(2);
-
-        companion object {
```
evan-masseau, line 47: remove unused

sdk/core/src/main/java/com/klaviyo/core/networking/KlaviyoNetworkMonitor.kt
```diff
@@ -57,15 +56,20 @@ internal object KlaviyoNetworkMonitor : NetworkMonitor {
...
-        if (connectivityManager.getNetworkCapabilities(connectivityManager.activeNetwork)?.hasTransport(NetworkCapabilities.TRANSPORT_WIFI) == true) {
-            return NetworkMonitor.NetworkType.WIFI.position
-        } else if (connectivityManager.getNetworkCapabilities(connectivityManager.activeNetwork)?.hasTransport(NetworkCapabilities.TRANSPORT_CELLULAR) == true) {
-            return NetworkMonitor.NetworkType.CELL.position
+    override fun getNetworkType(): NetworkMonitor.NetworkType {
+        val net = connectivityManager.getNetworkCapabilities(connectivityManager.activeNetwork)
+
+        return if (net?.hasTransport(NetworkCapabilities.TRANSPORT_WIFI) == true) {
+            NetworkMonitor.NetworkType.Wifi
+        } else if (net?.hasTransport(NetworkCapabilities.TRANSPORT_CELLULAR) == true ||
```
evan-masseau, line 66: Technically internet can come from a few other transport methods like bluetooth, so this will just catch those and treat them as cellular (maybe we should call our types something more generic like Stable and Unstable idk)

sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
```diff
@@ -120,25 +133,12 @@ object KlaviyoConfig : Config {
...
-        override fun networkFlushIntervalCell(networkFlushInterval: Int) = apply {
-            if (networkFlushInterval >= 0) {
-                this.networkFlushIntervals[NetworkMonitor.NetworkType.CELL.position] = networkFlushInterval
-            } else {
-                // TODO Logging
-            }
-        }
-
-        override fun networkFlushIntervalOffline(networkFlushInterval: Int) = apply {
+        override fun networkFlushInterval(
```
evan-masseau, line 136: Use a single method that takes type as an argument instead of 3 methods

sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
```diff
@@ -169,11 +169,14 @@ object KlaviyoConfig : Config {
...
 
             val permissions = applicationContext!!.packageManager.getPackageInfoCompat(
-                applicationContext!!.packageName, PackageManager.GET_PERMISSIONS
+                applicationContext!!.packageName,
+                PackageManager.GET_PERMISSIONS
             ).requestedPermissions ?: emptyArray()
 
-            if (Manifest.permission.ACCESS_NETWORK_STATE !in permissions) {
-                throw MissingPermission(Manifest.permission.ACCESS_NETWORK_STATE)
+            requiredPermissions.forEach { permission ->
```
evan-masseau, line 176: semi-related, we require two different network permissions, so this checks both.

sdk/core/src/main/java/com/klaviyo/core/networking/KlaviyoNetworkMonitor.kt
ndurell: Neat!

sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
ndurell: Is caught by anything in the sdk or it will crash the app potentially?

ndurell: I think that's ok since we can't do much without internet.

evan-masseau: Any of these config errors we throw will crash the app if uncaught, yes. Its either that or the system crashes us later when we try to access those things that are gated by this permission. 
However, we actually declare these permissions in our core sdk manifest. I'm not sure if there is some way to exclude our manifest or something, so I put this in for extra safety

ndurell: Makes sense.

evan-masseau:
Was merged with the logging branch

PR #44, opened 2023-03-02
Important caveat to the setProfile update

evan-masseau:
set profile should _merge_ into an anonymous profile. only use the reset behavior if the developers have set identifiers.

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
@@ -50,8 +50,8 @@ object Klaviyo {
     }
 
     /**
-     * Replaces the currently tracked profile with a new [Profile] object
-     * @see resetProfile is called first, then the new profile object is saved
+     * Assign new identifiers and attributes to the currently tracked profile.
+     * If a profile has already been identified, it will be overwritten, by calling [resetProfile].
      *
      * The SDK keeps track of current profile details to
      * build analytics requests with profile identifiers
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
@@ -51,7 +51,7 @@ object Klaviyo {
 
     /**
      * Assign new identifiers and attributes to the currently tracked profile.
-     * If a profile has already been identified, it will be overwritten, by calling [resetProfile].
+     * If a profile has already been identified it will be overwritten by calling [resetProfile].
      *
      * The SDK keeps track of current profile details to
      * build analytics requests with profile identifiers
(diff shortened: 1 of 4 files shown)


sdk/analytics/src/test/java/com/klaviyo/analytics/UserInfoTest.kt
```diff
@@ -26,17 +26,6 @@ internal class UserInfoTest : BaseTest() {
         assertUserInfoIdentifiers()
     }
 
-    @Test
-    fun `Updates UserInfo identifiers from a Profile object`() {
```
evan-masseau, line 30: deleted method from userinfo

sdk/analytics/src/main/java/com/klaviyo/analytics/UserInfo.kt
```diff
@@ -61,15 +61,9 @@ internal object UserInfo {
...
-     *
-     * @param profile
+     * Indicate whether we currently have externally-set profile identifiers
      */
-    fun updateFromProfile(profile: Profile) = apply {
-        externalId = profile.externalId ?: externalId
-        email = profile.email ?: email
-        phoneNumber = profile.phoneNumber ?: phoneNumber
-    }
+    val isIdentified get() = (externalId.isNotEmpty() || email.isNotEmpty() || phoneNumber.isNotEmpty())
```
evan-masseau, line 66: flags whether we're currently an anonymous only profile

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -60,8 +60,15 @@ object Klaviyo {
      * @return Returns [Klaviyo] for call chaining
      */
     fun setProfile(profile: Profile): Klaviyo = apply {
-        resetProfile()
-        UserInfo.updateFromProfile(profile)
+        if (UserInfo.isIdentified) {
```
evan-masseau, line 63: Heres the key piece: if currently just anonymous, we should still be merging
Also below I just figured since I'm setting userinfo var directly in the individual `setEmail` etc methods, I'd do the same here.

sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoTest.kt
```diff
@@ -121,42 +113,58 @@ internal class KlaviyoTest : BaseTest() {
...
 
-        Klaviyo.setProfile(
-            Profile()
-                .setExternalId(EXTERNAL_ID)
-        )
+        Klaviyo.setProfile(newProfile)
 
-        assert(UserInfo.externalId == EXTERNAL_ID)
-        assert(UserInfo.email == "")
+        assertEquals(EXTERNAL_ID, UserInfo.externalId)
```
evan-masseau, line 142: `assertEquals` produced better test out put than just `assert`, fixing these wherever i see them now.

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
evan-masseau: Yeah I was gonna ask to what extent should we explain anonymous profiles to the app developers?

evan-masseau: Updated the comment without going into the weeds on what anonymous profiles are -- we can decide if we want a section about that in docs later

ndurell (approved):

PR #50, opened 2023-03-10
Bugfix: for url encoding the push token payload

evan-masseau:
the token data payload must also be url-encoded. Otherwise a plus turns into a space

# Purpose

**Type of PR**
* [x] Bug Fix
* [ ] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Test Plan

fixed unit test... unfortunately a unit test is only as good as the person who writes the expectations.

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
@@ -4,6 +4,7 @@ import com.klaviyo.analytics.model.Profile
 import com.klaviyo.core.Registry
 import java.io.Serializable
 import java.net.HttpURLConnection
+import java.net.URLEncoder
 import org.json.JSONObject
 
 /**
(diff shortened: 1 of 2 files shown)


ajaysubra (approved):

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
```diff
@@ -26,6 +27,11 @@ internal class PushTokenApiRequest(token: String, profile: Profile) : KlaviyoApi
         const val ANDROID_TOKEN = "\$android_tokens"
     }
 
+    override var headers: Map<String, String> = mapOf(
+        HEADER_ACCEPT to TYPE_HTML,
```
kennyklaviyo, line 31: curious why we need to accept html?

kennyklaviyo: interesting....cant wait to get off this api

kennyklaviyo (approved):

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
evan-masseau: It beats base64 in a GET url, but... only barely

PR #51, opened 2023-03-10
Ecm/test app helpers

evan-masseau:
# Purpose

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

Adding little features to support test app development

## Changelog / Code Overview

- Fixed/improved the API request logging -- discovered that it was just logging "Request will (Action)" but couldn't tell what request it was.
- Exposed api request unique identifier to the logger interface
- Added an "inflight" status to api requests.

api requests in logcat now: 
Screen Shot 2023-03-10 at 12 29 00

## Test Plan

Unit test and test app

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
@@ -43,10 +43,11 @@ internal object KlaviyoApiClient : ApiClient {
             }
 
             r.responseBody?.let { response ->
+                val body = r.requestBody?.let { JSONObject(it).toString(2) }
                 Registry.log.verbose("${r.httpMethod}: ${r.url}")
                 Registry.log.verbose("Headers: ${r.headers}")
                 Registry.log.verbose("Query: ${r.query}")
-                Registry.log.verbose("Body: ${r.requestBody}")
+                Registry.log.verbose("Body: $body")
                 Registry.log.verbose("${r.responseCode} $response")
             }
         }
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
@@ -7,6 +7,7 @@ import com.klaviyo.analytics.model.Event
 import com.klaviyo.analytics.model.Profile
 import com.klaviyo.analytics.networking.requests.EventApiRequest
 import com.klaviyo.analytics.networking.requests.KlaviyoApiRequest
+import com.klaviyo.analytics.networking.requests.KlaviyoApiRequest.Status
 import com.klaviyo.analytics.networking.requests.ProfileApiRequest
 import com.klaviyo.analytics.networking.requests.PushTokenApiRequest
 import com.klaviyo.core.Registry
@@ -34,18 +35,10 @@ internal object KlaviyoApiClient : ApiClient {
     init {
         onApiRequest { r ->
             when (r.state) {
-                KlaviyoApiRequest.Status.Unsent.name -> Registry.log.debug(
-                    "${r.type} Request enqueued"
-                )
-                KlaviyoApiRequest.Status.Inflight.name -> Registry.log.debug(
-                    "${r.type} Request inflight"
-                )
-                KlaviyoApiRequest.Status.PendingRetry.name -> Registry.log.error(
-                    "${r.type} Request failed, will retry"
-                )
-                KlaviyoApiRequest.Status.Complete.name -> Registry.log.info(
-                    "${r.type} Request completed"
-                )
+                Status.Unsent.name -> Registry.log.debug("${r.type} Request enqueued")
+                Status.Inflight.name -> Registry.log.debug("${r.type} Request inflight")
+                Status.PendingRetry.name -> Registry.log.error("${r.type} Request retrying")
+                Status.Complete.name -> Registry.log.info("${r.type} Request completed")
                 else -> Registry.log.error("${r.type} Request failed")
             }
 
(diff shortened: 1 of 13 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
ndurell, line 1: Nice file!

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -32,17 +33,22 @@ internal object KlaviyoApiClient : ApiClient {
...
-        onApiRequest {
-            Registry.log.info(
-                when (it.state) {
-                    KlaviyoApiRequest.Status.Unsent.name -> "Request Enqueued"
-                    KlaviyoApiRequest.Status.PendingRetry.name -> "Request will be retried"
-                    KlaviyoApiRequest.Status.Complete.name -> "Request Completed"
-                    else -> "Request Failed"
-                }
-            )
+        onApiRequest { r ->
```
ndurell, line 36: So before this was logging one thing but now it's logging a number of things. I'm a little out of the loop but are you using this logging to drive the test app. Is there some way you re-assemble this on the other end?

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -241,13 +247,13 @@ internal object KlaviyoApiClient : ApiClient {
             while (apiQueue.isNotEmpty()) {
                 val request = apiQueue.poll()
 
-                when (request?.send()) {
-                    KlaviyoApiRequest.Status.Complete, KlaviyoApiRequest.Status.Failed -> {
+                when (request?.send { broadcastApiRequest(request) }) {
+                    Status.Complete, Status.Failed -> {
                         // On success or absolute failure, remove from queue and persistent store
                         Registry.dataStore.clear(request.uuid)
                         broadcastApiRequest(request)
```
ndurell, line 254: It looks like you broadcast above too...just checking if that's intentional.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -257,7 +263,7 @@ internal object KlaviyoApiClient : ApiClient {
                         break
                     }
                     // Offline or at the end of the queue... either way break the loop
-                    null, KlaviyoApiRequest.Status.Unsent -> break
+                    else -> break
```
ndurell, line 266: Do you think it might be better to be exhaustive in case you add more cases?

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -202,17 +206,20 @@ internal open class KlaviyoApiRequest(
      *
      * @returns The string value of the response body, if one was returned
      */
-    fun send(): Status {
+    fun send(beforeSend: (() -> Unit)? = null): Status {
         if (!Registry.networkMonitor.isNetworkConnected()) {
```
ndurell, line 210: Just curious: is this based on notifications from the network listener or is it asking the network if it's connected. If the latter might want to see how this affects battery usage.

ndurell (approved):
lgtm just had some questions.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -202,17 +206,20 @@ internal open class KlaviyoApiRequest(
      *
      * @returns The string value of the response body, if one was returned
      */
-    fun send(): Status {
+    fun send(beforeSend: (() -> Unit)? = null): Status {
```
kennyklaviyo, line 209: instead of doing nullable can just default to empty lambda: `beforeSend: (() -> Unit) = {}`

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -202,17 +206,20 @@ internal open class KlaviyoApiRequest(
...
         }
 
+        status = Status.Inflight
+
         return try {
             val connection = buildUrlConnection()
 
             try {
                 attempts++
+                beforeSend?.invoke()
```
kennyklaviyo, line 222: then theres no need to nullsafe here

kennyklaviyo (dismissed):

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
evan-masseau: Good question, I believe this spot check is asking the system directly. I can probably optimize that in the network service a bit

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
evan-masseau: Yep! The one above is a pre-send broadcast, this one broadcasts the response code/body

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
evan-masseau: This is _just_ debug logging, the test app can subscribe using the same `onApiRequest` callback mechanism that this block is using.

This particular change is just fixing a mistake I'd made before. Earlier, I just logged a message like `Request Queued` and then logged `Request.toString()`, but that toString wasn't doing what I expected. Decided to just make it much more explicit so the output logs are nice and readable. Also changed it to log level verbose so it can be filtered out more easily

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
evan-masseau: but that'd be like the safer and smarter thing to do

jordan-griffin (dismissed):

ajaysubra (approved):

PR #55, opened 2023-03-20
Add sub-type api requests JSON representation

evan-masseau:
# Purpose

**Type of PR**
* [x] Bug Fix
* [x] Feature Work
* [ ] Revert
* [x] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

There were some noticeable inconsistencies between the test app's events page output depending on whether I let the events send immediately, or if I terminated the app so that they had to send after being restored from the persistent store. I think it came from the fact that restoring from disk I was losing the API request class sub-types.

So this adds type-specific json de-serializing for our API request classes. I had to adjust the constructors a little bit, which makes the diff look large, but it was actually a pretty straightforward change. I didn't think we needed to worry about this originally, but as I've discovered idiosyncracies between the API endpoints, I need to be able to have some customizations in response handlers per endpoint. Without this, there is a difference of behavior between sending, say, a Push Token API request immediately vs json encoding it, decoding it, and sending it. 

## Changelog / Code Overview

Other things in this PR: 
- Added sub-type specific "success" status codes (e.g. 202 for v3 POST apis, 200 for v2 APIs)
- Cleaned up/ consolidated some properties and methods that I hadn't noticed were a bit confusing and duplicative (e.g. body, requestBody and formatBody)
- 

## Test Plan

- Unit tests
- Test app

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
@@ -18,87 +18,19 @@ import org.json.JSONObject
  * @property method The [RequestMethod] will determine the type of request being made
  * @property queuedTime The wall clock time that the request was enqueued
  * @property uuid unique identifier of this request
+ *
+ * @constructor
  */
 internal open class KlaviyoApiRequest(
     val urlPath: String,
     val method: RequestMethod,
-    override val queuedTime: Long = Registry.clock.currentTimeMillis(),
-    override val uuid: String = UUID.randomUUID().toString()
+    queuedTime: Long? = null,
+    uuid: String? = null
 ) : ApiRequest {
 
     internal enum class Status {
         Unsent, Inflight, PendingRetry, Complete, Failed
     }
-
-    override val type: String = urlPath
-    override val state: String get() = status.name
-
-    protected var status: Status = Status.Unsent
-        set(value) {
-            if (field == value) return
-            field = value
-
-            if (value == Status.Inflight) {
-                startTime = Registry.clock.currentTimeMillis()
-            } else if (status in arrayOf(Status.Complete, Status.Failed)) {
-                endTime = Registry.clock.currentTimeMillis()
-            }
-        }
-
-    override val httpMethod: String = method.name
-    override var headers: Map<String, String> = emptyMap()
-    override var query: Map<String, String> = emptyMap()
-    open var body: JSONObject? = null
-    override val requestBody: String? get() = formatBody()
-
-    var attempts = 0
-        private set
-
-    override var startTime: Long? = null
-        protected set
-    override var responseCode: Int? = null
-        protected set
-    override var responseBody: String? = null
-        protected set
-    override var endTime: Long? = null
-        protected set
-
-    /**
-     * Creates a representation of this [KlaviyoApiRequest] in JSON
-     *
-     * NOTE: subclass information is lost in the process.
-     * The important data to send the API request is retained.
-     *
-     * @return
-     */
-    fun toJson(): String = JSONObject()
-        .accumulate(PATH_JSON_KEY, urlPath)
-        .accumulate(METHOD_JSON_KEY, method.name)
-        .accumulate(TIME_JSON_KEY, queuedTime)
-        .accumulate(UUID_JSON_KEY, uuid)
-        .accumulate(HEADERS_JSON_KEY, JSONObject(headers))
-        .accumulate(QUERY_JSON_KEY, JSONObject(query))
-        .accumulate(BODY_JSON_KEY, body)
-        .toString()
-
-    override fun toString(): String = toJson()
-
-    open fun formatBody(): String? = body?.let { JSONObject(mapOf(DATA to it)).toString() }
-
-    /**
-     * To facilitate deduplication, we will treat UUID as a unique identifier
-     * such that even if a request is deserialized from JSON it is still
-     * "equal" to its original object, regardless of instance or subclass
-     *
-     * @param other
-     * @return
-     */
-    override fun equals(other: Any?): Boolean = when (other) {
-        is KlaviyoApiRequest -> uuid == other.uuid
-        else -> super.equals(other)
-    }
-
-    override fun hashCode(): Int = uuid.hashCode()
 
     companion object {
         // Common keywords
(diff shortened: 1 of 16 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -18,88 +18,20 @@ import org.json.JSONObject
...
-    override val uuid: String = UUID.randomUUID().toString()
+    queuedTime: Long? = null,
+    uuid: String? = null
 ) : ApiRequest {
 
     internal enum class Status {
         Unsent, Inflight, PendingRetry, Complete, Failed
     }
 
-    override val type: String = urlPath
```
evan-masseau, line 33: I did a bit of file organizing, usually I order it like this:
Subtypes
Companion objects
Properties
Constructors
Methods

Things had gotten disorganized here

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -18,88 +18,20 @@ import org.json.JSONObject
...
-        .accumulate(TIME_JSON_KEY, queuedTime)
-        .accumulate(UUID_JSON_KEY, uuid)
-        .accumulate(HEADERS_JSON_KEY, JSONObject(headers))
-        .accumulate(QUERY_JSON_KEY, JSONObject(query))
-        .accumulate(BODY_JSON_KEY, body)
-        .toString()
-
-    override fun toString(): String = toJson()
-
-    open fun formatBody(): String? = body?.let { JSONObject(mapOf(DATA to it)).toString() }
```
evan-masseau, line 86: I had some confusing duplication of logic surrounding the request body. I simplified it a little bit to just a JSON representation and a computed property that translates it to a string.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -149,7 +80,12 @@ internal open class KlaviyoApiRequest(
             val time = json.getLong(TIME_JSON_KEY)
             val uuid = json.getString(UUID_JSON_KEY)
 
-            return KlaviyoApiRequest(urlPath, method, time, uuid).apply {
+            return when (json.optString(TYPE_JSON_KEY)) {
```
evan-masseau, line 83: Using simple class name map to decode JSON payload back to the original subtype class

jordan-griffin: So was the problem that before you were instantiating a `KlaviyoApiRequest` instead of mapping back to what it was originally from the json provided?

jordan-griffin: Cause the logic here makes sense to me but I think the use of concrete classes within the parent class is a bit of an anti-pattern.

jordan-griffin: I'm wondering if it would make sense to move this logic up a bit higher into the caller (`KlaviyoApiClient`) in this case, to determine the class to instantiate from the given JSON before any request objects are invoked.

jordan-griffin: Ya I'd heavily prefer if we didn't have an abstract/parent class dependent on its children. Doing it dynamically as well also sounds fine. I'd just strive to keep these references out of here so do away with the "circular dependency" we got going on here.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -149,7 +80,12 @@ internal open class KlaviyoApiRequest(
...
+                ProfileApiRequest::class.simpleName -> ProfileApiRequest(time, uuid)
+                EventApiRequest::class.simpleName -> EventApiRequest(time, uuid)
+                PushTokenApiRequest::class.simpleName -> PushTokenApiRequest(time, uuid)
```
jordan-griffin, line 86: One other question I have, just to make sure my understanding is right:
We changed the constructors to each request object so that we can instantiate a barebones request object (cast to the correct type this time) and then the `fromJson` method will make sure that the jsonified fields (like `body` and `headers`) which are needed to send the request are re-populated. We have a second constructor taking in SDK-relavant information for instantiation when a user first queues a request which takes the given SDK params and translates them into the jsonified fields.
Correct?

evan-masseau: Exactly

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
evan-masseau: good call, definitely felt a little dirty doing it this way

evan-masseau: Tried dynamic, but it requires reflection and just felt wrong. Thinking now the most straightforward would be a helper object like `object KlaviyoApiRequestSerializer` so that this isn't directly in the `KlaviyoApiRequest` class. I liked putting it in `KlaviyoApiClient` at first, but it makes some dependency issues for unit tests

evan-masseau:
@jordan-griffin Does this change look better? I tried making it dynamic but it would have required reflection and just felt unreliable compared to this enumeration. So instead I extracted it to its own object (considered an extension function of JSONObject, but didn't feel quite right either). At least this way it isn't a super-class that refers to its child classes.

jordan-griffin:
> @jordan-griffin Does this change look better? I tried making it dynamic but it would have required reflection and just felt unreliable compared to this enumeration. So instead I extracted it to its own object (considered an extension function of JSONObject, but didn't feel quite right either). At least this way it isn't a super-class that refers to its child classes.

I think this is the best option you could have taken here. The second best alternative I think would have been to move the logic into the caller `KlaviyoApiClient` I guess, but that could have resulted in inconsistency or duplication if we had more callers in the future. Requiring reflection or extensions I don't think would have been ideal here either

jordan-griffin (approved):

evan-masseau:
Cool thanks, I liked the idea of putting it into `KlaviyoApiClient` but it introduced a timing issue because that object has an initializer. Simply decoding a JSON object would invoke the initializer (e.g. in my unit tests)

PR #56, opened 2023-03-21
Add full remote message handling to SDK

evan-masseau:
Handle all pushes as data messages. Added utilities for handling push notifications, heavily influenced by FCM sdk code.

# Purpose

**Type of PR**
* [ ] Bug Fix
* [x] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Changelog / Code Overview

- Adds message handling to the SDK (we will now use only data messages, so our SDK will get to handle all messages)
- Deep link support because we can consistently format the notification intent 
- Icon support a few ways (need to update readme)

## Test Plan

Unit test coverage will be hard because there's so much to mock. I updated test coverage of PushService, but the Notification class has no tests right now
I tested with the tester app and our latest server app changes, and this works (first try!)


sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/Extensions.kt
```diff
@@ -0,0 +1,213 @@
+package com.klaviyo.pushFcm
```
evan-masseau, line 1: This file adds extension properties to `RemoteMessage` as a nice shortcut for parsing all our data keys

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/Extensions.kt
```diff
@@ -0,0 +1,213 @@
...
+
+/**
+ * Determine the resource ID of the small icon
+ *
+ * NOTE: We have to use a discouraged API because we can't expect
+ *  developers to know the Int value of their icon resources
+ *
+ * @return
+ */
+val RemoteMessage.smallIcon: Int
```
evan-masseau, line 105: Get notification icon resource ID, first check the message payload, then the manifests, then fall back on system default.

ajaysubra: I like that this future proofs the SDK but are we adding too much code here that we may not get to test end to end sending stuff to the SDK?

sdk/push-fcm/src/main/AndroidManifest.xml
```diff
@@ -1 +1,4 @@
-<manifest />
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
```
jordan-griffin, line 3: I always forget this is a thing you want to include for Android 13+

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/Notification.kt
```diff
@@ -0,0 +1,140 @@
...
+    /**
+     * Formats and displays a notification based on the remote message data payload
+     *
+     * @param context
+     */
+    fun displayNotification(context: Context) {
+        if (ActivityCompat.checkSelfPermission(
+                context,
+                Manifest.permission.POST_NOTIFICATIONS
+            ) != PackageManager.PERMISSION_GRANTED
```
ajaysubra, line 59: Curious how adding this in the manifest translates to actually having the right permissions?

image

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/Extensions.kt
evan-masseau: I wasn't too concerned about it since this is copied from how fcm handles icons

evan-masseau: I could run a test with my local just to be extra sure

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/Extensions.kt
```diff
@@ -0,0 +1,213 @@
...
+/**
+ * Parse deep link into a [Uri] if present
+ */
+val RemoteMessage.deepLink: Uri? get() = this.data[Notification.URL_KEY]?.let { Uri.parse(it) }
+
+/**
+ * Parse click action (activity or intent filter)
+ * Click action could be explicitly sent, or we should use ACTION_VIEW if a deep link is sent
+ */
+val RemoteMessage.clickAction: String?
```
jordan-griffin, line 83: Question since its kind of up in the air right now: But could there ever be a situation in the future where we have `click_action` and `deep_link` both set? Is the idea here that a deep link on the payload would always overwrite click action? I guess I'm still a bit unclear on the benefits of keeping it around right now

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/Extensions.kt
jordan-griffin: Is this icon stuff ripped from FCM?

jordan-griffin: Oh actually I assume it is cause your comment on `isValidIcon` below

jordan-griffin: And if yes, I guess the follow up question is - are there legal implications to copying that code? Like whats the license on FCM stuff

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/Extensions.kt
evan-masseau: This is written to allow both to be set

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/Extensions.kt
evan-masseau: Good question... I mean it isn't copy pasted but certainly influenced. I'll see what the license is and go from there

jordan-griffin: Oh ok, sorry I assumed it was a copy/paste from your comment. I am probably over-thinking it then

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/Extensions.kt
jordan-griffin: Yup, I totally misread this. Nevermind

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/Extensions.kt
evan-masseau: It's a valid concern for sure. I couldn't copy paste anyway since fcm is written in java. I feel like I had to make substantial enough changes to make it work for me that it is pretty distinct, even though I also feel comfortable saying we're using the same approach as they do so it should work the same.

evan-masseau: FWIW I hardcoded some FCM message payloads to send from my local, and confirmed the `small_icon` data key will work with this logic

evan-masseau: And firebase sdk is licensed under Apache License 2.0 so quite open.

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoPushService.kt
```diff
@@ -11,30 +11,37 @@ import com.klaviyo.analytics.Klaviyo
...
      *
      * @param message Remote message that has been received
      */
     override fun onMessageReceived(message: RemoteMessage) {
-        // TODO could we provide a way to display the notification in system tray?
         super.onMessageReceived(message)
+
+        if (message.isKlaviyoNotification) {
+            Notification(message).displayNotification(this)
+        }
```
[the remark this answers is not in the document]
evan-masseau: Thanks, I think I have to move them a little though, or mark them internal maybe, since I've just noticed they are visible to the test app.

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/Notification.kt
evan-masseau: Kind of like having to put that you need a permission into your `plist`, but also still needing to ask the user for that permission at runtime. This is the runtime check

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/Extensions.kt
ajaysubra: Thanks for testing this!

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoPushService.kt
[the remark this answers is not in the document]
ajaysubra: hmm, good call. On swift I tend to keep extension `fileprivate` (or `private` can't remember) if it's very specific to the file/the class it supports.

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/Notification.kt
ajaysubra: Hmm, so if I didn't add this in and prompted the user for permission from the host app, the end user wouldn't see that prompt? Just trying to understand the underlying reason for this thing.

sdk/core/src/debug/java/com/klaviyo/core/Logger.kt
```diff
@@ -45,7 +45,7 @@ open class Logger : Log {
          * this will use the class name without any anonymous class suffixes (e.g., `Foo$1`
          * becomes `Foo`).
          *
-         * Note: This will not be called if a manual tag is specified
+         * NOTE: This will not be called if a manual tag is specified
```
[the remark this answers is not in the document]
ajaysubra: Can you also make it bold with font color red? JK

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/Notification.kt
ajaysubra: Got it. If I'm making the right connection, this is like adding the entitlement in iOS.

jordan-griffin (dismissed):
My concerns have been addressed, I think this looks great!

sdk/core/src/debug/java/com/klaviyo/core/Logger.kt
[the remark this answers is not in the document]
evan-masseau: @ndurell knows what I'm talking about

jordan-griffin (approved):

jordan-griffin (approved):

PR #57, opened 2023-03-23
Takes out my hacky attempt to "disassociate" push tokens on reset profile

evan-masseau:
# Purpose

**Type of PR**
* [x] Bug Fix
* [ ] Feature Work
* [ ] Revert
* [ ] Refactor / Code Cleanup
* [ ] Other (fill in...)

## Short Description

The reset profile functionality now matches iOS, doesn't create dangling anonymous profiles, but it does clear push token from state at least. 

## Test Plan

Updated a unit test, and I tested android and ios apps together and they behave the same now.

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
@@ -238,8 +238,8 @@ object Klaviyo {
     /**
      * Clears all stored profile identifiers (e.g. email or phone) and starts a new tracked profile
      *
-     * NOTE: if a push token was registered to the current profile, Klaviyo will disassociate it
-     * from the current profile. Call `setPushToken` again to associate this device to a new profile
+     * NOTE: if a push token was registered to the current profile, you will need to
+     * call `setPushToken` again to associate this device to a new profile
      *
      * This should be called whenever an active user in your app is removed
      * (e.g. after a logout)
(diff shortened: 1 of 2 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -252,16 +252,8 @@ object Klaviyo {
         // Clear profile identifiers from state
         UserInfo.reset()
 
-        // If we had a push token, re-associate it with an anonymous ID, then erase the local copy
```
ajaysubra, line 255: Do the comments above need to be updated? Mainly the note

>      * NOTE: if a push token was registered to the current profile, Klaviyo will disassociate it
>      * from the current profile. Call `setPushToken` again to associate this device to a new profile

ajaysubra (approved):

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
evan-masseau: ah good catch, and maybe there's mention in the readme too

evan-masseau: thanks, fixed, and no change was needed in the readme

jordan-griffin (approved):

PR #64, opened 2023-04-03
Update issue templates

evan-masseau:
Updated the existing issue template to be for bugs, added one for enhancements.
We already have a good template for PRs, IMO.


jordan-griffin (commented):
I think Noah was working on the iOS version of this ticket and had some ideas for something a bit more involved. If you can, might be good to just sync with him tomorrow on this to at least get an idea of consistency if nothing else!

evan-masseau:
@jordan-griffin oh I see he made like a legit form input, very cool

PR #73, opened 2023-05-24
Add User-Agent to headers

kennyklaviyo:
# Description

Introduces a custom `User-Agent` header to all API requests
# Check List

### DIFF
--- a/sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
@@ -77,6 +77,7 @@ object KlaviyoConfig : Config {
     override var baseUrl: String = BuildConfig.KLAVIYO_SERVER_URL
         private set
     override lateinit var apiKey: String private set
+    override lateinit var userAgent: String private set
     override lateinit var applicationContext: Context private set
     override var debounceInterval = DEBOUNCE_INTERVAL
         private set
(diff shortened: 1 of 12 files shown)


sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
```diff
@@ -214,6 +211,26 @@ object KlaviyoConfig : Config {
...
+internal fun PackageInfo.assertRequiredPermissions(requiredPermissions: Array<String>) {
+    val permissions = requestedPermissions?.toSet() ?: emptySet()
+    requiredPermissions.firstOrNull { it !in permissions }?.let { throw MissingPermission(it) }
+}
+
+internal fun buildUserAgent(context: Context, packageInfo: PackageInfo): String {
+    val versionCode = packageInfo.getVersionCode()
+    val applicationName = context.packageManager.getApplicationLabel(context.applicationInfo)
+
+    return "$applicationName/${packageInfo.versionName} (${packageInfo.packageName}; build:$versionCode; Android ${Build.VERSION.SDK_INT}) klaviyo-android/${BuildConfig.VERSION}"
```
evan-masseau, line 223: super nit, but is there a way to break this down so it isn't a super long line? Like maybe
```
"%s/%s (%s; build:%s ...etc".format(
   applicationName,
   versionName,
   etc...
 )
```
IDK which is more readable anyways.

evan-masseau (approved):

evan-masseau:
We might want to put this into a `1.0.2` release sooner than later, given we'll want to hold off on releasing the metadata addition to events until serverside events are ready to update too.

PR #74, opened 2023-05-24
Update Workflows

kennyklaviyo:
# Description

Use Ubuntu 22.04 and v3 of setup-java action and include ktlint check as job
# Check List


.github/workflows/android-feature.yml
```diff
@@ -29,4 +29,25 @@ jobs:
...
+      - name: Set up JDK 11
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu'
+          java-version: '11'
+          java-package: jdk
+
+      # Call Gradle to run ktlint check
+      - name: KtlintCheck
+        run: bash ./gradlew ktlintCheck
```
evan-masseau, line 53: good call

evan-masseau (approved):

PR #80, opened 2023-06-22
V3 API updates and Device Metadata

evan-masseau:
# Description
- Adds device metadata collection
- Tag all events created by Swift SDK with device metadata.
- Upgrades from using the V2 identify API for push tokens to new V3 Push Tokens API. 
- [x] TODO - Finalize V3 APIs payload structure (e.g. meta.identifiers being removed in 7-15)
- [x] TODO - Set final revision value

# Check List

- [x] Are you changing anything with the public API? No
- [x] Have you tested this change on real device? Yes
- [x] Have you added unit test coverage for your changes? Yes
- [x] Have you verified that your changes are compatible with all Android versions the SDK currently supports? Yes

## Test Plan

New unit tests
Android tester app


sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -40,14 +40,15 @@ internal open class KlaviyoApiRequest(
...
+        const val PUSH_TOKEN = "push-token"
 
         // Headers
         const val HEADER_CONTENT = "Content-Type"
         const val HEADER_USER_AGENT = "User-Agent"
         const val HEADER_ACCEPT = "Accept"
         const val HEADER_REVISION = "Revision"
         const val TYPE_JSON = "application/json"
-        const val V3_REVISION = "2023-01-24"
+        const val V3_REVISION = "2023-06-06.pre"
```
kennyklaviyo, line 51: feels weird to be using a `pre` revision of the api

sdk/core/src/main/java/com/klaviyo/core/model/SharedPreferencesDataStore.kt
```diff
@@ -77,6 +77,20 @@ internal object SharedPreferencesDataStore : DataStore {
...
+     * @param key
+     * @param fallback
+     */
+    override fun fetchOrCreate(key: String, fallback: () -> String) = fetch(key)
+        .let { it ?: "" }
+        .ifEmpty {
+            val value = fallback()
+            store(key, value)
+            value
+        }
```
kennyklaviyo, line 92: feels a bit clunky, how about:
```
    override fun fetchOrCreate(key: String, fallback: () -> String): String {
        return fetch(key) ?: fallback().also {
            store(key, it)
        }
    }
```

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
```diff
@@ -19,42 +17,49 @@ internal class PushTokenApiRequest(
...
-                PROPERTIES to JSONObject(properties)
+        body = jsonMapOf(
+            DATA to mapOf(
+                TYPE to PUSH_TOKEN,
+                ATTRIBUTES to filteredMapOf(
+                    "token_id" to token,
+                    "platform" to DeviceProperties.platform,
+                    "vendor" to "FCM",
+                    "enablement_status" to "AUTHORIZED",
+                    "background" to "AVAILABLE",
```
kennyklaviyo, line 54: is it worth making these a const in this class or somewhere else?

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
evan-masseau: 💯, can't release this till the tokens API is public.

sdk/fixtures/src/main/java/com/klaviyo/fixtures/InMemoryDataStore.kt
```diff
@@ -22,6 +22,20 @@ class InMemoryDataStore : DataStore {
         store.remove(key)
     }
 
+    /**
+     * Uses fetch/store implementations above
+     *
+     * @param key
+     * @param fallback
+     */
+    override fun fetchOrCreate(key: String, fallback: () -> String) = fetch(key)
```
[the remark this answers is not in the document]
evan-masseau: Good point, I could add an abstract class that they both inherit from instead of straight from the interface. Or maybe an extension method ...

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
evan-masseau: Also, I shouldn't be hardcoding to authorized and available! I think I meant to circle back to that

evan-masseau: Fixed: added proper fetching of background and enablement statuses, and extracted constants here

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -40,14 +40,15 @@ internal open class KlaviyoApiRequest(
...
+        const val PUSH_TOKEN = "push-token"
 
         // Headers
         const val HEADER_CONTENT = "Content-Type"
         const val HEADER_USER_AGENT = "User-Agent"
         const val HEADER_ACCEPT = "Accept"
         const val HEADER_REVISION = "Revision"
         const val TYPE_JSON = "application/json"
-        const val V3_REVISION = "2023-01-24"
+        const val V3_REVISION = "2023-07-15.pre"
```
ajaysubra, line 51: do you need to remove the pre suffix for the release?

ajaysubra (dismissed):

jordan-griffin (approved):

Issue #78, opened 2023-06-19
Manually logging OPENED_PUSH not showing on klaviyo's dashboard


iballan:
### Description

We are trying to integrate Klaviyo's notification into project that already uses Firebase Messaging.
In `FirebaseMessagingService`'s onMessageReceived we customize how the notification looks, so we are not able to use `KlaviyoNotification` to display notification.

So we extract the information from Klaviyo's notification and display.

        val isKlaviyoNotification = remoteMessage.isKlaviyoNotification
        val title = if (isKlaviyoNotification) remoteMessage.title else remoteMessage.notification?.title
        val body = if (isKlaviyoNotification) remoteMessage.body else remoteMessage.notification?.body
        val deepLinkUri = if (isKlaviyoNotification) remoteMessage.deepLink else remoteMessage.data[DEEP_LINK_KEY]?.toUri()
// ... displaying notification with custom intent

Then when notification clicked, if it is Klaviyo's notification we are trying to track it like this:
```kotlin
            if (isKlaviyo) {
                storage.getPushNotificationsToken()?.let { token ->
                    klaviyoManager.logOpenedPush(token, id.orEmpty())
                }
            }
```

Tracking method:
```kotlin
    fun logOpenedPush(pushToken: String, notificationId: String) {
        val event = Event(EventType.OPENED_PUSH)
            .setProperty(EventKey.PUSH_TOKEN, pushToken)
        if (notificationId.isNotEmpty()) {
            event.setProperty(EventKey.EVENT_ID, notificationId)
        }
        klaviyo.createEvent(event)
    }
```

I couldn't find any document related to loggin EventType.OPENED_PUSH manually except in the code of EventKey:
```kotlin
/**
 * All event property keys recognised by the Klaviyo APIs
 * Custom properties can be defined using the [CUSTOM] inner class
 */
sealed class EventKey(name: String) : Keyword(name) {
    object EVENT_ID : EventKey("\$event_id")
    object VALUE : EventKey("\$value")

    /**
     * For [EventType.OPENED_PUSH] events, append the device token as an event property
     */
    object PUSH_TOKEN : EventKey("push_token")

    class CUSTOM(propertyName: String) : EventKey(propertyName)
}
```

So I added push_token as shown above in my tracking opened push

### Checklist

- [x] I have determined whether this bug is also reproducible in a vanilla Android project
- [x] If possible, I've reproduced the issue using the `main` branch of this package.
[...]

PR #81, opened 2023-06-23
Helper methods for developers who use a custom workflow to handle notifications

evan-masseau:
# Description

Created extension methods as helpers for displaying klaviyo notifications and handling the opened event.

# Check List

- [x] Are you changing anything with the public API? -- No, but this adds extension methods to `Intent` that devs can acces
- [x] Have you tested this change on real device? Yes
- [x] Have you added unit test coverage for your changes? Yes - updated existing coverage
- [x] Have you verified that your changes are compatible with all Android versions the SDK currently supports? Yes

## Changelog / Code Overview

- extension method to `Intent` for passing on `RemoteMessage` data in the intent
- extension property to identify whether an intent has klaviyo tracking params
- the opened push event keys `internal`

## Test Plan

- Tested with flow from prod to confirm notifications still display properly and opens log properly.

## Related Issues/Tickets

#78

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
@@ -289,16 +289,15 @@ object Klaviyo {
             return@apply
         }
 
-        val payload = intent.extras?.let { extras ->
-            extras.keySet().associateWith { key -> extras.getString(key, "") }
-        } ?: emptyMap()
+        val event = Event(EventType.OPENED_PUSH)
+        val extras = intent.extras
 
-        val event = Event(
-            EventType.OPENED_PUSH,
-            payload.mapKeys {
-                EventKey.CUSTOM(it.key)
+        extras?.keySet()?.forEach { key ->
+            if (key.contains("com.klaviyo")) {
+                val eventKey = EventKey.CUSTOM(key.replace("com.klaviyo.", ""))
+                event[eventKey] = extras.getString(key, "")
             }
-        )
+        }
 
         getPushToken()?.let { event[EventKey.PUSH_TOKEN] = it }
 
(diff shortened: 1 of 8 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -284,24 +284,25 @@ object Klaviyo {
      * @param intent the [Intent] from opening a notification
      */
     fun handlePush(intent: Intent?) = apply {
-        val payload = intent?.extras?.let { extras ->
+        if (intent?.isKlaviyoIntent != true) {
+            Registry.log.info("Non-klaviyo intent ignored")
```
kennyklaviyo, line 288: nit:
```suggestion
            Registry.log.info("Non-Klaviyo intent ignored")
```

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoRemoteMessage.kt
```diff
@@ -23,6 +23,20 @@ import com.klaviyo.core.Registry
...
+    /**
+     * Append requisite data from a remote message to an intent
+     * for displaying a notification
+     *
+     * @param message
+     */
+    fun Intent.appendKlaviyoExtras(message: RemoteMessage) = apply {
+        if (message.isKlaviyoMessage) {
+            message.data.forEach {
+                this.putExtra(it.key, it.value)
```
kennyklaviyo, line 35: was looking at the docstring for `putExtra`:
```
    /**
     * Add extended data to the intent.  The name must include a package
     * prefix, for example the app com.android.contacts would use names
     * like "com.android.contacts.ShowAll".
     *
     * @param name The name of the extra data, with package prefix.
     * @param value The String data value.
     *
     * @return Returns the same Intent object, for chaining multiple calls
     * into a single statement.
     *
     * @see #putExtras
     * @see #removeExtra
     * @see #getStringExtra(String)
     */
    public @NonNull Intent putExtra(String name, @Nullable String value) {
```
do you think the package name prefix part is very important?

evan-masseau: Good catch!

evan-masseau: Updated to use namespaced keys `com.klaviyo.*` and only send our keys from the intent also

ajaysubra (approved):

ajaysubra (dismissed):

ajaysubra (dismissed):

ajaysubra (approved):

PR #82, opened 2023-06-27
Updates to 7-15 revision data structures

evan-masseau:
Here is what I think the API payloads need to look like for upcoming 7-15 V3 API changes.
Note the target branch! This is on top of this pending PR #80 
I tested these changes against prod on the VPN so that I could target 7-15.pre revision

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/DeviceProperties.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/DeviceProperties.kt
@@ -49,7 +49,8 @@ internal object DeviceProperties {
     }
 
     val sdkName: String by lazy {
-        "klaviyo-android-sdk"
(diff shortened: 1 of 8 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/DeviceProperties.kt
```diff
@@ -49,7 +49,8 @@ internal object DeviceProperties {
     }
 
     val sdkName: String by lazy {
-        "klaviyo-android-sdk"
+//        "klaviyo-android-sdk"
```
evan-masseau, line 52: temporary till a server fix is deployed

jordan-griffin (approved):
Looks good, seems to match up perfectly with the expected payload for the V3 API as far as I can tell!

PR #84, opened 2023-06-30
1.2.0-beta version increment

evan-masseau:


klaviyoit:
https://klaviyo.tpondemand.com/entity/1

kennyklaviyo (dismissed):
update readme version too

ajaysubra (approved):

versions.gradle
```diff
@@ -12,8 +12,8 @@ ext {
     targetSDKVersion = 33
 
     // project versioning
-    versionCode = 4
-    versionName = '1.1.1'
+    versionCode = 6
+    versionName = '1.2.0-beta1'
```
kennyklaviyo, line 16: if you want to do a beta version, https://semver.org/#spec-item-9 should follow the semantic versioning for it, should be `1.2.0-beta.1`

versions.gradle
```diff
@@ -12,8 +12,8 @@ ext {
     targetSDKVersion = 33
 
     // project versioning
-    versionCode = 4
-    versionName = '1.1.1'
+    versionCode = 6
```
kennyklaviyo, line 15: why does this go up 2?

versions.gradle
ajaysubra: Thanks for finding the semver on this. I should have looked it up but I followed an example from the cocoapods doc. I will leave iOS as is for now and update android to not have the number so `1.2.0-beta` and if we have more patches on the beta related to rich push we can append the number.

image

versions.gradle
ajaysubra: oops my bad. I was looking at the diff which Evan had to 5 and instead of leaving it as is incremented! Fixed this now!

README.md
```diff
@@ -349,9 +349,9 @@ refer to [android developer documentation](https://developer.android.com/trainin
 
 ### Rich push notification 
 
-In order for rich push notifications to work on your app, your app needs to be on or above version 1.2.0 of the Klaviyo SDK
+In order for rich push notifications to work on your app, your app needs to be on or above version 1.2.0-beta1 of the Klaviyo SDK
```
kennyklaviyo, line 352:
```suggestion
In order for rich push notifications to work on your app, your app needs to be on or above version 1.2.0-beta of the Klaviyo SDK
```

ajaysubra: Thank you!

ajaysubra: actually I think I need to fix this to be `In order for rich push notifications to work on your app, your app needs to be on version 1.2.0-beta`

ajaysubra: Done.

kennyklaviyo (approved):

PR #88, opened 2023-07-10
Added in-repo reminders about versioning procedures.

evan-masseau:
# Description
Since I forgot both to update the README and to generate 1.1.1 dokka docs, thought I would add an inline reminder about those things.


README.md
```diff
@@ -352,6 +352,7 @@ refer to [android developer documentation](https://developer.android.com/trainin
 In order for rich push notifications to work on your app, your app needs to be on version 1.2.0-beta of the Klaviyo SDK. This feature is currently only available for a focus group and should be available to everyone else fairly soon. 
 
 ## Code Documentation
+[//]: # (REMINDER: Update documentation branch with new dokka docs when updating version number)
```
ajaysubra, line 355: Thanks for adding this in. When I read the confluence doc I assumed that since I wasn't updating public APIs I didn't have to generate any new docs but this makes sense at least from keeping the version number update to date.

Is there a recommendation for pre release versions?

ajaysubra (approved):

README.md
evan-masseau: Maybe we should link to the beta dokka docs in that beta release notes?

ajaysubra: no harm I guess in just linking it in every release notes irrespective of it there is a API change or not.

evan-masseau: Yeah, dokka captures doc blocks and other stuff, so its worth updating for every release. besides, a user would't necessarily know whether there's documentation-worthy changes in 1.1.0 vs 1.1.1 and would rather look at docs for version 1.1.1 if thats the one they are using. Even if it is the same exact docs.

PR #96, opened 2023-08-30
Close stale issues via github action

jordan-griffin:
# Description

Just adds an update to our Github Actions workflow to mark issues as stale and close them out after a period of inactivity

# Check List

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -22,8 +22,12 @@ jobs:
     - uses: actions/stale@v5
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        days-before-issue-stale: -1
-        days-before-issue-close: -1
+        days-before-issue-stale: 14
+        days-before-issue-close: 10
+        stale-issue-label: 'stale'
+        stale-issue-message: 'This issue is stale due to inactivity. Please let us know if you are still encountering issues, otherwise this issue will automatically close in 4 days.'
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -22,8 +22,12 @@ jobs:
     - uses: actions/stale@v5
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        days-before-issue-stale: -1
-        days-before-issue-close: -1
+        days-before-issue-stale: 14
+        days-before-issue-close: 10
+        stale-issue-label: 'stale'
+        stale-issue-message: 'This issue is stale due to inactivity. Please let us know if you are still encountering issues, otherwise this issue will automatically close in 4 days.'
```
kennyklaviyo, line 28: if we want to close in 4 days, then line 26 should be 4 not 10

jordan-griffin: 4 might be too short. If an issue is marked stale at the start of the weekend, it could be closed before anyone has a chance to respond to it. I totally chose 10 days pretty arbitrarily though

kennyklaviyo: oh sorry i dont have a preference on how long, just that this message says 4 days but the number above isnt 4

jordan-griffin: Oooh I totally missed that. Will change the message for now. Probably the same on the Swift SDK so will correct that too

PR #98, opened 2023-09-22
Ensure SDK sends device_id properly

ndurell:
# Description
Device id was being returned as "" which resulted in it not being set properly.

# Check List

## Test Plan
Tested locally with changes to ensure device gets perisisted.

### DIFF
--- a/sdk/core/src/main/java/com/klaviyo/core/model/DataStore.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/model/DataStore.kt
@@ -56,4 +56,6 @@ interface DataStore {
  * @param fallback
  */
 fun DataStore.fetchOrCreate(key: String, fallback: () -> String): String =
-    fetch(key) ?: fallback().also { store(key, it) }
+    fetch(key) ?: fallback().also {
+        store(key, it)
+    }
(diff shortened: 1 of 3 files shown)


klaviyoit:
https://klaviyo.tpondemand.com/entity/199177

sdk/core/src/main/java/com/klaviyo/core/model/DataStore.kt
```diff
@@ -56,4 +56,6 @@ interface DataStore {
  * @param fallback
  */
 fun DataStore.fetchOrCreate(key: String, fallback: () -> String): String =
-    fetch(key) ?: fallback().also { store(key, it) }
+    fetch(key) ?: fallback().also {
```
ajaysubra, line 59: there is no change here no?

sdk/core/src/main/java/com/klaviyo/core/model/SharedPreferencesDataStore.kt
```diff
@@ -74,7 +74,7 @@ internal object SharedPreferencesDataStore : DataStore {
      * @return The value read from the shared preferences for the given key
      */
     override fun fetch(key: String): String? {
-        return openSharedPreferences().getString(key, "")
+        return openSharedPreferences().getString(key, null)
```
ajaysubra, line 77: 👍🏽 nice find.

ndurell, line 77: We should probably check if this is safe for all fetch calls.

ajaysubra (approved):

jordan-griffin (approved):

PR #105, opened 2024-01-05
Change Event.value from string to double

evan-masseau:
Minor fix to change the type of event.value. Also corrected some syntax in Profile model

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/model/Event.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/model/Event.kt
@@ -17,7 +17,7 @@ class Event(val type: EventType, properties: Map<EventKey, Serializable>?) :
 
     constructor(type: String) : this(type, null)
 
-    fun setValue(value: Double) = apply { this.value = value }
+    fun setValue(value: Double?) = apply { this.value = value }
     var value: Double?
         get() = when (val value = this[EventKey.VALUE]) {
             is Double -> value
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/model/Event.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/model/Event.kt
@@ -17,9 +17,9 @@ class Event(val type: EventType, properties: Map<EventKey, Serializable>?) :
 
     constructor(type: String) : this(type, null)
 
-    fun setValue(value: String) = apply { this.value = value }
-    var value: String
-        get() = (this[EventKey.VALUE] ?: "") as String
+    fun setValue(value: Double) = apply { this.value = value }
+    var value: Double?
+        get() = this[EventKey.VALUE] as? Double
         set(value) {
             this[EventKey.VALUE] = value
         }
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/model/Event.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/model/Event.kt
@@ -19,16 +19,22 @@ class Event(val type: EventType, properties: Map<EventKey, Serializable>?) :
 
     fun setValue(value: Double) = apply { this.value = value }
     var value: Double?
-        get() = this[EventKey.VALUE] as? Double
+        get() = when (val value = this[EventKey.VALUE]) {
+            is Double -> value
+            else -> try {
+                value.toString().toDouble()
+            } catch (e: NumberFormatException) {
+                null
+            }
+        }
         set(value) {
             this[EventKey.VALUE] = value
         }
 
-    override fun setProperty(key: EventKey, value: Serializable) = apply {
+    override fun setProperty(key: EventKey, value: Serializable?) = apply {
         this[key] = value
     }
 
-    override fun setProperty(key: String, value: Serializable) = apply {
-        this[EventKey.CUSTOM(key)] = value
-    }
+    override fun setProperty(key: String, value: Serializable?) =
+        setProperty(EventKey.CUSTOM(key), value)
 }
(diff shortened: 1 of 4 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/model/Profile.kt
```diff
@@ -12,28 +12,28 @@ class Profile(properties: Map<ProfileKey, Serializable>?) :
 
     fun setExternalId(identifier: String) = apply { this.externalId = identifier }
     var externalId: String?
-        get() = (this[ProfileKey.EXTERNAL_ID]) as String?
+        get() = this[ProfileKey.EXTERNAL_ID] as? String
```
ndurell, line 15: Hmm would this have failed if you passed in a non string before?

sdk/analytics/src/main/java/com/klaviyo/analytics/model/Event.kt
```diff
@@ -17,9 +17,9 @@ class Event(val type: EventType, properties: Map<EventKey, Serializable>?) :
 
     constructor(type: String) : this(type, null)
 
-    fun setValue(value: String) = apply { this.value = value }
-    var value: String
-        get() = (this[EventKey.VALUE] ?: "") as String
+    fun setValue(value: Double) = apply { this.value = value }
```
kennyklaviyo, line 20: should we allow setting null (clearing the value)?

ndurell (approved):

sdk/analytics/src/main/java/com/klaviyo/analytics/model/Event.kt
evan-masseau: tru

sdk/analytics/src/main/java/com/klaviyo/analytics/model/Profile.kt
kennyklaviyo:
```
print(3 as String?)
> java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.String (java.lang.Integer and java.lang.String are in module java.base of loader 'bootstrap')
	at Line_1.<init>(Line_1.kts:1)

print(3 as? String)
> null
```

kennyklaviyo: yeah looks like it would have raised

sdk/analytics/src/main/java/com/klaviyo/analytics/model/Event.kt
kennyklaviyo: after looking at the profile model, should those setter methods also allow null?

evan-masseau: Agreed. Also thinking we might need type validation for those keys in the `setProperty` method so you can't work around the typing that way.

evan-masseau: 🎶 edge cases 🎶

kennyklaviyo: could maybe use `.toString` instead of casting in some cases, seems more graceful or yeah we can enforce types more strictly

evan-masseau: yeah, it already has to be serializable, so toString is probably sufficient and simpler

sdk/analytics/src/main/java/com/klaviyo/analytics/model/Event.kt
```diff
@@ -17,18 +17,24 @@ class Event(val type: EventType, properties: Map<EventKey, Serializable>?) :
...
+    var value: Double?
+        get() = when (val value = this[EventKey.VALUE]) {
+            is Double -> value
+            else -> try {
```
evan-masseau, line 24: Alternative would be to have a case for each Int, Float, etc.. this just looks cleaner 🤷

sdk/analytics/src/main/java/com/klaviyo/analytics/model/Profile.kt
```diff
@@ -10,51 +10,40 @@ class Profile(properties: Map<ProfileKey, Serializable>?) :
...
-    }
+    override fun setProperty(key: String, value: Serializable?) =
+        this.setProperty(ProfileKey.CUSTOM(key), value)
 
     override fun merge(other: Profile) = apply { super.merge(other) }
-
-    /**
-     * Get a map of just the unique identifiers of this profile
-     */
-    internal fun getIdentifiers(): Map<ProfileKey, String> = mapOf(
```
evan-masseau, line 54: ktlint informed me this was unused

sdk/analytics/src/test/java/com/klaviyo/analytics/model/ModelTests.kt
```diff
@@ -79,21 +79,44 @@ internal class ModelTests : BaseTest() {
         assertEquals("custom", profileMap["custom"])
     }
 
+    @Test
```
evan-masseau, line 82: I think this behavior is good enough™️ , but an alternative would be to make all these named properties _actual properties_ rather than being backed by the overload dict. Then we'd have to override the the `toMap` function as well. Thats a considerably bigger change to the public API though.

sdk/analytics/src/main/java/com/klaviyo/analytics/model/Event.kt
kennyklaviyo: looks like you missed updating this method to accept null

evan-masseau: certainly did. fixed now!

ajaysubra (approved):

kennyklaviyo (approved):

evan-masseau:
Oof, accidentally deleted this before merging

PR #107, opened 2024-01-10
Fix event metric names - backward compatible

evan-masseau:
# Description

This PR adds new `EventMetric` class with corrected/removed metrics, and deprecates the old one. I used a superclass so that the old one is still backwards compatible. I've tested with the test app, and there's no breaking changes. Deprecation warnings provide guidance, and there is a migration guide with an explanation. I introduce _new_ non-breaking API methods, but expect 2.0 to have breaking API changes including argument re-labelling.

# Check List

- [x] Are you changing anything with the public API? - YES, but only new methods in this PR
- [x] Are your changes backwards compatible with previous SDK Versions? - YES
- [ ] Have you tested this change on real device? - 50% - I have tested the test app as-is. Will also update test app according to migration guide and test again. 
- [x] Have you added unit test coverage for your changes? - YES
- [x] Have you verified that your changes are compatible with all Android versions the SDK currently supports? - No android compatibility concerns

## Changelog / Code Overview

- `EventMetric` superclass will replace `EventType`
- `EventType` is marked deprecated
- Added migration guide with explanation and instructions
- Where possible, added new overloads that explicitly take the new `EventMetric` with new argument labels, to reduce friction when upgrading to 2.0 later on.
- Introduces some missing event types like Added to Cart (`EventMetric` only)

## Test Plan

Test with test app as-is.
Update test app according to deprecation warnings and test again. 

## Related Issues/Tickets
https://github.com/klaviyo/klaviyo-swift-sdk/issues/124
https://github.com/klaviyo/klaviyo-swift-sdk/issues/119

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,12 +135,12 @@ Klaviyo.setEmail("robin@example.com")
 ### Tracking Events
 
 The SDK also provides tools for tracking analytics events to the Klaviyo API.
-A list of common Klaviyo-defined event names is provided in [EventType](https://klaviyo.github.io/klaviyo-android-sdk/1.0.0/sdk/analytics/com.klaviyo.analytics.model/-event-type/), or
-you can use `EventType.CUSTOM("name")` for custom event names.
+A list of common Klaviyo-defined event names is provided in [MetricName](https://klaviyo.github.io/klaviyo-android-sdk/1.0.0/sdk/analytics/com.klaviyo.analytics.model/-event-type/), or
+you can use `MetricName.CUSTOM("name")` for custom event names.
 Additional event properties can be specified as part of `EventModel`
 
 ```kotlin
-val event = Event(EventType.VIEWED_PRODUCT)
+val event = Event(MetricName.VIEWED_PRODUCT)
     .setProperty(EventKey.VALUE, "10")
     .setProperty(EventKey.CUSTOM("custom_key"), "value")
 Klaviyo.createEvent(event)
--- a/README.md
+++ b/README.md
@@ -135,12 +135,12 @@ Klaviyo.setEmail("robin@example.com")
 ### Tracking Events
 
 The SDK also provides tools for tracking analytics events to the Klaviyo API.
-A list of common Klaviyo-defined event names is provided in [MetricName](https://klaviyo.github.io/klaviyo-android-sdk/1.0.0/sdk/analytics/com.klaviyo.analytics.model/-event-type/), or
-you can use `MetricName.CUSTOM("name")` for custom event names.
+A list of common Klaviyo-defined event metrics is provided in `EventMetric`, or
+you can use `EventMetric.CUSTOM("name")` for custom event metric names.
 Additional event properties can be specified as part of `EventModel`
 
 ```kotlin
-val event = Event(MetricName.VIEWED_PRODUCT)
+val event = Event(EventMetric.VIEWED_PRODUCT)
     .setProperty(EventKey.VALUE, "10")
     .setProperty(EventKey.CUSTOM("custom_key"), "value")
 Klaviyo.createEvent(event)
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
@@ -5,7 +5,7 @@ import android.content.Context
 import android.content.Intent
 import com.klaviyo.analytics.model.Event
 import com.klaviyo.analytics.model.EventKey
-import com.klaviyo.analytics.model.EventType
+import com.klaviyo.analytics.model.MetricName
 import com.klaviyo.analytics.model.Profile
 import com.klaviyo.analytics.model.ProfileKey
 import com.klaviyo.analytics.networking.ApiClient
@@ -272,13 +272,13 @@ object Klaviyo {
      *
      * Convenience method for creating an event with no other properties
      *
-     * @param eventType [EventType] to create
+     * @param eventType [MetricName] to create
      * @return Returns [Klaviyo] for call chaining
      */
-    fun createEvent(eventType: EventType): Klaviyo = createEvent(Event(eventType))
+    fun createEvent(eventType: MetricName): Klaviyo = createEvent(Event(eventType))
 
     /**
-     * From an opened push Intent, creates an [EventType.OPENED_PUSH] [Event]
+     * From an opened push Intent, creates an [MetricName.OPENED_PUSH] [Event]
      * containing appropriate tracking parameters
      *
      * @param intent the [Intent] from opening a notification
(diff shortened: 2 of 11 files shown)


sdk/analytics/src/test/java/com/klaviyo/analytics/model/KeywordsTest.kt
```diff
@@ -60,10 +67,16 @@ class KeywordsTest {
...
         assertEquals(expectedCustomKey, EventType.CUSTOM(expectedCustomKey).name)
 
         // Test the equals operator works properly on custom keys
-        val custom = EventType.CUSTOM(expectedCustomKey)
+        var custom: EventMetric = EventMetric.CUSTOM(expectedCustomKey)
+        assert(custom == EventMetric.CUSTOM(expectedCustomKey))
+        assert(custom != EventMetric.CUSTOM(expectedCustomKey + "1"))
+
+        // Test the equals operator works properly on custom keys
+        custom = EventType.CUSTOM(expectedCustomKey)
```
jordan-griffin, line 79: Other tests seem to rip out `EventType` entirely. Is there a reason this test in particular is testing the deprecated `EventType` as well? I feel like if we want to keep tests for some deprecated logic around it may be better doing it in a standalone test separate from the test checking `EventMetrics`

jordan-griffin: Total nit since this test doesnt really do much in the first place

evan-masseau: The v2 pull request does rip out the `EventType` tests bc that PR is for the breaking changes where I'm deleting `EventType`.

jordan-griffin: Just referring to the other tests like those in `KlaviyoTest` and `KlaviyoApiClientTest` where the use of `EventType` was replaced. But ya, they are literally just changing which object is called - its otherwise identical logic - so it really doesn't matter.

But ya its just a change in which object is called so thinking about it more I really dont think it matters

jordan-griffin (dismissed):

sdk/analytics/src/test/java/com/klaviyo/analytics/model/KeywordsTest.kt
evan-masseau: Ooh I see what you mean, yeah that's actually a good point I'm probably missing test coverage now.

evan-masseau: Gonna fix that juuust to be safe.

evan-masseau: Added back a test against EventType in `KlaviyoTest` so coverage is 100% on event model initializers. Don't need to test both paths in the networking test stubs

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -275,10 +286,14 @@ object Klaviyo {
...
      */
+    @Deprecated(
+        "EventType is deprecated in favor of EventMetric. See migration guide for details.",
+        ReplaceWith("com.klaviyo.analytics.Klaviyo.createEvent(metric)")
+    )
     fun createEvent(eventType: EventType): Klaviyo = createEvent(Event(eventType))
 
     /**
-     * From an opened push Intent, creates an [EventType.OPENED_PUSH] [Event]
+     * From an opened push Intent, creates an [EventMetric.OPENED_PUSH] [Event]
```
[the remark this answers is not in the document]
ndurell: nvm!

README.md
```diff
@@ -135,12 +135,12 @@ Klaviyo.setEmail("robin@example.com")
 ### Tracking Events
 
 The SDK also provides tools for tracking analytics events to the Klaviyo API.
-A list of common Klaviyo-defined event names is provided in [EventType](https://klaviyo.github.io/klaviyo-android-sdk/1.0.0/sdk/analytics/com.klaviyo.analytics.model/-event-type/), or
-you can use `EventType.CUSTOM("name")` for custom event names.
+A list of common Klaviyo-defined event metrics is provided in `EventMetric`, or
```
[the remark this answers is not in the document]
evan-masseau: Hm lost a git comment. 
Yeah I took the link out because we were only doing this link to documentation in one place, thought we should take a consistent approach. Maybe when we re-write the readme we can re-consider. It is just one more thing to remember at release time too.

ajaysubra (approved):

sdk/analytics/src/main/java/com/klaviyo/analytics/model/Event.kt
```diff
@@ -5,18 +5,24 @@ import java.io.Serializable
 /**
  * Controls the data that can be input into a map of event attributes recognised by Klaviyo
  */
-class Event(val type: EventType, properties: Map<EventKey, Serializable>?) :
```
ndurell, line 8: Is this public should you have a deprecated version of this?

ndurell (approved):

README.md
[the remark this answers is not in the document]
jordan-griffin: We also have that epic in Jira we've backlogged for a long time now where we retro'd on our readmes and had a whole plan to rewrite everything to encourage consistency between iOS and Android setup steps. Assuming we can get around to actually implementing some of that in time, maybe small decisions here aren't super super important right now.

sdk/analytics/src/main/java/com/klaviyo/analytics/model/EventMetric.kt
```diff
@@ -0,0 +1,67 @@
+package com.klaviyo.analytics.model
+
+/**
+ * Common clientside event metrics recognized by Klaviyo
+ * Custom metrics can be defined with the [CUSTOM] inner class
+ *
+ * @property name String that represents the name of the metric
+ */
+sealed class EventMetric(name: String) : Keyword(name) {
```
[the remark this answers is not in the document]
evan-masseau: I think the slightly different naming convention is OK. On iOS the enum is inside a struct called Event, on Android all the models are in one namespace so I opted to keep `Event` in the name to indicate it belongs to Events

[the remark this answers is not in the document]
ndurell: Sounds good to me!

sdk/analytics/src/main/java/com/klaviyo/analytics/model/Event.kt
evan-masseau: Oh good idea. I did try that out just now but I can't get that to compile. However, this signature change is not breaking, since EventMetric is a superclass of EventType. It just means their code will only show a deprecated warning on the EventType constructor, but not for the Event constructor.

README.md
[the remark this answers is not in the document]
evan-masseau: Exactly

jordan-griffin (approved):

PR #108, opened 2024-01-10
Fix nullable model properties

evan-masseau:
# Description

Fixes a potential edge case that could cause a crash in case of type mismatch for any of these named properties. Discovered while changing `Event.value` from string to double.

This can go into the next minor release

# Check List

- [x] Are you changing anything with the public API? - Yes but adding nullability is backward compatible.
- [x] Have you tested this change on real device? - With the test app

## Changelog / Code Overview

- Added nullability to all the setter methods in our models, to be more consistent
- Fixed the type casting on any named properties so that they can't cause a crash in case of type mismatch

## Test Plan

Added unit tests

## Related Issues/Tickets

CHNL-4540

### DIFF
--- a/sdk/fixtures/src/main/java/com/klaviyo/fixtures/BaseTest.kt
+++ b/sdk/fixtures/src/main/java/com/klaviyo/fixtures/BaseTest.kt
@@ -97,6 +97,9 @@ abstract class BaseTest {
     /**
      * Gross way to modify a final static field through reflection
      *
+     * NOTE This will break in future Java versions, but is the only way to mock
+     * SDK version and retain compiler verification of our compatibility tests
+     *
      * @param field
      * @param newValue
      */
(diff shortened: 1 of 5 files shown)


sdk/fixtures/src/main/java/com/klaviyo/fixtures/BaseTest.kt
```diff
@@ -97,6 +97,9 @@ abstract class BaseTest {
     /**
      * Gross way to modify a final static field through reflection
      *
+     * NOTE This will break in future Java versions, but is the only way to mock
```
evan-masseau, line 100: Kind of an unrelated comment, it just came up because I couldn't run unit tests until i fixed which JDK the IDE was using to run my tests.

kennyklaviyo (approved):

ajaysubra (approved):

PR #110, opened 2024-01-11
Change Event.Value to a double

evan-masseau:
# Description

Corrects the type of `Event.value` property from string to double. 
Adds value as an optional arg to the `createEvent` convenience method.

# Check List

### DIFF
--- a/.github/workflows/android-feature.yml
+++ b/.github/workflows/android-feature.yml
@@ -10,14 +10,14 @@ jobs:
   # A job for running our tests
   test:
     name: Run Unit Tests
-    
+
     # The type of runner that the job will run on
     runs-on: ubuntu-22.04
 
     steps:
     # Checks-out your repository under $GITHUB_WORKSPACE, so the job can access it
     - uses: actions/checkout@v2
-  
+
     # Sets up a JDK on the job so that we can run Java code
     - name: Set up JDK 11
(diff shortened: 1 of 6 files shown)


.github/workflows/android-feature.yml
```diff
@@ -10,14 +10,14 @@ jobs:
   # A job for running our tests
   test:
     name: Run Unit Tests
-    
+
```
evan-masseau, line 13: IDE changed some whitespace and I committed it mistakenly

README.md
```diff
@@ -141,7 +141,7 @@ Additional event properties can be specified as part of `EventModel`
 
 ```kotlin
 val event = Event(EventMetric.VIEWED_PRODUCT)
-    .setProperty(EventKey.VALUE, "10")
+    .setProperty(EventKey.VALUE, 10)
```
kennyklaviyo, line 144: this is also a breaking change we should write out right?

kennyklaviyo: sorry meant this to be for the setValue func

evan-masseau: Ah yeah, mention it explicitly in the migration guide you mean?

kennyklaviyo: yup!

evan-masseau: Added code sample 👍

kennyklaviyo (approved):

Issue #116, opened 2024-01-17
UninitializedPropertyAccessException on setPushToken


KamilSucharski:
### Description

```markdown
When setting a push token the app might crash. It's a rare crash that I can't recreate, but it does happen for some people.
```

### Checklist

- [ ] I have determined whether this bug is also reproducible in a vanilla Android project
- [ ] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/klaviyo/klaviyo-android-sdk/issues) or [pull request](https://github.com/klaviyo/klaviyo-android-sdk/pulls).

### Expected behavior

```markdown
Push token should be set (and it is most of the time!).
```

### Actual behavior

```markdown
Fatal Exception: kotlin.UninitializedPropertyAccessException: lateinit property initialBody has not been initialized
       at com.klaviyo.analytics.networking.requests.PushTokenApiRequest.equals(PushTokenApiRequest.kt:96)
       at java.util.concurrent.ConcurrentLinkedDeque.contains(ConcurrentLinkedDeque.java:1081)
       at com.klaviyo.analytics.networking.KlaviyoApiClient.enqueueRequest(KlaviyoApiClient.kt:118)
       at com.klaviyo.analytics.networking.KlaviyoApiClient.enqueuePushToken(KlaviyoApiClient.kt:81)
       at com.klaviyo.analytics.Klaviyo.setPushToken(Klaviyo.kt:153)
```
```


### Steps to reproduce

```markdown
Klaviyo.initialize(BuildConfig.KLAVIYO_API_KEY, application)
application.registerActivityLifecycleCallbacks(Klaviyo.lifecycleCallbacks)
FirebaseMessaging.getInstance().token.addOnSuccessListener { pushToken ->
    Klaviyo.setPushToken(pushToken)
}
```

### The Klaviyo Android SDK version information

1.3.5

### Device Information

Samsung Galaxy S21 FE 5G, Android 14 | Samsung Galaxy Note9, Android 10 | Samsung Galaxy S9+, Android 10 | Samsung Galaxy A42 5G, Android 12

### Android Studio Version

2023.1.1

### Android API Level

34

jordan-griffin:
Hey, thanks for reporting. Though we have failed to reproduce on our end either - we're looking into this and have a few ideas on how we might remediate the issue

evan-masseau:
[...]

PR #117, opened 2024-01-17
Quick fix to bug with push token request equality operator

evan-masseau:
# Description

First pass at fixing #116 by moving the setter of that lateinit property. However, after deserializing from the persistent queue, the equality operator won't work the same way since body is already mutated.

# Check List

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
@@ -57,6 +57,13 @@ internal class PushTokenApiRequest(
     override val successCodes: IntRange get() = HTTP_ACCEPTED..HTTP_ACCEPTED
 
     override var body: JSONObject? = null
+        set(value) {
+            if (!this::initialBody.isInitialized) {
+                initialBody = value?.toString() ?: ""
+            }
+
+            field = value
+        }
         get() {
             // Update body to include Device metadata whenever the body is retrieved (typically during sending) so the latest data is included
             field?.getJSONObject(DATA)?.getJSONObject(ATTRIBUTES)?.apply {
@@ -86,9 +93,7 @@ internal class PushTokenApiRequest(
                     VENDOR to VENDOR_FCM
                 )
             )
-        ).also {
-            initialBody = it.toString()
-        }
+        )
     }
 
     override fun equals(other: Any?): Boolean {
--- a/sdk/analytics/src/test/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequestTest.kt
+++ b/sdk/analytics/src/test/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequestTest.kt
@@ -3,6 +3,7 @@ package com.klaviyo.analytics.networking.requests
 import com.klaviyo.analytics.model.Profile
 import org.json.JSONObject
 import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotEquals
 import org.junit.Test
 
 internal class PushTokenApiRequestTest : BaseRequestTest() {
@@ -21,6 +22,18 @@ internal class PushTokenApiRequestTest : BaseRequestTest() {
         .setEmail(EMAIL)
         .setPhoneNumber(PHONE)
         .setExternalId(EXTERNAL_ID)
+
+    @Test
+    fun `Equality operator`() {
+        val aRequest = PushTokenApiRequest(PUSH_TOKEN, stubProfile)
+        val bRequest = PushTokenApiRequest(PUSH_TOKEN, stubProfile)
+        assertEquals(aRequest, bRequest)
+
+        // At this time, decoding from JSON does not preserve equality check for push token requests
+        // But at least it should not crash to compare them
+        val bRequestDecoded = KlaviyoApiRequestDecoder.fromJson(bRequest.toJson())
+        assertNotEquals(aRequest, bRequestDecoded)
+    }
 
     @Test
     fun `Uses correct endpoint`() {
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/ProfileApiRequest.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/ProfileApiRequest.kt
@@ -33,7 +33,7 @@ internal class ProfileApiRequest(
             return arrayOf(
                 DATA to mapOf(
                     TYPE to PROFILE,
-                    ATTRIBUTES to filteredMapOf( // All of the enumerated keys are "attributes"
+                    ATTRIBUTES to filteredMapOf( // All the enumerated keys are "attributes"
                         extract(ProfileKey.EMAIL),
                         extract(ProfileKey.PHONE_NUMBER),
                         extract(ProfileKey.EXTERNAL_ID),


kennyklaviyo (dismissed):

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequestTest.kt
```diff
@@ -22,6 +23,18 @@ internal class PushTokenApiRequestTest : BaseRequestTest() {
...
+    @Test
+    fun `Equality operator`() {
+        val aRequest = PushTokenApiRequest(PUSH_TOKEN, stubProfile)
+        val bRequest = PushTokenApiRequest(PUSH_TOKEN, stubProfile)
+        assertEquals(aRequest, bRequest)
+
+        // At this time, decoding from JSON does not preserve equality check for push token requests
+        // But at least it should not crash to compare them
+        val bRequestDecoded = KlaviyoApiRequestDecoder.fromJson(bRequest.toJson())
+        assertNotEquals(aRequest, bRequestDecoded)
```
ndurell, line 35: What's different between these?

evan-masseau: The cause of the bug was that we use a different constructor when decoding this class from JSON. So the first is just create 2 objects, test their equality. The second is encode and decode it, and re-test equality

evan-masseau: Previously, testing equality failed on a decoded json object because the lateinit var was not initialized. With this patch, equality check should not crash, though expected behavior is that it will not be equal anymore, because the payload body has been mutated

evan-masseau: Exactly, this was my failing-test repro.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
```diff
@@ -57,6 +57,13 @@ internal class PushTokenApiRequest(
     override val successCodes: IntRange get() = HTTP_ACCEPTED..HTTP_ACCEPTED
 
     override var body: JSONObject? = null
+        set(value) {
+            if (!this::initialBody.isInitialized) {
```
ndurell, line 61: Is initialBody part of the parent class?

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
```diff
@@ -57,6 +57,13 @@ internal class PushTokenApiRequest(
...
+                initialBody = value?.toString() ?: ""
```
ndurell, line 62: Could you just initialize `initialBody` to empty string you do that by default or would that be confusing?

evan-masseau: Its a fair point, that would eliminate the risk of using `lateinit` at all.

ndurell: tbh though I don't know much about lateinit and when to use it.

evan-masseau: Its a way to say "this property is non-nullable, but I can't set it in the constructor... so I promise not to access it before its set" and will crash if you break your promise. Has some inherent risk

sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
```diff
@@ -11,7 +11,7 @@ import com.klaviyo.core.Registry
 import com.klaviyo.core.networking.NetworkMonitor
 
 /**
- * Exception that is thrown when the the Klaviyo API token is missing from the config
+ * Exception that is thrown when the Klaviyo API token is missing from the config
```
evan-masseau, line 14: can't help fixing grammar things when i see them

kennyklaviyo: is there a linter that helps with this?

evan-masseau: Good Q, grammar problems are highlighted in a shade of red in my IDE color theme, so it always catches my eye.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
```diff
@@ -86,19 +67,33 @@ internal class PushTokenApiRequest(
...
                 )
             )
-        ).also {
-            initialBody = it.toString()
-        }
+        )
     }
 
+    // Update body to include Device metadata whenever the body is retrieved (typically during sending) so the latest data is included
+    override val requestBody: String?
```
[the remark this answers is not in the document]
ndurell: Nice!

sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
kennyklaviyo: we should find a way to share some IDE configs

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
```diff
@@ -86,19 +67,33 @@ internal class PushTokenApiRequest(
...
+                    if (DeviceProperties.backgroundData) BG_AVAILABLE else BG_UNAVAILABLE
+                )
+                put(METADATA, JSONObject(DeviceProperties.buildMetaData()))
+            }
+        }?.toString()
+
     override fun equals(other: Any?): Boolean {
         return when (other) {
-            is PushTokenApiRequest -> initialBody == other.initialBody
+            is PushTokenApiRequest -> body.toString() == other.body.toString()
```
ndurell, line 91: Sorry if I missed what's the reasoning behind using toString() vs comparing body to other.body?

evan-masseau: I didn't actually try that, let me see. My assumption was separate object instances won't appear equal

kennyklaviyo (approved):

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
ndurell: Ahh might compare references?

kennyklaviyo: one difficult thing is order:
```
JSONObject("{'keyA': 'value', 'keyB': 1}").toString() == JSONObject("{'keyB': 1, 'keyA': 'value'}").toString() // false
```

kennyklaviyo: but yeah, without to string, its a reference check:
```
JSONObject("{'keyA': 'value'}") == JSONObject("{'keyA': 'value'}")
// false
JSONObject("{'keyA': 'value'}").toString() == JSONObject("{'keyA': 'value'}").toString()
// true
```

evan-masseau: Thanks for the check kenny. Yeah I was surprised the order thing didn't cause my test to fail actually. Unless its just coincidentally passing. I had to add a json compare function to the unit tests for that same reason.

ndurell (approved):

PR #119, opened 2024-01-18
Documentation: top-level redirect for github pages

evan-masseau:
Adds a top-level redirect so that the root page of GH pages doesn't 404 and always points to latest docs (we just have to update this html file in the version bump commit)


docs/index.html
```diff
@@ -0,0 +1,2 @@
+<!-- Redirect to latest version -->
+<meta HTTP-EQUIV="REFRESH" content="0; url=./1.4.1/index.html">
```
evan-masseau, line 2: So we'd just update this version instead of the link in the readme in the future.

kennyklaviyo: should we call this out in the comment where we update version to do it here instead of the readme?

evan-masseau: yes

jordan-griffin: Probably want to make sure the process doc on Android releases is up to date as well.

jordan-griffin (approved):

kennyklaviyo (approved):

docs/index.html
evan-masseau: I Will keep wiki synced

PR #127, opened 2024-02-05
Add note about anonymous profiles to readme

evan-masseau:
Akin to a note in the iOS sdk readme


jordan-griffin (approved):

mark-piana (approved):

README.md
```diff
@@ -158,6 +158,11 @@ Klaviyo.setEmail("robin@example.com")
     .setProfileAttribute(ProfileKey.FIRST_NAME, "Robin")
 ```
 
+### Anonymous Tracking
+By default, Klaviyo will begin tracking unidentified users in your app when a push token is set, or if it receives an
+event to track. That way, you can track events and collect push tokens from users prior to collecting profile identifiers.
+When an email or other identifier is provided, Klaviyo will merge the data from the anonymous user to an identified user.
```
evan-masseau, line 164:
```suggestion
Klaviyo will track unidentified users with an autogenerated ID whenever a push token is set or an event is created.
That way, you can collect push tokens and track events prior to collecting profile identifiers such as email or
phone number. When an identifier is provided, Klaviyo will merge the anonymous user with an identified user.
```

PR #129, opened 2024-02-08
Wrap public methods in try/catch

evan-masseau:
# Description
Adds a try/catch around our public API so that errors are not fatal to the containing application, only log an error. 

# Check List

- [x] Are you changing anything with the public API?  Sort of, our public API will no longer throw `KlaviyoException`s
- [x] Have you tested this change on real device? n/a

## Changelog / Code Overview
- Added private inline functions that wrap a block in try/catch
- Wrapped all the public methods in `Klaviyo` in those try/catch methods
- Added tests for each public API method when called without initializing

## Test Plan
- Unit tests, test app.

## Related Issues/Tickets
CHNL-4688
CHNL-5324
CHNL-5490


sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -311,4 +317,21 @@ object Klaviyo {
...
+    /**
+     * Safely invoke a function and log KlaviyoExceptions rather than crash
+     */
+    private inline fun <T>safeCall(block: () -> T): T? {
+        return try {
+            block()
+        } catch (e: KlaviyoException) {
+            // KlaviyoException is self-logging
+            null
+        }
```
evan-masseau, line 330: We could go even farther and swallow all exceptions by adding 
```
        catch (e: Throwable) {
            Registry.log.error("Klaviyo SDK encountered an unexpected exception", e)
            null
        }
```

ajaysubra:
This is a slight change between iOS and Android where in iOS we crash in debug builds. We'd discussed if crashing is required but does seem like a effective way to get folks to setup correctly. Does the log bubble up and avoid being missed?

evan-masseau:
> This is a slight change between iOS and Android where in iOS we crash in debug builds. We'd discussed if crashing is required but does seem like a effective way to get folks to setup correctly. Does the log bubble up and avoid being missed?

I thought we'd come to the conclusion that crashing on debug and not release was confusing, but I'm open to continuing that discussion. I don't know of any other options for android other than crashing or logging, like the XCode thing you showed me.

ajaysubra:
Also, QQ - does this also show up on RN if they don't open the IDE but build from terminal. I'm guessing yes but just confirming.

ajaysubra (approved):

evan-masseau:
> Also, QQ - does this also show up on RN if they don't open the IDE but build from terminal. I'm guessing yes but just confirming.

Very glad we checked this. No it does not, and I don't think it even logs in a regular android application, because we turn logging off in the "release" flavor of the Android SDK. Gonna have to address this in the Android SDK.

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
ajaysubra: >            // KlaviyoException is self-logging

I guess this statement is self explanatory but since I don't really understand android curious how the exception here is logged to console?

ajaysubra (approved):

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
evan-masseau: In the `KlaviyoException` class, I call something like `Registry.log.error(this)` from its initializer.

sdk/analytics/src/main/java/com/klaviyo/analytics/OperationQueue.kt
```diff
@@ -0,0 +1,70 @@
+package com.klaviyo.analytics
+
+import com.klaviyo.analytics.model.Profile
+import com.klaviyo.analytics.networking.ApiClient
+import com.klaviyo.core.KlaviyoException
+import com.klaviyo.core.Registry
+import com.klaviyo.core.config.Clock
+
+internal class OperationQueue {
```
evan-masseau, line 9: Rather than keep adding private methods to the Klaviyo object, figured I'd extract an operations manager class like this

evan-masseau: This is where we would add an in-memory operations queue for methods invoked prior to initialize, if we do that.

sdk/analytics/src/main/java/com/klaviyo/analytics/OperationQueue.kt
```diff
@@ -0,0 +1,70 @@
...
+}
+
+/**
+ * Safely invoke a function and log KlaviyoExceptions rather than crash
+ */
+internal inline fun <T>safeCall(block: () -> T): T? {
+    return try {
+        block()
+    } catch (e: KlaviyoException) {
+        // KlaviyoException is self-logging
```
ndurell, line 62: Logged at the default level?

evan-masseau: I think they're all `.error()`

sdk/analytics/src/main/java/com/klaviyo/analytics/OperationQueue.kt
```diff
@@ -0,0 +1,70 @@
...
+        Registry.log.info("Flushing profile update")
+        Registry.get<ApiClient>().enqueueProfile(it)
+        pendingProfile = null
+    }
+}
+
+/**
+ * Safely invoke a function and log KlaviyoExceptions rather than crash
+ */
+internal inline fun <T>safeCall(block: () -> T): T? {
```
ndurell, line 58: Should this code be in here? Feels maybe unrelated...

sdk/analytics/src/main/java/com/klaviyo/analytics/OperationQueue.kt
ndurell: Sorry to quibble with the name wonder if ProfileOperationQueue is more appropriate here?

sdk/analytics/src/main/java/com/klaviyo/analytics/OperationQueue.kt
evan-masseau: In this file, or in this PR?

sdk/analytics/src/main/java/com/klaviyo/analytics/OperationQueue.kt
ndurell: Or could this be made into a more general purpose thing?

sdk/analytics/src/main/java/com/klaviyo/analytics/OperationQueue.kt
kennyklaviyo: in this file, since its being used only in Klaviyo.kt and not OperationQueue

sdk/analytics/src/main/java/com/klaviyo/analytics/OperationQueue.kt
```diff
@@ -0,0 +1,70 @@
...
+    fun debounceProfileUpdate(profile: Profile) {
+        // Log for traceability
+        val operation = pendingProfile?.let { "Merging" } ?: "Starting"
+        Registry.log.info("$operation profile update")
+
+        // Merge new changes into pending transaction
+        pendingProfile = pendingProfile?.merge(profile) ?: profile
+
+        // Add current identifiers from UserInfo to pending transaction
+        pendingProfile = UserInfo.getAsProfile().merge(pendingProfile!!)
```
ndurell, line 35: Is `!!` safe here?

sdk/analytics/src/main/java/com/klaviyo/analytics/OperationQueue.kt
ndurell: Yah meant file.

sdk/analytics/src/main/java/com/klaviyo/analytics/OperationQueue.kt
evan-masseau: Yeah I see what you mean, right now its just the profile debouncing, and the try/catch inline methods. This name was more generic and forward-looking if we did the in-memory queue of method invoked prior to initializing I'd put it in here too

sdk/analytics/src/main/java/com/klaviyo/analytics/OperationQueue.kt
evan-masseau: Yeah fair! Again was kind of forward-looking for the pre-initialize in-memory queue thing. But as is, it doesn't seem to belong here.

sdk/analytics/src/main/java/com/klaviyo/analytics/OperationQueue.kt
evan-masseau: Yes because it gets set above, but it could probably be written to avoid using `!!` at all

sdk/analytics/src/main/java/com/klaviyo/analytics/ProfileOperationQueue.kt
```diff
@@ -0,0 +1,52 @@
+package com.klaviyo.analytics
+
+import com.klaviyo.analytics.model.Profile
+import com.klaviyo.analytics.networking.ApiClient
+import com.klaviyo.core.Registry
+import com.klaviyo.core.config.Clock
+
+internal class ProfileOperationQueue {
```
evan-masseau, line 8: @ndurell and @kennyklaviyo - renamed the class and file to its actual purpose

sdk/analytics/src/main/java/com/klaviyo/analytics/ProfileOperationQueue.kt
```diff
@@ -0,0 +1,52 @@
...
+     */
+    fun debounceProfileUpdate(profile: Profile) {
+        // Log for traceability
+        val operation = pendingProfile?.let { "Merging" } ?: "Starting"
+        Registry.log.info("$operation profile update")
+
+        // Merge new changes into pending transaction and
+        // add current identifiers from UserInfo to pending transaction
+        pendingProfile = UserInfo.getAsProfile().merge(
+            pendingProfile?.merge(profile) ?: profile
```
evan-masseau, line 33: @ndurell thanks for the idea, this is cleaner bc it doesn't require `!!`

sdk/core/src/main/java/com/klaviyo/core/KlaviyoException.kt
```diff
@@ -12,3 +12,20 @@ abstract class KlaviyoException(final override val message: String) : Exception(
 
     private fun log() = Registry.log.error(message, this)
 }
+
+/**
+ * Safely invoke a function and log KlaviyoExceptions rather than crash
+ */
+inline fun <T> safeCall(block: () -> T): T? {
```
evan-masseau, line 19: @ndurell and @kennyklaviyo I think this is a more appropriate place for these wrapper functions, lmk what you think

kennyklaviyo (approved):

PR #130, opened 2024-02-13
Logging updates

evan-masseau:
# Description
We intentionally disabled all log output from the SDK in our earlier releases. In order to provide some useful logging, I'm enabling it and merging log implementations. Rather than silence all output, I'm putting in a configurable log level that defaults to less verbose output. `Warning+` for debug and `Error+` for release.

# Check List

- [x] Are you changing anything with the public API? `NO`
- [x] Are your changes backwards compatible with previous SDK Versions? `YES`
- [x] Have you tested this change on real device? `YES`
- [x] Have you added unit test coverage for your changes? `YES`
- [x] Have you verified that your changes are compatible with all Android versions the SDK currently supports? `YES`

## Changelog / Code Overview

- Enable logging in published SDK
- Added warning level
- Added level-aware logging with a configurable threshold
- Threshold has a default based on debug vs release, and can be configured from code and/or manifest file
- Adjusted log levels throughout the code
- Updated tests, removed some old log related-test code from earlier build variants/flavors

## Test Plan

- Unit tests
- Will test more extensively from android and react native test apps as I work on the wrapped API methods

## Related Issues/Tickets

CHNL-5491


ajaysubra (approved):
NICE!

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -152,13 +152,13 @@ internal object KlaviyoApiClient : ApiClient {
...
                 Registry.log.wtf("Invalid persistent queue JSON", exception)
-                Registry.log.debug(it)
+                Registry.log.info(it)
                 emptyArray<String>()
             }
         }?.forEach { uuid ->
             Registry.dataStore.fetch(uuid).let { json ->
                 if (json == null) {
-                    Registry.log.error("Missing request JSON for $uuid")
+                    Registry.log.debug("Missing request JSON for $uuid")
```
ndurell, line 161: hmm why go from error to debug?

evan-masseau: Previously, all these levels were only here for us to see, because in a published build there was no logging. 
I didn't think this was worth logging in the client's application, so I lowered it so it won't be visible by the default logging level in a release build.

sdk/core/src/main/java/com/klaviyo/core/KLog.kt
```diff
@@ -0,0 +1,116 @@
...
+         * 1 = Assert only
+         * 2 = Error and above
+         * 3 = Warning and above
+         * 4 = Info and above
+         * 5 = Debug and above
+         * 6 = Verbose and above
+         */
+        private const val LOG_LEVEL = "com.klaviyo.core.log_level"
+
+        private val defaultLogLevel = if (BuildConfig.DEBUG) {
```
ndurell, line 64: Can you override this with the manifest setting?

evan-masseau: Yep, this is the default based on build config, then the host app can set it from the manifest, or by importing `core` and changing `logLevel` at runtime (which I probably won't advertise, but wanted for the test app)

sdk/core/src/main/java/com/klaviyo/core/KLog.kt
```diff
@@ -0,0 +1,116 @@
...
+                            "KlaviyoLog",
+                            "Invalid log level $it detected in manifest, defaulting to $default."
+                        )
+                    }
+                }
+            }
+
+            return _logLevel ?: defaultLogLevel
+        }
+        set(value) {
```
evan-masseau, line 27: Since this is a public setter, you can also change log level from runtime if you import `core` from gradle

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -152,13 +152,13 @@ internal object KlaviyoApiClient : ApiClient {
             } catch (exception: JSONException) {
                 wasMutated = true
                 Registry.log.wtf("Invalid persistent queue JSON", exception)
-                Registry.log.debug(it)
+                Registry.log.info(it)
```
ndurell, line 155: Just noting this jumped from debug to info whereas other jumps were maybe not as severe _I think_.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -169,8 +169,8 @@ internal object KlaviyoApiClient : ApiClient {
                     } catch (exception: JSONException) {
                         wasMutated = true
                         Registry.log.wtf("Invalid request JSON $uuid", exception)
+                        Registry.log.info(json)
```
ndurell, line 172: Just double checking here...you changed the order of operations but this seems consistent with the block above so I think it's fine.

sdk/analytics/src/test/java/com/klaviyo/analytics/DevicePropertiesTest.kt
```diff
@@ -13,6 +13,8 @@ import org.junit.Test
 internal class DevicePropertiesTest : BaseTest() {
 
     private val mockVersionCode = 123
+
+    @Suppress("DEPRECATION")
```
ndurell, line 17: PackageInfo is deprecated?

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
evan-masseau: You're correct. Almost all cases I reduced log level. But there were two cases where we had a `wtf` error followed by some additional explanatory data. I thought in those cases `info` made more sense.

sdk/core/src/main/java/com/klaviyo/core/KLog.kt
```diff
@@ -0,0 +1,116 @@
...
+
+open class KLog : Log {
+
+    private var _logLevel: Level? = null
+    override var logLevel: Level
+        get() {
+            if (_logLevel == null && Registry.isRegistered<Config>()) {
+                _logLevel = Registry.config.getManifestInt(LOG_LEVEL, defaultLogLevel.ordinal).let {
+                    Level.entries.getOrNull(it) ?: defaultLogLevel.also { default ->
+                        default.log(
```
ndurell, line 17: That's a mindbender logging that you can't get the log level! 😂

sdk/core/src/main/java/com/klaviyo/core/KLog.kt
```diff
@@ -0,0 +1,116 @@
...
+                            "KlaviyoLog",
```
ndurell, line 18: Worth a constant _maybe_?

sdk/core/src/main/java/com/klaviyo/core/LogLevel.kt
```diff
@@ -0,0 +1,17 @@
...
+ * Bridge between the Klaviyo Log levels and Android static log functions
+ */
+fun Level.log(tag: String, msg: String, ex: Throwable? = null) = when (this) {
+    Level.Verbose -> Log.v(tag, msg, ex)
+    Level.Debug -> Log.d(tag, msg, ex)
+    Level.Info -> Log.i(tag, msg, ex)
+    Level.Warning -> Log.w(tag, msg, ex)
+    Level.Error -> Log.e(tag, msg, ex)
+    Level.Assert -> Log.wtf(tag, msg, ex)
+    Level.None -> 0
```
ndurell, line 16: Does log return a number or just have to return something?

evan-masseau: Yeah all the android util methods return Int so I just kept it consistent. I could also convert it to return void/Unit i suppose.

sdk/core/src/main/java/com/klaviyo/core/KLog.kt
```diff
@@ -0,0 +1,116 @@
...
+import com.klaviyo.core.config.Log
+import com.klaviyo.core.config.Log.Level
+import java.util.regex.Pattern
+
+open class KLog : Log {
+
+    private var _logLevel: Level? = null
+    override var logLevel: Level
+        get() {
+            if (_logLevel == null && Registry.isRegistered<Config>()) {
```
ndurell, line 14: Why check isRegistered here?

evan-masseau: This is checking if `applicationContext` will be available through `Config`. Without it, we can't read from the manifest.

evan-masseau: Otherwise, accessing `Registry.Config` would throw an exception, then we'd try to log the exception... ♾️

sdk/core/src/main/java/com/klaviyo/core/config/Log.kt
```diff
@@ -2,6 +2,34 @@ package com.klaviyo.core.config
...
+        Error,
+        Assert
+    }
+
+    /**
+     * The minimum log level to output
+     *
+     * To configure from manifest, specify the minimum log level to output:
+     *
+     * 0 = Disable logging
```
ndurell, line 23: Hmm...is this reversed from the setting we documented?

evan-masseau: Oh crud, yeah good catch, now to remember which one is correct 🤣

ndurell: Maybe not reversed is assert in the wrong spot though or doesn't matter?

evan-masseau: I think the readme is currently incorrect. I think I went with 0 = disable, 1 = verbose+, 2 = debug+ etc.

ndurell: https://developer.android.com/reference/android/util/Log it should match stuff in here right?

evan-masseau: Yes, in spirit anyway. For some reason theirs go from 2 to 7. But lower number = lower severity was my intent. I fixed values in the readme

sdk/core/src/test/java/com/klaviyo/core/RegistryTest.kt
```diff
@@ -10,6 +14,17 @@ class RegistryTest {
     private interface TesWrongDependency
```
ndurell, line 14: Just noticed this very minor should it `Test` instead of `Tes`?

evan-masseau: Ya, I think I caught it on one of the other PRs

sdk/core/src/main/java/com/klaviyo/core/KLog.kt
```diff
@@ -0,0 +1,116 @@
...
+    private companion object {
+        /**
+         * Log level key for manifest, specifying the minimum log level to output:
+         * 0 = disable logging entirely
+         * 1 = Assert only
+         * 2 = Error and above
+         * 3 = Warning and above
+         * 4 = Info and above
+         * 5 = Debug and above
+         * 6 = Verbose and above
```
kennyklaviyo, line 60: these seem inconsistent from the readme

evan-masseau: dangit how many places did I write this...

evan-masseau: @kennyklaviyo fixed , thx

kennyklaviyo (approved):

PR #131, opened 2024-02-15
Attach lifecycle listeners internally in initialize method

evan-masseau:
# Description
While working on a ticket to automatically attach android lifecycle listeners for react native SDK, I've realized we could extend that convenience to the android SDK. 

# Check List

- [x] Are you changing anything with the public API? YES - marked `lifecycleCallbacks` deprecated
- [x] Have you tested this change on real device? YES
- [x] Have you added unit test coverage for your changes? Yes
- [x] Have you verified that your changes are compatible with all Android versions the SDK currently supports? - YES

## Changelog / Code Overview
- Added lifecycle listeners in the `initialize` method instead of requiring app developers to do it. 
- Mark the public property to be deprecated and replace with a no-op implementation to avoid duplicate listeners.
- Update readme instructions, migration guide.

## Test Plan
- Updated unit test
- Verify lifecycle listeners working in android test app
- Pull this code into the React Native SDK to make sure it works there too

## Related Issues/Tickets

CHNL-4635


sdk/core/src/main/java/com/klaviyo/core/lifecycle/NoOpLifecycleCallbacks.kt
```diff
@@ -0,0 +1,21 @@
...
+import android.app.Application
+import android.os.Bundle
+
+/**
+ * A no-op implementation of ActivityLifecycleCallbacks
+ * to temporarily replace the public property Klaviyo.lifecycleCallbacks
+ * and prevent duplicate registration of the KlaviyoLifecycleMonitor
+ * until next major release when we can make the breaking change to remove that public property
+ */
+object NoOpLifecycleCallbacks : Application.ActivityLifecycleCallbacks {
```
jordan-griffin, line 13: So this would also be removed in next major release once we no longer need to worry about duplicate lifecycle registration?

evan-masseau: Correct, I just can't remove it in a minor release by semantic versioning rules

MIGRATION_GUIDE.md
```diff
@@ -2,6 +2,17 @@
...
 
+## 2.1.0 Deprecations
+#### Deprecated `Klaviyo.lifecycleCallbacks`
+In an effort to reduce setup code required to integrate the Klaviyo Android SDK, we have deprecated the public property 
+`Klaviyo.lifecycleCallbacks` and will now register for lifecycle callbacks automatically upon `initialize`.
+It is no longer required to have this line in your `Application.onCreate()` method:
+```kotlin
+registerActivityLifecycleCallbacks(Klaviyo.lifecycleCallbacks)
+```
+For version 2.1.x, `Klaviyo.lifecycleCallbacks` has been replaced with a no-op implementation to avoid duplicative
```
ndurell, line 13: 💯

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -47,6 +52,12 @@ object Klaviyo {
                 .applicationContext(applicationContext)
                 .build()
         )
+
+        val application = applicationContext.applicationContext as? Application
+        application?.apply {
+            unregisterActivityLifecycleCallbacks(Registry.lifecycleCallbacks)
+            registerActivityLifecycleCallbacks(Registry.lifecycleCallbacks)
+        } ?: throw LifecycleException()
```
ndurell, line 60: Would this crash their app? Should it log a warning instead?

evan-masseau: Yep! It will log a warning when #129 and #130 are merged

ndurell (approved):

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
ndurell: oh nice!

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -47,6 +52,12 @@ object Klaviyo {
                 .applicationContext(applicationContext)
                 .build()
         )
+
+        val application = applicationContext.applicationContext as? Application
```
evan-masseau, line 56: Here is the [documentation](https://developer.android.com/reference/android/content/Context.html#getApplicationContext()) on `applicationContext`

In earlier speccing last year, I shied away from using this method to access the `Application` object because of a stackoverflow post from 2011 that warned off using it. Now that I've read more about it, I think that post was incorrect. For example, this description from official docs describes exactly what we're doing here: 

> Return the context of the single, global Application object of the current process. This generally should only be used if you need a Context whose lifecycle is separate from the current context, that is tied to the lifetime of the process rather than the current component.

kennyklaviyo (approved):

PR #139, opened 2024-02-27
Doc-bot: commit straight to documentation branch

evan-masseau:
Commit documentation updates directly to branch instead of generating a pull request!


.github/workflows/doc-bot.yml
```diff
@@ -12,7 +10,7 @@ jobs:
     runs-on: ubuntu-22.04
     env:
       CI_COMMIT_MESSAGE: Generated docs for ${{ github.sha }}
-      GH_TOKEN: ${{ github.token }}
+      GH_TOKEN: ${{ secrets.GH_ACTION_ACCESS_TOKEN }}
```
kennyklaviyo, line 13: do we need this if we have set the token below?

evan-masseau: good point, probably not!

evan-masseau: I re-read the docs, don't think we need the one below and just need to set the env variable

kennyklaviyo (approved):

PR #148, opened 2024-03-22
Update network retry behavior to extend retries, add jitter

evan-masseau:
# Description
This should match the Swift SDK retry behavior as I understood it. I tried to make the test very readable, would appreciate @ndurell or @ajaysubra confirming my logic. 

First of two PRs regarding network rate limits.

# Check List

  - No
  - Yes
  - Emulator is sufficient
  - Yes
  - N/A

## Changelog / Code Overview

- Retry 50 times before failure
- Updated my exponential backoff calculation
- Added jitter
- Re-worked the unit test based on better understanding of the thread Handler behavior

## Test Plan

- Unit test and debugged on a device after forcing artificial 429 responses

## Related Issues/Tickets

CHNL-6779

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
@@ -14,6 +14,9 @@ import com.klaviyo.analytics.networking.requests.PushTokenApiRequest
 import com.klaviyo.core.Registry
 import com.klaviyo.core.lifecycle.ActivityEvent
 import java.util.concurrent.ConcurrentLinkedDeque
+import kotlin.math.max
+import kotlin.math.min
+import kotlin.math.pow
 import org.json.JSONArray
 import org.json.JSONException
 import org.json.JSONObject
@@ -38,9 +41,18 @@ internal object KlaviyoApiClient : ApiClient {
             when (r.state) {
                 Status.Unsent.name -> Registry.log.verbose("${r.type} Request enqueued")
                 Status.Inflight.name -> Registry.log.verbose("${r.type} Request inflight")
-                Status.PendingRetry.name -> Registry.log.debug("${r.type} Request retrying")
-                Status.Complete.name -> Registry.log.verbose("${r.type} Request completed")
-                else -> Registry.log.debug("${r.type} Request failed")
+                Status.PendingRetry.name -> {
+                    val attemptsRemaining = Registry.config.networkMaxAttempts - r.attempts
+                    Registry.log.warning(
+                        "${r.type} Request failed with code ${r.responseCode}, and will be retried up to $attemptsRemaining more times."
+                    )
+                }
+                Status.Complete.name -> Registry.log.verbose(
+                    "${r.type} Request succeeded with code ${r.responseCode}"
+                )
+                else -> Registry.log.error(
+                    "${r.type} Request failed with code ${r.responseCode}, and will be dropped"
+                )
             }
 
             r.responseBody?.let { response ->
@@ -143,9 +155,9 @@ internal object KlaviyoApiClient : ApiClient {
         // Keep track if there's any errors restoring from persistent store
         var wasMutated = false
 
-        Registry.log.verbose("Restoring persisted queue")
-
         Registry.dataStore.fetch(QUEUE_KEY)?.let {
+            Registry.log.verbose("Restoring persisted queue")
+
             try {
                 val queue = JSONArray(it)
                 Array(queue.length()) { i -> queue.optString(i) }
@@ -227,16 +239,18 @@ internal object KlaviyoApiClient : ApiClient {
      */
     private fun startBatch(force: Boolean = false) {
         stopBatch() // we only ever want one batch job running
-        handler?.post(NetworkRunnable(force))
-        Registry.log.verbose("Started background handler")
+        handler?.post(NetworkRunnable(force)).also {
+            Registry.log.verbose("Posted job to network handler message queue")
+        }
     }
 
     /**
      * Stop all jobs on our handler thread
      */
     private fun stopBatch() {
-        handler?.removeCallbacksAndMessages(null)
-        Registry.log.verbose("Stopped background handler")
+        handler?.removeCallbacksAndMessages(null).also {
+            Registry.log.verbose("Cleared jobs from network handler message queue")
+        }
     }
 
     /**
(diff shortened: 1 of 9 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -318,10 +332,22 @@ internal object KlaviyoApiClient : ApiClient {
          * Re-queue the job to run again after [flushInterval] milliseconds
          */
         private fun requeue() {
-            Registry.log.verbose("Retrying network batch in $flushInterval")
+            Registry.log.verbose("Network batch will run in $flushInterval ms")
             force = false
+            enqueuedTime = Registry.clock.currentTimeMillis()
```
jordan-griffin, line 337: Whats the thought behind now updating the enqueued time whenever an item is re-queued, instead of having it represent when the request was initially made?

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -318,10 +332,22 @@ internal object KlaviyoApiClient : ApiClient {
...
             handler?.postDelayed(this, flushInterval)
         }
+
+        private fun computeRetryInterval(attempts: Int): Long {
```
jordan-griffin, line 341: Logic here looks totally sound. Just wondering if this is a mirror of what we do on iOS? Like are the limits we apply to our jitter the same, etc.

jordan-griffin: Default min/max retry intervals as well

jordan-griffin (approved):

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
evan-masseau: Yeah good catch. Its kind of a technicality, but this change makes the tests better match actual behavior on the device.

Previously, I was using `flushInterval *= request.attempts + 1`, the idea being that the flush interval is incrementing relative to the initial queue time like so:
- First attempt is 10s after queueing
- Second attempt is 20s after queueing, but 10s after the previous attempt
- Third attempt is 60s after queueing, but 30s after the prior attempt
- etc

And thats what my test was simulating. But that didn't match the true behavior on the device, because below I'm doing `postDelayed(flushInterval)`, which will only post the job after that _total_ wait time. So in reality the behavior was
- First attempt is 10s after queueing
- Second attempt is 30s after queueing, 20s after the previous attempt
- Third attempt is 90s after queueing, 60s after the prior attempt
- etc

evan-masseau: And actually, re-visiting this is making me realize I think I could leave the initial queued time immutable as before, but use `+= computeRetryInterval(request.attempts)` instead

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
evan-masseau: Yes I believe so. Would like a confirmation of my reading from Noah or Ajay. The difference is mostly just that I compute the retry time all in one function, whereas on iOS its split up into a couple different parts of the network queueing code.

jordan-griffin: @ndurell @ajaysubra when one of you has a chance ^

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
jordan-griffin: Either way I dont think I really care if `enqueuedTime` is immutable or not. I just thought it seemed a little misleading but we dont have any use for an original queue time being stored on a request right now and I dont think there's any particular reason we should care.

Maybe there could be something like dropping certain really old requests given the original queue time (user returns to app after months of not using it triggering some redundant data to send over)

evan-masseau: Yeah I think I see your point. In this case, `enqueuedTime` just refers to the batch job running on the background thread. It could be valuable to track enqueued time on the ApiRequest object (and serialize that to the persistent store JSON) though. 
Using `+=` and leaving `enqueuedTime` immutable doesn't actually work. I think I'd opt to keep as is until perhaps we add parallel network request and/or update to use co-routines or other async method.

sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
```diff
@@ -79,10 +79,16 @@ object KlaviyoConfig : Config {
...
      *
-     * Reasoning: Most likely the rate limit should be cleared within 2 retries with exp backoff.
-     * However, I wanted some extra padding for edge cases, since the consequence is lost data.
+     * Reasoning: Most likely the rate limit should be cleared within 2-3 retries with exp backoff.
      */
-    private const val NETWORK_MAX_RETRIES_DEFAULT: Int = 4
+    private const val NETWORK_MAX_RETRIES_DEFAULT: Int = 50
+
+    /**
+     * Maximum interval between retries for the exponential backoff
```
kennyklaviyo, line 87: nit: add seconds in the docstring

kennyklaviyo: wait, or is this in milliseconds

evan-masseau: A good reason to state it in the doc string

evan-masseau: Its 3 minutes, in milliseconds

evan-masseau:
```suggestion
     * Maximum interval between retries for the exponential backoff, in milliseconds
```

ajaysubra (approved):

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -318,10 +332,22 @@ internal object KlaviyoApiClient : ApiClient {
...
+            val minRetryInterval = flushInterval
+            val jitterSeconds = Registry.config.networkJitterRange.random()
+            val exponentialBackoff = (2.0.pow(attempts).toLong() + jitterSeconds).times(1_000)
```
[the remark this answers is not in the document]
evan-masseau: Yep!

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
evan-masseau: Thanks for the verification!

kennyklaviyo (approved):

PR #149, opened 2024-03-23
Add X-Klaviyo- request headers and use Retry-After response header

evan-masseau:
# Description
Add new request headers, and make use of `Retry-After` response header on rate limited requests(429s)

# Check List

## Changelog / Code Overview
- New request header to every API call that identifies SDK traffic by `X-Klaviyo-Mobile=1`
- New request header to specify the attempt number by `X-Klaviyo-Retry-Attempt=N/MAX`
- Check for `Retry-After` header on rate limited requests (previously unavailable for `client/*` 429s) and obey the server's retry timing, plus jitter.
- [x] Awaiting discussion: should we be persisting the attempt count to disk so that count does not reset on app re-launch.
  - Not adding persistence of attempt count at this time. 

Sample of new request headers from test app:
Screenshot 2024-04-01 at 14 06 34

## Test Plan
- Added unit tests for the new headers
- Consolidated and improved API Request tests
- Added an API rate limit override for a test company to 1/min on the client/push_tokens endpoint and verified behavior in test app. 

## Related Issues/Tickets

CHNL-6842
CHNL-6835


sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
```diff
@@ -87,6 +90,17 @@ internal class KlaviyoApiClientTest : BaseRequestTest() {
         }
     }
 
+    @After
```
evan-masseau, line 93: Small thing, but I grouped setup and cleanup functions together for readability.

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequestTest.kt
```diff
@@ -44,17 +49,25 @@ internal class KlaviyoApiRequestTest : BaseRequestTest() {
     @Before
     override fun setup() {
         super.setup()
-
         every { networkMonitorMock.isNetworkConnected() } returns true
         every { configMock.networkTimeout } returns 1
         every { configMock.networkFlushIntervals } returns longArrayOf(10_000L, 30_000L, 60_000L)
     }
 
+    @After
```
evan-masseau, line 57: for clarity, I annotated all setup/cleanup, even though overrides technically inherit it

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/requests/BaseApiRequestTest.kt
```diff
@@ -0,0 +1,67 @@
+package com.klaviyo.analytics.networking.requests
+
+import com.klaviyo.analytics.DevicePropertiesTest
+import com.klaviyo.analytics.model.Profile
+import com.klaviyo.fixtures.BaseTest
+import org.junit.Assert
+import org.junit.Before
+import org.junit.Test
+
+internal abstract class BaseApiRequestTest<T> : BaseTest() where T : KlaviyoApiRequest {
```
evan-masseau, line 10: Abstract base class for API request tests. The tests in this class get applied to all API Request tests that inherit from here.

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/requests/BaseApiRequestTest.kt
```diff
@@ -0,0 +1,67 @@
...
+    fun `Uses expected HTTP headers`() {
+        Assert.assertEquals(expectedHeaders, makeTestRequest().headers)
+    }
+
+    @Test
+    fun `Uses expected URL query`() {
+        Assert.assertEquals(expectedQuery, makeTestRequest().query)
+    }
+
+    inline fun <reified T> testJsonInterop(request: T) where T : KlaviyoApiRequest {
```
evan-masseau, line 61: Couldn't quite get this to work as a @Test with generics alone, so made a reusable function and call it in one-liner tests in the subclasses.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/ApiRequest.kt
```diff
@@ -76,6 +76,11 @@ interface ApiRequest {
      */
     val responseCode: Int?
 
+    /**
+     * HTTP Response Headers
+     */
+    val responseHeaders: Map<String, List<String>>?
```
kennyklaviyo, line 82: nullable vs empty map?

kennyklaviyo: id lean towards empty map

evan-masseau: I was copying the pattern set by `responseCode`, of being null till request has been sent. But empty map would be fine for this, and less null checks

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -220,7 +243,7 @@ internal open class KlaviyoApiRequest(
         .accumulate(METHOD_JSON_KEY, method.name)
         .accumulate(TIME_JSON_KEY, queuedTime)
         .accumulate(UUID_JSON_KEY, uuid)
-        .accumulate(HEADERS_JSON_KEY, JSONObject(headers))
+        .accumulate(HEADERS_JSON_KEY, JSONObject(headers as Map<String, String>))
```
kennyklaviyo, line 246:
```suggestion
        .accumulate(HEADERS_JSON_KEY, JSONObject(headers.toMap()))
```

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -1,13 +1,17 @@
...
 
+import com.klaviyo.analytics.DeviceProperties
 import com.klaviyo.core.Registry
 import java.io.BufferedReader
 import java.io.IOException
 import java.io.InputStreamReader
 import java.net.HttpURLConnection
 import java.net.URL
-import java.util.UUID
+import java.util.*
```
evan-masseau, line 10:
```suggestion
import java.util.UUID
```
Autoformatting rule I thought i'd disabled in my IDE...

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -171,7 +185,10 @@ internal open class KlaviyoApiRequest(
      * Tracks number of attempts to limit retries
      */
     final override var attempts = 0
```
kennyklaviyo, line 187: should this start at 1?

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -150,7 +157,14 @@ internal open class KlaviyoApiRequest(
...
      * HTTP request headers
      */
-    override var headers: Map<String, String> = emptyMap()
+    override var headers: MutableMap<String, String> = mutableMapOf(
+        HEADER_CONTENT to TYPE_JSON,
+        HEADER_ACCEPT to TYPE_JSON,
+        HEADER_REVISION to V3_REVISION,
+        HEADER_USER_AGENT to DeviceProperties.userAgent,
+        HEADER_KLAVIYO_MOBILE to "1",
+        HEADER_KLAVIYO_ATTEMPT to "1/${Registry.config.networkMaxAttempts}"
```
kennyklaviyo, line 166: should this use attempts instead of hardcoded 1?

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
evan-masseau: Pretty trivial, but won't `toMap` make a copy of the mutable map, while `as` will just treat it as immutable without copying? Which in this case is sufficient bc we're just writing it into a json object?

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -150,7 +157,14 @@ internal open class KlaviyoApiRequest(
     /**
      * HTTP request headers
      */
-    override var headers: Map<String, String> = emptyMap()
+    override var headers: MutableMap<String, String> = mutableMapOf(
```
kennyklaviyo, line 160: should we enforce one or the other, either the field is mutable or the value itself it mutable? otherwise it might open itself to multiple forms of mutation which could get confusing, one place sets the header field vs another place that changes the header field?

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
evan-masseau: Yeah lemme see about that. I also considered not even putting it in here as a default, since it is always added/updated at send time.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
evan-masseau: great point

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
kennyklaviyo: ah yes, good point, i just like avoiding type casts in kotlin, but this makes sense to keep

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
evan-masseau: Will check. Its certainly covered by other tests, but it'd be good to have a standalone test for that.

kennyklaviyo: also, when we persist requests to disk, have we also accounted for storing the attempts? or are we saying, if the app restarts, we'll reset the attempts to 0?

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -315,6 +338,7 @@ internal open class KlaviyoApiRequest(
     protected open fun parseResponse(connection: HttpURLConnection): Status {
         // https://developers.klaviyo.com/en/docs/rate_limits_and_error_handling
         responseCode = connection.responseCode
+        responseHeaders = connection.headerFields
```
kennyklaviyo, line 341: btw, from what i can tell, the typing here is non-nullable already so it should be safe to make the type of `responseHeaders` non-nullable without needing to handle it

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
evan-masseau: Attempts is not persisted right now on either platform

evan-masseau: Cut a ticket for discussion about persisting attempt count to disk. I think it makes more sense to be persisted, its probably how we _intended_ that 50 retry limit to behave. Best we have a team chat about this though, since it hasn't been working that way up till now.

kennyklaviyo (approved):

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -126,6 +133,15 @@ internal open class KlaviyoApiRequest(
             }
         }
 
+    /**
+     * Tracks number of attempts to limit retries
+     */
+    final override var attempts = 0
```
ajaysubra, line 139: should the default be 1?

ajaysubra (approved):
LGTM, just one question.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
evan-masseau: Defaults to 0 and always increments before send (there's a test for this). That allows distinguishing a totally unsent request from first attempt actually pending.

PR #150, opened 2024-03-25
Deprecate duplicative profile key constants

evan-masseau:
# Description
Mark these keys deprecated in advance of next major release in which we can change them to internal. They're inconsistent with iOS and duplicative of explicit setter methods anyway. 

# Check List

- [x] Are you changing anything with the public API? - YES, deprecating public constants
- [x] Are your changes backwards compatible with previous SDK Versions? - YES, deprecating = no change required yet.
- [x] Have you added unit test coverage for your changes? N/A

## Changelog / Code Overview

Deprecate the profile keys that correspond to identifiers, which have their own setter methods. 
- [ ] TODO add new constructor so identifiers can optionally be set from constructor

## Related Issues/Tickets

CHNL-6547

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,12 +112,15 @@ Identifiers are persisted to local storage so that the SDK can keep track of the
 Profile identifiers and other attributes can be set all at once using the `Profile` data class:
 
 ```kotlin
-val profile = Profile()
-    .setEmail("kermit@example.com")
-    .setPhoneNumber("+12223334444")
-    .setExternalId("USER_IDENTIFIER")
-    .setProperty(ProfileKey.FIRST_NAME, "Kermit")
-    .setProperty(ProfileKey.CUSTOM("instrument"), "banjo")
+val profile = Profile(
+    externalId = "USER_IDENTIFIER",
+    email = "kermit@example.com",
+    phoneNumber = "+12223334444",
+    properties = mapOf(
+        ProfileKey.FIRST_NAME to "Kermit",
+        ProfileKey.CUSTOM("instrument") to "banjo"
+    )
+)
 
 Klaviyo.setProfile(profile)
 ```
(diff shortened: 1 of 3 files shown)


ajaysubra (approved):

kennyklaviyo (approved):

evan-masseau:
Safe to merge now that the doc bot won't mess up previous verion's dokka docs.

README.md
```diff
@@ -109,15 +109,16 @@ A profile can be identified by any combination of the following:
 
 Identifiers are persisted to local storage so that the SDK can keep track of the current profile.
 
-Profile identifiers and other attributes can be set all at once as a `map`:
+Profile identifiers and other attributes can be set all at once using the `Profile` data class:
 
 ```kotlin
-val profile = Profile(
-    mapOf(
-        ProfileKey.EMAIL to "kermit@example.com",
```
evan-masseau, line 117: In hindsight, this syntax of initializing in all one map is exactly why I made those three keys public. Oh well.

kennyklaviyo (approved):

README.md
evan-masseau: Spoke offline, we're going to add a new constructor that can take external_id/email/phone in addition to the map. Not sure we even need to deprecate the original constructor.

README.md
```diff
@@ -109,24 +109,28 @@ A profile can be identified by any combination of the following:
...
 val profile = Profile(
-    mapOf(
-        ProfileKey.EMAIL to "kermit@example.com",
-        ProfileKey.FIRST_NAME to "Kermit"
+    externalId = "USER_IDENTIFIER",
+    email = "kermit@example.com",
+    phoneNumber = "+12223334444",
+    properties = mapOf(
+        ProfileKey.FIRST_NAME to "Kermit",
+        ProfileKey.CUSTOM("instrument") to "banjo"
```
kennyklaviyo, line 121: should so also update the `setProperty` setter to look for Email, Phone, ExternalID keys and route them to the dedicated setters?

evan-masseau: Doesn't make much difference, dedicated setter still just does `this[ProfileKey.PHONE_NUMBER] = value`. I guess other than that you could set them to any `serializable`, not just `String`

PR #151, opened 2024-03-25
Stateful push notification API to avoid unneeded calls

evan-masseau:
# Description
This will prevent duplicate push token API calls by tracking push token state locally. 

# Check List

  - NO
  - YES
  - TODO
  - YES
  - N/A

## Changelog / Code Overview
To avoid duplicate/unnecessary push token API calls:
- I moved push token into UserInfo state object
- And added permission and background into state as well
- Use that prior state to tell if we need an API request when setPushToken is called

## Test Plan
Wrote unit tests already, doing manual tests on devices now. 

## Related Issues/Tickets
CHNL-6782

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
@@ -167,15 +167,19 @@ object Klaviyo {
      * @param pushToken The push token provided by the device push service
      */
     fun setPushToken(pushToken: String) = safeApply {
-        Registry.dataStore.store(EventKey.PUSH_TOKEN.name, pushToken)
-        Registry.get<ApiClient>().enqueuePushToken(pushToken, UserInfo.getAsProfile())
+        if (UserInfo.shouldUpdatePush(pushToken)) {
+            UserInfo.setPushToken(pushToken)
+            Registry.get<ApiClient>().enqueuePushToken(pushToken, UserInfo.getAsProfile())
+        } else {
+            Registry.log.verbose("Push token is already set")
+        }
     }
 
     /**
      * @return The device push token, if one has been assigned to currently tracked profile
      */
     fun getPushToken(): String? = safeCall {
-        Registry.dataStore.fetch(EventKey.PUSH_TOKEN.name)?.ifEmpty { null }
+        UserInfo.pushToken.ifEmpty { null }
     }
 
     /**
@@ -226,9 +230,6 @@ object Klaviyo {
 
         // Clear profile identifiers from state
         UserInfo.reset()
-
-        // If we had a push token, erase the local copy
-        Registry.dataStore.clear(EventKey.PUSH_TOKEN.name)
     }
 
     /**
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/UserInfo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/UserInfo.kt
@@ -1,11 +1,18 @@
 package com.klaviyo.analytics
 
+import com.klaviyo.analytics.DeviceProperties.backgroundData
+import com.klaviyo.analytics.DeviceProperties.notificationPermission
 import com.klaviyo.analytics.model.Profile
 import com.klaviyo.analytics.model.ProfileKey
 import com.klaviyo.analytics.model.ProfileKey.ANONYMOUS_ID
 import com.klaviyo.analytics.model.ProfileKey.EMAIL
 import com.klaviyo.analytics.model.ProfileKey.EXTERNAL_ID
 import com.klaviyo.analytics.model.ProfileKey.PHONE_NUMBER
+import com.klaviyo.analytics.model.ProfileKey.PUSH_STATE
+import com.klaviyo.analytics.model.ProfileKey.PUSH_TOKEN
+import com.klaviyo.analytics.networking.ApiClient
+import com.klaviyo.analytics.networking.requests.KlaviyoApiRequest.Status.Complete
+import com.klaviyo.analytics.networking.requests.PushTokenApiRequest
 import com.klaviyo.core.Registry
 import java.util.UUID
 
@@ -13,6 +20,19 @@ import java.util.UUID
  * Stores information on the currently active user
  */
 internal object UserInfo {
+
+    init {
+        Registry.get<ApiClient>().onApiRequest {
+            // Update state when a push token is successfully saved to backend
+            // Make sure it's still the same token we're tracking in state
+            if (it is PushTokenApiRequest &&
+                it.status == Complete &&
+                it.token == pushToken
+            ) {
+                setPushToken(pushToken, it.notificationPermission, it.backgroundData)
+            }
+        }
+    }
 
     /**
      * Save or clear an identifier in the persistent store and return it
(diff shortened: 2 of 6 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/UserInfo.kt
```diff
@@ -1,11 +1,18 @@
 package com.klaviyo.analytics
 
+import com.klaviyo.analytics.DeviceProperties.backgroundData
```
evan-masseau, line 3: I don't love what this file is turning into. I had solutions in our very first SDK architecture designs that would separate state itself from important side-effects. I would like to fast follow this PR by implementing those designs. Roughly speaking that PR would move some of these changes to other places, but the overall logic would remain the same. Just better organized.

sdk/analytics/src/main/java/com/klaviyo/analytics/UserInfo.kt
```diff
@@ -97,4 +131,23 @@ internal object UserInfo {
...
+        pushToken = token
+        pushState = formatPushState(pushToken, permission, background)
+    }
+
+    private fun formatPushState(
+        token: String,
+        permission: Boolean? = null,
+        background: Boolean? = null
+    ): String =
+        "$token-${permission ?: notificationPermission}-${background ?: backgroundData}"
```
jordan-griffin, line 152: Sorry am I overlooking something, but where is `notificationPermission` getting set?

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -167,15 +167,19 @@ object Klaviyo {
      * @param pushToken The push token provided by the device push service
      */
     fun setPushToken(pushToken: String) = safeApply {
-        Registry.dataStore.store(EventKey.PUSH_TOKEN.name, pushToken)
```
jordan-griffin, line 170: What was our reason for popping push tokens into a different store originally? `UserInfo` was just initially designed to store top-level identifiers?

jordan-griffin (approved):

jordan-griffin:
Approved but do have one question. Looks real solid

sdk/analytics/src/main/java/com/klaviyo/analytics/UserInfo.kt
evan-masseau: Oh yeah that reads funny, without context. `notificationPermission` is an imported member off the DeviceProperties object. A computed var that gets permission setting from the device (background is the same)

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
evan-masseau: Yeah, and we were trying to separate push functionality from other analytics

evan-masseau: At one point I think the whole setpushtoken function was defined in the push model even

jordan-griffin: Ah gotcha, thanks for the context. Makes sense

ajaysubra (approved):

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
```diff
@@ -70,17 +70,23 @@ internal class PushTokenApiRequest(
         )
     }
 
+    val notificationPermission by lazy { DeviceProperties.notificationPermission }
+
+    val backgroundData by lazy { DeviceProperties.backgroundData }
+
+    val token: String? get() = body?.optJSONObject(DATA)?.optString(TOKEN)
```
kennyklaviyo, line 77: should these be public?

kennyklaviyo: and what exactly does this achieve?

sdk/analytics/src/main/java/com/klaviyo/analytics/UserInfo.kt
```diff
@@ -97,4 +131,23 @@ internal object UserInfo {
             profile.setPhoneNumber(phoneNumber)
         }
     }
+
+    fun shouldUpdatePush(
```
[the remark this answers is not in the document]
evan-masseau: Ah i see. Well maybe I could just hash a whole payload instead.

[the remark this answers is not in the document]
evan-masseau: Actually this is a _very_ good catch. With my code here, we wouldn't get updated device metadata like new build number or whatever until the token or permission changed 😬

sdk/analytics/src/main/java/com/klaviyo/analytics/UserInfo.kt
```diff
@@ -22,14 +20,10 @@ import java.util.UUID
...
-        Registry.get<ApiClient>().onApiRequest {
-            // Update state when a push token is successfully saved to backend
-            // Make sure it's still the same token we're tracking in state
-            if (it is PushTokenApiRequest &&
-                it.status == Complete &&
-                it.token == pushToken
-            ) {
-                setPushToken(pushToken, it.notificationPermission, it.backgroundData)
+        Registry.get<ApiClient>().onApiRequest { request ->
+            // If push token request totally fails, we must remove it from state
```
jordan-griffin, line 24: Is this some mechanism to enforce us to send it again in the future since next time we call for push token it'll always be a value thats not empty?

evan-masseau: Exactly, this is the failure mode of the "optimistic update" approach. If a request fails to get to the backend, clearing pushState will mean a "duplicate" request could be enqueued again

sdk/analytics/src/main/java/com/klaviyo/analytics/UserInfo.kt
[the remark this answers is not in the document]
evan-masseau: now tracking the whole request body as state, thanks for this catch

sdk/analytics/src/main/java/com/klaviyo/analytics/UserInfo.kt
```diff
@@ -56,9 +70,42 @@ internal object UserInfo {
...
+
+        if (newPushState != pushState) {
+            // Optimistic update algorithm: expect request to get to backend,
+            // on failure reset push state (see initializer). The main advantage to
+            // this algorithm is it prevents queueing duplicate requests immediately
+            pushState = newPushState ?: ""
+            pushToken = token
+            onChanged()
+        }
+    }
```
evan-masseau, line 103: Said it earlier, but the comment got lost.
I don't love how state and side effects are being managed, especially with this addition. I had solutions in our very first SDK architecture designs that would separate public API (`Klaviyo` object), state (`UserInfo` object) and side-effects. I would like to fast follow this PR by implementing those designs

kennyklaviyo (approved):

PR #160, opened 2024-04-22
Separation of API, State, and Side Effects

evan-masseau:
# Description
Separation of concerns: 
- `Klaviyo` exposes the public API, including initialize logic.
- `UserInfo` handles internal user state including persistence
- `SideEffectCoordinator` observes and responds to state changes 

# Check List

  - No, this should be an internal re-organization only
  - Yes
  - I tried to make my tests purely additive, aside from some required setup and mocking changes.

## Changelog / Code Overview

- 

## Test Plan

- Existing tests
- [x] I've added new tests, but have some more to go
- [x] Update internal test app

## Related Issues/Tickets

CHNL-7012


sdk/analytics/src/main/java/com/klaviyo/analytics/state/UserState.kt
```diff
@@ -0,0 +1,146 @@
...
+    override var externalId by _externalId
+
+    private val _email = PersistentObservableString(EMAIL, ::broadcastChange)
+    override var email by _email
+
+    private val _phoneNumber = PersistentObservableString(PHONE_NUMBER, ::broadcastChange)
+    override var phoneNumber by _phoneNumber
+
+    private val _anonymousId = PersistentObservableString(ANONYMOUS_ID, ::broadcastChange) {
+        UUID.randomUUID().toString()
```
evan-masseau, line 33: This is similar to the pattern we used for auto-generating the anonymous ID previously. I'm considering perhaps moving that behavior to the new side effects class, but opted for the lesser change for now.

evan-masseau: Oh, to elaborate, this method is called as a fallback by `PersistentObservableString` when a value is not currently stored for this key in the persistent store. Thats what we did before too, but all the logic lived inside `UserInfo` as setter/getter functions of the `anonymousId` property.

sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoTest.kt
```diff
@@ -51,53 +51,50 @@ internal class KlaviyoTest : BaseTest() {
...
-        every { Registry.clock } returns staticClock
-        every { apiClientMock.onApiRequest(any(), any()) } returns Unit
-        every { apiClientMock.enqueueProfile(capture(capturedProfile)) } returns Unit
-        every { apiClientMock.enqueueEvent(any(), any()) } returns Unit
-        every { apiClientMock.enqueuePushToken(any(), any()) } returns Unit
-        every { configMock.debounceInterval } returns debounceTime
+        every { Registry.configBuilder } returns builderMock
         DevicePropertiesTest.mockDeviceProperties()
-        UserInfo.reset()
+        Klaviyo.initialize(
```
evan-masseau, line 89: Had to adjust the setup for these tests because of how I changed `UserInfo` from an Object to a registered service that needs to be instantiated.

sdk/analytics/src/main/java/com/klaviyo/analytics/state/StateSideEffects.kt
```diff
@@ -0,0 +1,85 @@
...
+                }
+                else -> onUserStateChange()
+            }
+        }
+    }
+
+    private fun onPushStateChange() {
+        if (!state.pushState.isNullOrEmpty()) {
+            state.pushToken?.let { apiClient.enqueuePushToken(it, state.get()) }
+        }
```
kennyklaviyo, line 45:
```suggestion
        state.pushToken.takeIf { !it.isNullOrEmpty() }?.let { apiClient.enqueuePushToken(it, state.get()) }
```

sdk/analytics/src/main/java/com/klaviyo/analytics/state/StateSideEffects.kt
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    private fun onUserStateChange() {
+        val profile = state.get(true)
```
kennyklaviyo, line 49:
```suggestion
        val profile = state.get(withAttributes = true)
```
prefer using keywords so its clearer when reading what the args are for

sdk/analytics/src/main/java/com/klaviyo/analytics/state/StateSideEffects.kt
evan-masseau: The if condition is looking at `pushState` though. Could still use this syntax, but it really shorten anything.

sdk/analytics/src/main/java/com/klaviyo/analytics/state/PersistentObservableProperty.kt
```diff
@@ -0,0 +1,97 @@
...
+     * Deserialize from persistent store
+     */
+    abstract fun deserialize(storedValue: String?): T?
+
+    /**
+     * Save or clear property in the persistent store
+     */
+    private fun persist(value: T?) = value?.toString()?.also { serializedValue ->
+        Registry.dataStore.store(key.name, serializedValue)
+    } ?: Registry.dataStore.clear(key.name).let { "" }
```
kennyklaviyo, line 87: whats the trailing `.let { "" }` for?

evan-masseau: Ah thats from an earlier draft where this needed to return a string. I optimized that out, this can return Unit now.

sdk/analytics/src/main/java/com/klaviyo/analytics/state/State.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Update user state from a new [Profile] model object
+     */
+    fun set(profile: Profile)
+
+    /**
+     * Set an individual attribute
+     */
+    fun set(key: ProfileKey, value: String)
```
kennyklaviyo, line 44: hmm reading around, i feel that these names can be a little clearer, `getProfile`, `setProfile`, `setAttribute`

evan-masseau: fair, this is swift-y, but without the argument label enforcement of swift

sdk/analytics/src/main/java/com/klaviyo/analytics/state/StateSideEffects.kt
kennyklaviyo: ops missed that

sdk/analytics/src/main/java/com/klaviyo/analytics/state/State.kt
kennyklaviyo: when reading within the `State` class, when I see `get()` its a bit of processing to remember that `get()` returns a `Profile` otherwise initially i would expect it to return a copy of `State`

evan-masseau: Yeah it used to be `UserInfo.getAsProfile`, I'm going to go back to that.

sdk/analytics/src/main/java/com/klaviyo/analytics/state/StateSideEffects.kt
```diff
@@ -31,7 +31,7 @@ internal class StateSideEffects(
             when (key) {
                 ProfileKey.PUSH_STATE -> onPushStateChange()
                 ProfileKey.PUSH_TOKEN -> { /* Token is a no-op, push changes are captured by push state */ }
-                PROFILE_ATTRIBUTES -> if (state.get(true).attributes.propertyCount() > 0) {
+                PROFILE_ATTRIBUTES -> if (state.getAsProfile(true).attributes.propertyCount() > 0) {
```
kennyklaviyo, line 34:
```suggestion
                PROFILE_ATTRIBUTES -> if (state.getAsProfile(withAttributes 
 = true).attributes.propertyCount() > 0) {
```

kennyklaviyo (approved):

PR #162, opened 2024-05-01
Add Unregister Push Token Request

kennyklaviyo:
# Description

# Check List


sdk/analytics/src/test/java/com/klaviyo/analytics/networking/requests/UnregisterPushTokenApiRequestTest.kt
```diff
@@ -0,0 +1,80 @@
...
+                      "type": "profile",
+                      "attributes": {
+                        "email": "$EMAIL",
+                        "phone_number": "$PHONE",
+                        "external_id": "$EXTERNAL_ID",
+                        "anonymous_id": "$ANON_ID"
+                      }
+                    }
+                  },
+                  "device_metadata": {
```
evan-masseau, line 58: Oh actually, does unregister need device metadata?

kennyklaviyo: haha yeah i was just removing this

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/UnregisterPushTokenApiRequest.kt
```diff
@@ -0,0 +1,87 @@
+package com.klaviyo.analytics.networking.requests
```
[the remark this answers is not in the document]
evan-masseau: NVM, i think without the metadata there will be very little duplication i think

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/UnregisterPushTokenApiRequest.kt
```diff
@@ -0,0 +1,61 @@
...
+                    PLATFORM to DeviceProperties.platform,
+                    VENDOR to VENDOR_FCM
+                )
+            )
+        )
+    }
+
+    override fun equals(other: Any?): Boolean {
+        return when (other) {
+            is UnregisterPushTokenApiRequest -> body.toString() == other.body.toString()
```
evan-masseau, line 53: Ooh, will company ID need to be accounted for here? Same would go for the register call.

evan-masseau: Although, I guess anonymous ID should always be present and different if company has changed.

evan-masseau (approved):
Looks good, I think we can test that question about company ID in the next phase.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/UnregisterPushTokenApiRequest.kt
kennyklaviyo: ill merge this so i can test with your changes but i think we'll just end up removing this entirely, no need to dedup unregister i think

PR #163, opened 2024-05-02
Trigger Unregister on API Key Change

kennyklaviyo:
# Description

# Check List

### DIFF
--- a/versions.properties
+++ b/versions.properties
@@ -9,29 +9,20 @@
 ####
 #### NOTE: Some versions are filtered by the rejectVersionIf predicate. See the settings.gradle.kts file.
 
-## unused
-version.androidx.compose.compiler=1.5.13
-
 # Project versioning, run the following gradle command to update version numbers automatically:
 # ./gradlew bumpVersion --nextVersion=X.Y.Z
-## unused
 version.klaviyo.versionCode=18
 
-## unused
 version.klaviyo.versionName=3.0.0-alpha.1
 
 # Android versioning
 
-## unused
 version.android.minSdk=23
 
-## unused
 version.android.targetSdk=34
 
-## unused
 version.android.compileSdk=34
 
-## unused
 version.android.buildTools=34.0.0
 
 # Project dependencies


sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -54,6 +62,7 @@ object Klaviyo {
 
         Registry.register<State>(KlaviyoState())
         Registry.register<StateSideEffects>(StateSideEffects())
+        Registry.get<State>().apiKey = apiKey
```
kennyklaviyo, line 65: trigger an apiKey `braodcastChange` if necessary, results in registering a push token

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
```diff
@@ -78,7 +78,7 @@ internal class PushTokenApiRequest(
 
     override fun equals(other: Any?): Boolean {
         return when (other) {
-            is PushTokenApiRequest -> body.toString() == other.body.toString()
+            is PushTokenApiRequest -> body.toString() == other.body.toString() && query.toString() == other.query.toString()
```
kennyklaviyo, line 81: since query contains api key

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/UnregisterPushTokenApiRequest.kt
```diff
@@ -50,7 +50,7 @@ internal class UnregisterPushTokenApiRequest(
 
     override fun equals(other: Any?): Boolean {
         return when (other) {
-            is UnregisterPushTokenApiRequest -> body.toString() == other.body.toString()
+            is UnregisterPushTokenApiRequest -> body.toString() == other.body.toString() && query.toString() == other.query.toString()
```
kennyklaviyo, line 53: since query contains api key

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -39,6 +39,14 @@ object Klaviyo {
      * @param applicationContext
      */
     fun initialize(apiKey: String, applicationContext: Context) = safeCall {
+        if (Registry.isRegistered<Config>() && Registry.get<Config>().apiKey != apiKey) {
+            Registry.get<State>().let { state ->
```
evan-masseau, line 43: This logic could live in side effects coordinator now, right?

versions.properties
```diff
@@ -9,29 +9,48 @@
 ####
 #### NOTE: Some versions are filtered by the rejectVersionIf predicate. See the settings.gradle.kts file.
 
+## unused
```
evan-masseau, line 12: I'd take this dep out, and all the `#unused` comments on our custom version properties... they are used, via the `versionFor` function, but I guess the plugin can't detect that.

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
evan-masseau: ooh sorry right because unregister needs the old api key value. hm. 
It would be a pretty easy add to "broadcast" the old value of the property. I had that originally, but none of my callbacks were using it, so I simplified.

kennyklaviyo: i can work to rewriting the request to take in the api key in the constructor instead?

evan-masseau: Ah yeah good point. LMK If you want to pair/discuss adding in the broadcast of the property's old value as well.

evan-masseau: I remember now that I didn't do it partly because none of my callbacks needed the old value and partly because I didn't want to deal with trying to make `StateObserver` generic. Maybe it'll just  have to be `typealias StateObserver = (key: Keyword?, oldValue: Any?) -> Unit`

sdk/analytics/src/main/java/com/klaviyo/analytics/model/ProfileKey.kt
```diff
@@ -13,6 +13,7 @@ sealed class ProfileKey(name: String) : Keyword(name) {
     internal object ANONYMOUS_ID : ProfileKey("anonymous_id")
     internal object PUSH_TOKEN : ProfileKey("push_token")
     internal object PUSH_STATE : ProfileKey("push_state")
+    internal object API_KEY : ProfileKey("api_key")
```
jordan-griffin, line 16: Maybe just semantics to how state is organized in the Android SDK but this isn't really a _profile_ property, right? Is there potential this could be better exposed from some other place?

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
evan-masseau: I actually still had this partially implemented in a stash locally. I cleaned it up and put it in a [draft PR](https://github.com/klaviyo/klaviyo-android-sdk/pull/164) to share

sdk/analytics/src/main/java/com/klaviyo/analytics/model/ProfileKey.kt
evan-masseau: Truth, could do this instead
```
internal object API_KEY : Keyword("api_key")
```

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
kennyklaviyo: @evan-masseau should i also update the hashCode function too?

sdk/analytics/src/main/java/com/klaviyo/analytics/state/PersistentObservableString.kt
```diff
@@ -1,11 +1,11 @@
 package com.klaviyo.analytics.state
 
-import com.klaviyo.analytics.model.ProfileKey
+import com.klaviyo.analytics.model.Keyword
 import com.klaviyo.core.Registry
 import kotlin.reflect.KProperty
 
 internal class PersistentObservableString(
-    key: ProfileKey,
+    key: Keyword,
```
kennyklaviyo, line 8: this is why i originally had the API_KEY subclass from ProfileKey, seems like changing this type should be fine but lmk if you feel otherwise

evan-masseau: Ooh I see, no its better if this is `Keyword`!

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/PushTokenApiRequest.kt
evan-masseau: ah yeah totally. actually that makes me wonder if we should just be comparing the hashcodes in this equals function

ajaysubra (approved):

PR #168, opened 2024-05-20
Using token endpoint for profile requests when a push token is present in SDK state

ajaysubra:
# Description

Previously we used Klaviyo's `client/profiles` endpoint for all profile related updates and the `client/push-tokens` endpoint for just setting the push token when it was passed to the SDK. This works fine for most cases but in the case a profile changes, we would require developers to call our SDK's `setPushToken` method to reset the push token when seemed unnecessary since push token doesn't change with the changing profile.  

To address this, we are now using the token endpoint for even the profile updates so that when a profile or any profile identifier changes the token is automatically moved over to the new profile w/o the developer needing to do this. This should not only lower the number of calls to our endpoint but also improve network bandwidth consumption of host apps.

One thing to note is if a push token is not set then we will always call the profiles endpoint.

Below is a table reflecting the change in more detail - 

Token Set? | SDK method | Tokens API | Profiles API
-- | -- | -- | -- |
⛔️ | set(profile: Profile) |   | ✅
⛔️ | set(email: String) |   | ✅
⛔️ | set(phoneNumber: String) |   | ✅
⛔️ | set(externalId: String) |   | ✅
⛔️ | set(profileAttribute: Profile.ProfileKey, value: Any) |   | ✅
✅ | set(profile: Profile) | ✅ |  
✅ | set(email: String) | ✅ |  
✅ | set(phoneNumber: String) | ✅ |  
✅ | set(externalId: String) | ✅ |  
✅ | set(profileAttribute: Profile.ProfileKey, value: Any) | ✅ |  

# Check List

## Test Plan

Installed a proxy and made sure that all profile related calls like set profile, set email, set external id were routed through the profiles endpoint when a push token wasn't set. When a push token was set in the SDK state verified that all profiles related called including then profile attributes were made to the token endpoint.

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
@@ -128,7 +128,6 @@ internal class KlaviyoState : State {
         _phoneNumber.reset()
         _anonymousId.reset()
         _attributes.reset()
-        _pushToken.reset()
         _pushState.reset()
 
         broadcastChange(null, oldProfile)
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
@@ -128,7 +128,6 @@ internal class KlaviyoState : State {
         _phoneNumber.reset()
         _anonymousId.reset()
         _attributes.reset()
-        _pushState.reset()
 
         broadcastChange(null, oldProfile)
         Registry.log.verbose("Reset internal user state")
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/state/StateSideEffects.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/state/StateSideEffects.kt
@@ -87,7 +87,9 @@ internal class StateSideEffects(
     private fun flushProfile() = pendingProfile?.let {
         timer?.cancel()
         Registry.log.verbose("Flushing profile update")
-        apiClient.enqueueProfile(it.copy())
+        state.pushToken?.let {
+            apiClient.enqueuePushToken(it, state.getAsProfile())
+        } ?: apiClient.enqueueProfile(it.copy())
         state.resetAttributes() // Once captured in a request, we don't keep profile attributes in state/on disk
         pendingProfile = null
     }
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/state/StateSideEffects.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/state/StateSideEffects.kt
@@ -4,6 +4,7 @@ import com.klaviyo.analytics.model.API_KEY
 import com.klaviyo.analytics.model.ImmutableProfile
 import com.klaviyo.analytics.model.Keyword
 import com.klaviyo.analytics.model.PROFILE_ATTRIBUTES
+import com.klaviyo.analytics.model.Profile
 import com.klaviyo.analytics.model.ProfileKey
 import com.klaviyo.analytics.networking.ApiClient
 import com.klaviyo.analytics.networking.requests.ApiRequest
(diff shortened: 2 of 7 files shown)


ajaysubra:
Looking at the failing tests.

ajaysubra:
> Looking at the failing tests.

Fixed tests. Looking at adding some tests.

evan-masseau (changes requested):
A+ Branch name. 

When `resetProfile` is called, are we meant to immediately enqueue a token API request to register the token to a new anonymous profile (i.e. is that what iOS does, and/or is that what we _should_ be doing)? Let's add a test to this PR that verifies the expected push token API behavior after `resetProfile`

evan-masseau:
```
com.klaviyo.analytics.KlaviyoUninitializedTest > Push token getter is protected FAILED
    java.lang.AssertionError at KlaviyoUninitializedTest.kt:95
```
This test might be relying on reset state to clear the push token from the data store, might have to update the setup or teardown function.

sdk/analytics/src/main/java/com/klaviyo/analytics/state/StateSideEffects.kt
```diff
@@ -87,11 +88,27 @@ internal class StateSideEffects(
...
+     * Enqueue pending profile changes as an API call to either push token endpoint or profile endpoint.
+     *
+     * Why? - Profile changes are sent to push token API when there is a push token present in state.
+     * This is done to avoid resetting the push token in state, making a profile request and then another
+     * request to the push token endpoint to set the push token.
+     *
+     * By just using the push token API we can avoid the extra request and also ensure that the push token
+     * is set on the new profile in Klaviyo.
+     */
+    private fun enqueueTokenOrProfile(profile: Profile) {
```
evan-masseau, line 106: 👍

sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoTest.kt
```diff
@@ -399,6 +398,54 @@ internal class KlaviyoTest : BaseTest() {
...
+    @Test
+    fun `Push token request is made for profile attributes when token is set`() {
+        Klaviyo.setPushToken(PUSH_TOKEN)
+        assertEquals(PUSH_TOKEN, dataStoreSpy.fetch("push_token"))
+
+        verify(exactly = 1) {
+            apiClientMock.enqueuePushToken(PUSH_TOKEN, any())
+        }
+
+        Klaviyo.setProfileAttribute(ProfileKey.FIRST_NAME, "Larry")
```
evan-masseau, line 442: pretty... pretty good.

ajaysubra:
> KlaviyoUninitializedTest

Can you help me understand which test? None of the tests in `KlaviyoUninitializedTest` is failing for me.

ajaysubra:
> A+ Branch name.
> 
> When `resetProfile` is called, are we meant to immediately enqueue a token API request to register the token to a new anonymous profile (i.e. is that what iOS does, and/or is that what we _should_ be doing)? Let's add a test to this PR that verifies the expected push token API behavior after `resetProfile`

Luckily since `flushProfile` was already being called when a profile was reset and my changes were downstream from there I didn't have to do anything additional. Also, since now when resetting we don't clear the push token state, my downstream changes to can make a token request for any profile updates when a token is set. 

Updated the existing test and added one for verifying that token endpoint is indeed being called when resetting.

sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
```diff
@@ -128,7 +128,6 @@ internal class KlaviyoState : State {
         _phoneNumber.reset()
         _anonymousId.reset()
         _attributes.reset()
-        _pushToken.reset()
```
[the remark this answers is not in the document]
ajaysubra: Yeah. Since there is no near term need for a full reset I'm just vary of leaving dead code paths in. However, you have better context of this codebase so let me know if you think otherwise.

ajaysubra:
> > KlaviyoUninitializedTest
> 
> Can you help me understand which test? None of the tests in `KlaviyoUninitializedTest` is failing for me.

nvm I see it now. Looking into fixing it.

sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoUninitializedTest.kt
```diff
@@ -92,7 +92,7 @@ internal class KlaviyoUninitializedTest {
 
     @Test
     fun `Push token getter is protected`() {
-        assertNull(Klaviyo.getPushToken())
+        Klaviyo.getPushToken()
```
evan-masseau, line 95: Might be a pre-existing issue, but if this is returning a token it means some test(s) are "leaking", I'm going to take a quick look at that. I'd prefer to keep the `assertNull` and plug the leak

ajaysubra: Yeah, I wasn't sure what was causing this since I'm unable to trigger the failure running the tests for android studio. Debugging with breakpoints wasn't working either but one thing that I did notice is if I added the line that I removed where we reset the token this test passes.

evan-masseau: Yeah its an interaction between tests, somehow state and/or data store are being shared. I can repro it if i run all tests without breakpoints. Running tests separately is fine, and adding a breakpoint must throw off the timing.

evan-masseau (approved):

PR #169, opened 2024-05-22
Fix some setup/teardown in tests to take care of isolation issues.

evan-masseau:
To fix an issue introduced when I made state into a registered service: `KlaviyoTest` is leaking some registered services.

I have an idea for a better long-term fix that'd change `Registry` a little bit to make it easier to isolate tests.

### DIFF
--- a/sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoTest.kt
+++ b/sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoTest.kt
@@ -10,6 +10,7 @@ import com.klaviyo.analytics.model.Profile
 import com.klaviyo.analytics.model.ProfileKey
 import com.klaviyo.analytics.networking.ApiClient
 import com.klaviyo.analytics.state.State
+import com.klaviyo.analytics.state.StateSideEffects
 import com.klaviyo.core.Registry
 import com.klaviyo.core.config.Config
 import com.klaviyo.fixtures.BaseTest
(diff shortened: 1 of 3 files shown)


sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoTest.kt
```diff
@@ -94,6 +95,10 @@ internal class KlaviyoTest : BaseTest() {
     @After
     override fun cleanup() {
         Registry.get<State>().reset()
+        Registry.unregister<Config>()
```
evan-masseau, line 98: This is the key part for the failing test on Ajay's PR.

sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoUninitializedTest.kt
```diff
@@ -6,36 +6,41 @@ import com.klaviyo.analytics.model.ProfileKey
...
     fun setup() {
-        Registry.unregister<Config>()
         Registry.register<Log>(logger)
         Registry.register<ApiClient>(mockApiClient)
-        every { mockApiClient.onApiRequest(any(), any()) } returns Unit
+    }
+
+    @After
+    fun cleanup() {
+        Registry.unregister<Log>()
```
evan-masseau, line 42: This was a similar leak, just wasn't causing any failed tests (yet)

ajaysubra (approved):
Thanks! LGTM.

sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoTest.kt
ajaysubra: Can you tell me more on how this was causing the issue. I would have thought it's related to state.

evan-masseau: Oh sorry, it is state, commented on the wrong line

evan-masseau: The issue was that in the `setup` of this test class we call `Klaviyo.initialize` and within initialize we register config, state, state side effects, so in teardown we have to unregister those. 
My idea for a bigger fix would be to isolate registry a bit more so that each test starts with a clean registry and you don't have to unregister individually like this

PR #170, opened 2024-05-22
API Header Fix

evan-masseau:
Move attempt count increment to before we write request headers into the url connection.

# Description
Made a silly mistake in the attempt count header. The request header value is updated correctly... but only _after_ the request header has already been written in to the url connection object. Functionally this means our requests still only retry up to 50 times, but the header that the server sees will be `0/50` -> `49/50`

# Check List

- [x] Have you tested this change on real device? YES
- [x] Have you verified that your changes are compatible with all Android versions the SDK currently supports? YES

## Changelog / Code Overview
Fix attempt count header that is actually sent to server

## Test Plan
Fixed the unit test that should have caught this, so it is verifying the actual header value written to url connection. 

## Related Issues/Tickets
CHNL-8887

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
@@ -282,12 +282,12 @@ internal open class KlaviyoApiRequest(
         }
 
         status = Status.Inflight
+        attempts++
 
         return try {
             val connection = buildUrlConnection()
 
             try {
-                attempts++
                 beforeSend.invoke()
                 connection.connect()
                 parseResponse(connection)
--- a/sdk/analytics/src/test/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequestTest.kt
+++ b/sdk/analytics/src/test/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequestTest.kt
@@ -129,14 +129,14 @@ internal class KlaviyoApiRequestTest : BaseApiRequestTest<KlaviyoApiRequest>() {
     }
 
     @Test
-    fun `Increments attempt counter on send`() {
-        withConnectionMock(URL(expectedFullUrl))
+    fun `Increments attempt counter on send and uses correct attempt number in header`() {
+        val connectionMock = withConnectionMock(URL(expectedFullUrl))
         val request = makeTestRequest()
 
         assertEquals(0, request.attempts)
         request.send()
         assertEquals(1, request.attempts)
-        assertEquals(request.headers["X-Klaviyo-Attempt-Count"], "1/50")
+        verify { connectionMock.setRequestProperty("X-Klaviyo-Attempt-Count", "1/50") }
     }
 
     @Test


sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -282,12 +282,12 @@ internal open class KlaviyoApiRequest(
         }
 
         status = Status.Inflight
+        attempts++
 
         return try {
             val connection = buildUrlConnection()
```
evan-masseau, line 288: Incrementing `attempts` is what updates the header value on this request object. But `buildUrlConnection()` is where we actually write the headers into the request. So incrementing has to come first.

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequestTest.kt
```diff
@@ -129,14 +129,14 @@ internal class KlaviyoApiRequestTest : BaseApiRequestTest<KlaviyoApiRequest>() {
...
-    fun `Increments attempt counter on send`() {
-        withConnectionMock(URL(expectedFullUrl))
+    fun `Increments attempt counter on send and uses correct attempt number in header`() {
+        val connectionMock = withConnectionMock(URL(expectedFullUrl))
         val request = makeTestRequest()
 
         assertEquals(0, request.attempts)
         request.send()
         assertEquals(1, request.attempts)
-        assertEquals(request.headers["X-Klaviyo-Attempt-Count"], "1/50")
```
evan-masseau, line 139: The test that should have caught this was only looking at the API request object, not what was written to the url connection. Once I'd updated this test, the prior code failed. Moving `attempts++` fixes the test.

evan-masseau: BTW the debug logs and the internal test app also display `KlaviyoApiRequest.headers`, which must be part of why this slipped through the cracks.

evan-masseau: It found it while looking in API developer logs serverside. Would have also been visible in a proxy, must not have tested that way.

ajaysubra: Yeah I've had better luck using proxy for testing stuff related to the SDK.

ajaysubra (approved):
LGTM.

ajaysubra:
Also, is this going into the android release branch or directly to master as a patch?

evan-masseau:
> Also, is this going into the android release branch or directly to master as a patch?

I guess just the 3.0 branch, its not worth a patch/hotfix

evan-masseau:
> > Also, is this going into the android release branch or directly to master as a patch?
> 
> I guess just the 3.0 branch, its not worth a patch/hotfix
> 
> Agree, do you need to update the base branch in this PR then?

Lol @ajaysubra i think you edited my comment there. 
Fixed based branch, thanks for noticing.

PR #171, opened 2024-05-24
Ecm/services refactor

evan-masseau:
# Description

# Check List


sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoUninitializedTest.kt
```diff
@@ -27,20 +26,15 @@ internal class KlaviyoUninitializedTest {
...
     }
 
-    private val mockApiClient = mockk<ApiClient>().apply {
-        every { onApiRequest(any(), any()) } returns Unit
-    }
-
     @Before
     fun setup() {
-        Registry.register<Log>(logger)
-        Registry.register<ApiClient>(mockApiClient)
```
evan-masseau, line 37: By mocking this in the "uninitialized" error state, I was masking an error that occurs when the (real) API client gets accessed prior to klaviyo.initialize.

PR #172, opened 2024-05-24
Refactor singleton service initializers

evan-masseau:
# Description
Update: Some of the work originally in this PR wound up in the 2.4 release instead. I've now merged master into rel/3.0 and then merged that into this branch. At this point, this PR is some lower priority changes to some service initializers and a lot of test fixture naming convention corrections.

We had some assumptions about the order in which dependencies are initialized, and encountered unrecoverable exceptions when those assumptions were incorrect. Most importantly, accessing `KlaviyoApiClient` before `applicationContext` was available would throw an exception in the initializer. An initialization error cannot be recovered, even if `applicationContext` were later provided, because `init` will only run once. 

# Check List

## Changelog / Code Overview

As a safety precaution, I've removed all the object initializers that had an implicit dependency, though the API client was the only one I've observed an issue with from testing with our test apps. 

## Test Plan

Updated all the unit tests already
Functional changes are smaller now that some of this branch's original work was cherry picked and released in 2.4 after testing.

## Related Issues/Tickets

CHNL-8569 -- this does not address all the AC of that ticket.

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
@@ -34,53 +34,20 @@ internal object KlaviyoApiClient : ApiClient {
      */
     private var apiObservers = mutableListOf<ApiObserver>()
 
-    init {
-        onApiRequest { r ->
-            when (r.state) {
-                Status.Unsent.name -> Registry.log.verbose("${r.type} Request enqueued")
-                Status.Inflight.name -> Registry.log.verbose("${r.type} Request inflight")
-                Status.PendingRetry.name -> {
-                    val attemptsRemaining = Registry.config.networkMaxAttempts - r.attempts
-                    Registry.log.warning(
-                        "${r.type} Request failed with code ${r.responseCode}, and will be retried up to $attemptsRemaining more times."
-                    )
-                }
-                Status.Complete.name -> Registry.log.verbose(
-                    "${r.type} Request succeeded with code ${r.responseCode}"
-                )
-                else -> Registry.log.error(
-                    "${r.type} Request failed with code ${r.responseCode}, and will be dropped"
-                )
-            }
+    /**
+     * Initialize logic including lifecycle observers and reviving the queue from persistent store
+     */
+    override fun startService() {
+        Registry.lifecycleMonitor.offActivityEvent(::onLifecycleActivity)
+        Registry.lifecycleMonitor.onActivityEvent(::onLifecycleActivity)
 
-            r.responseBody?.let { response ->
-                val body = r.requestBody?.let { JSONObject(it).toString(2) }
-                Registry.log.verbose("${r.httpMethod}: ${r.url}")
-                Registry.log.verbose("Headers: ${r.headers}")
-                Registry.log.verbose("Query: ${r.query}")
-                Registry.log.verbose("Body: $body")
-                Registry.log.verbose("${r.responseCode} $response")
-            }
+        Registry.networkMonitor.offNetworkChange(::onNetworkChange)
+        Registry.networkMonitor.onNetworkChange(::onNetworkChange)
+
+        if (getQueueSize() == 0) {
+            // We only need to restore queue from persistent store once
+            restoreQueue()
         }
-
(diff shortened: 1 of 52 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -34,53 +34,20 @@ internal object KlaviyoApiClient : ApiClient {
      */
     private var apiObservers = mutableListOf<ApiObserver>()
 
-    init {
```
evan-masseau, line 37: The critical part of this PR: an exception inside an `init` is re-thrown, and besides the initializer doesn't finish properly and can't be re-invoked for a singleton object. Instead, I'm moving these listener startups to a method that we invoke from `Klaviyo.initialize`

sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoTest.kt
```diff
@@ -61,34 +61,35 @@ internal class KlaviyoTest : BaseTest() {
     }
 
     private val capturedProfile = slot<Profile>()
-    private val apiClientMock: ApiClient = mockk<ApiClient>().apply {
+    private val mockApiClient: ApiClient = mockk<ApiClient>().apply {
```
evan-masseau, line 64: Naming convention fix, sometimes using mock as a prefix, sometimes as a suffix. Suggesting we consolidate to prefix.

sdk/core/src/main/java/com/klaviyo/core/KLog.kt
```diff
@@ -6,7 +6,26 @@ import com.klaviyo.core.config.Log
 import com.klaviyo.core.config.Log.Level
 import java.util.regex.Pattern
 
-open class KLog : Log {
+object KLog : Log {
```
evan-masseau, line 9: `KLog` doesn't need to be instantiated, it is essentially just a set of static methods anyway.

sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
```diff
@@ -14,10 +14,6 @@ internal object KlaviyoLifecycleMonitor : LifecycleMonitor, Application.Activity
 
     private var activityObservers = mutableListOf<ActivityObserver>()
 
-    init {
-        onActivityEvent { Registry.log.verbose(it.type) }
```
evan-masseau, line 18: This one wasn't causing any exceptions, but I thought it best to get rid of any unnecessary object inits to head off the same issues.

evan-masseau:
Reverted to draft, will need to update once 2.4.0 is release. I think these changes were still valuable.

evan-masseau (commented):
Merged master into 3.0 and that back into this branch, after the 2.4 release. 
At this point, this PR is some lower priority changes to some service initializers and a lot of test fixture naming convention corrections.

ajaysubra (approved):

sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
dan-peluso: Yeah inits can be notoriously hard to track - better safe than sorry

versions.properties
```diff
@@ -11,7 +11,7 @@
 
 # Project versioning, run the following gradle command to update version numbers automatically:
 # ./gradlew bumpVersion --nextVersion=X.Y.Z
-version.klaviyo.versionCode=18
+version.klaviyo.versionCode=19
```
dan-peluso, line 14: Is this going to be the last commit before we cut the release?

dan-peluso (approved):

versions.properties
evan-masseau: Hm not sure why I had this. I haven't reviewed the 3.0.0 release contents yet, so I'm not certain. We must be getting close though

evan-masseau:
```suggestion
version.klaviyo.versionCode=18
```
Lets bump this in a standalone PR

PR #173, opened 2024-05-24
Buffer failed operations that were attempted pre-initialize

evan-masseau:
# Description

Allow event operations to be retried if the event was logged prior to initialize. Critical to a timing issue on react native related to opened push events. This is branched off initial work in `ecm/services-refactor` so PR currently targets that branch. 

# Check List

- [x] Are you changing anything with the public API? - Nothing breaking, but this does change the behavior of `createEvent` and `handlePush` when uninitialized.
- [x] Are your changes backwards compatible with previous SDK Versions? - Yes

## Changelog / Code Overview

- Added an in-memory buffer of operations that were attempted prior to `Klaviyo.initialize` 
- Retry the buffer when initialization happens
- Applied this buffer to event creation since those are subject to potential timing/race condition problems.

## Test Plan

- New unit tests
- [ ] TODO Verify this fixes the react native timing issue for users who handle push from native code but initialize from typescript/js code.

## Related Issues/Tickets

CHNL-8569

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
@@ -13,11 +13,13 @@ import com.klaviyo.analytics.networking.KlaviyoApiClient
 import com.klaviyo.analytics.state.KlaviyoState
 import com.klaviyo.analytics.state.State
 import com.klaviyo.analytics.state.StateSideEffects
+import com.klaviyo.core.Operation
 import com.klaviyo.core.Registry
 import com.klaviyo.core.config.Config
 import com.klaviyo.core.config.LifecycleException
 import com.klaviyo.core.safeApply
 import com.klaviyo.core.safeCall
+import java.util.*
 
 /**
  * Public API for the core Klaviyo SDK.
@@ -25,6 +27,11 @@ import com.klaviyo.core.safeCall
  * to be processed and sent to the Klaviyo backend
  */
 object Klaviyo {
+
+    /**
+     * Queue of failed operations attempted prior to [initialize]
+     */
+    private val preInitQueue: Queue<Operation<Unit>> = LinkedList()
 
     init {
         /**
@@ -63,6 +70,10 @@ object Klaviyo {
         Registry.register<StateSideEffects>(StateSideEffects())
 
         Registry.get<State>().apiKey = apiKey
+
+        while (preInitQueue.isNotEmpty()) {
+            preInitQueue.poll()?.let { it() }
+        }
     }
 
     /**
@@ -207,7 +218,7 @@ object Klaviyo {
      * @param event A map-like object representing the event attributes
      * @return Returns [Klaviyo] for call chaining
      */
-    fun createEvent(event: Event): Klaviyo = safeApply {
+    fun createEvent(event: Event): Klaviyo = safeApply(preInitQueue) {
         Registry.get<ApiClient>().enqueueEvent(event, Registry.get<State>().getAsProfile())
     }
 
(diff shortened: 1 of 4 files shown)


sdk/core/src/main/java/com/klaviyo/core/KlaviyoException.kt
```diff
@@ -18,16 +22,20 @@ abstract class KlaviyoException(final override val message: String) : Exception(
...
-        // KlaviyoException is self-logging
-        null
-    }
+fun <ReturnType> safeCall(
+    errorQueue: Queue<Operation<ReturnType>>? = null,
+    block: Operation<ReturnType>
+): ReturnType? = try {
+    block()
+} catch (e: KlaviyoException) {
+    // KlaviyoException is self-logging
```
ndurell, line 31: What does self logging mean?

evan-masseau: `KlaviyoException` logs an error message from its initializer... considering this is like the 3rd time someone's asked me that I feel like I should change that though if its causing (minor) confusion

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -207,7 +219,7 @@ object Klaviyo {
      * @param event A map-like object representing the event attributes
      * @return Returns [Klaviyo] for call chaining
      */
-    fun createEvent(event: Event): Klaviyo = safeApply {
+    fun createEvent(event: Event): Klaviyo = safeApply(preInitQueue) {
```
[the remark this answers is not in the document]
evan-masseau: Yeah, this is just to ensure we get opened push events even if there's a timing issue with initializing. It does apply to any type of event though

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -63,6 +71,10 @@ object Klaviyo {
         Registry.register<StateSideEffects>(StateSideEffects())
 
         Registry.get<State>().apiKey = apiKey
+
+        while (preInitQueue.isNotEmpty()) {
+            preInitQueue.poll()?.let { it() }
```
evan-masseau, line 76: TODO
- The operations in the queue are unprotected, so they should be try-catched here. 
- Log a message about retrying here.

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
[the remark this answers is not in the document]
evan-masseau: TODO: only apply this buffer to opened push events, not all events.

sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoUninitializedTest.kt
```diff
@@ -118,4 +118,10 @@ internal class KlaviyoUninitializedTest {
         Klaviyo.handlePush(KlaviyoTest.mockIntent(KlaviyoTest.stubIntentExtras))
         assertCaught<MissingConfig>()
     }
+
+    @Test
+    fun `HandlePushToken gets reinvoked once initialized`() {
```
evan-masseau, line 123: this shouldn't even be here

sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoUninitializedTest.kt
```diff
@@ -118,4 +118,10 @@ internal class KlaviyoUninitializedTest {
...
+        Klaviyo.handlePush(KlaviyoTest.mockIntent(KlaviyoTest.stubIntentExtras))
+        assertCaught<MissingConfig>()
+    }
```
evan-masseau, line 126:
```suggestion
```

evan-masseau:
Closing for now, will need to revisit this code once we've put out the patch build that is adding similar, but stripped down changes.

PR #174, opened 2024-05-30
In-memory buffer to allow opened push events to be handled prior to initialize

evan-masseau:
# Description

Adds a buffer for `handlePush` calls invoked prior to initializing, and replays them during init. 

I condensed the critical parts of #172 and #173 into one PR, branched off master instead, in the interest of getting a hotfix out earlier. 

# Check List

## Changelog / Code Overview

- Moved init code from the api client object, to avoid unrecoverable initializer  error
- For `handlePush` only, save operations in memory prior to initializing and replay the queue from `initialize` (though it is a generic solution if we need to extend in future) 
- Added/fixed unit tests

## Test Plan

- [x] Unit tests
- [ ] Test against RN test app.

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
@@ -12,12 +12,15 @@ import com.klaviyo.analytics.model.Profile
 import com.klaviyo.analytics.model.ProfileKey
 import com.klaviyo.analytics.networking.ApiClient
 import com.klaviyo.analytics.networking.KlaviyoApiClient
+import com.klaviyo.core.Operation
 import com.klaviyo.core.Registry
 import com.klaviyo.core.config.Config
 import com.klaviyo.core.config.LifecycleException
 import com.klaviyo.core.lifecycle.NoOpLifecycleCallbacks
 import com.klaviyo.core.safeApply
 import com.klaviyo.core.safeCall
+import java.util.LinkedList
+import java.util.Queue
 
 /**
  * Public API for the core Klaviyo SDK.
@@ -25,6 +28,11 @@ import com.klaviyo.core.safeCall
  * to be processed and sent to the Klaviyo backend
  */
 object Klaviyo {
+
+    /**
+     * Queue of failed operations attempted prior to [initialize]
+     */
+    private val preInitQueue: Queue<Operation<Unit>> = LinkedList()
 
     @Deprecated(
         """
@@ -38,7 +46,12 @@ object Klaviyo {
     private val profileOperationQueue = ProfileOperationQueue()
 
     init {
-        // Since analytics platform owns ApiClient, we must register the service on initialize
+        /**
+         * Since analytics module owns ApiClient, we must register it.
+         *
+         * This registration is a lambda invoked when the API service is required.
+         * KlaviyoApiClient service is not being initialized here.
+         */
         if (!Registry.isRegistered<ApiClient>()) Registry.register<ApiClient> { KlaviyoApiClient }
     }
 
@@ -64,6 +77,21 @@ object Klaviyo {
         } ?: throw LifecycleException()
 
         UserInfo.startObservers()
+
+        Registry.get<ApiClient>().startService()
+
+        if (preInitQueue.isNotEmpty()) {
+            val operationCount = preInitQueue.count()
+            val operation = if (operationCount == 1) "operation" else "operations"
+
+            Registry.log.info(
+                "Replaying $operationCount $operation that were invoked prior to Klaviyo initialization."
+            )
+
+            while (preInitQueue.isNotEmpty()) {
+                preInitQueue.poll()?.let { safeCall(null, it) }
+            }
+        }
     }
 
     /**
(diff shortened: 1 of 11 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -26,6 +29,11 @@ import com.klaviyo.core.safeCall
  */
 object Klaviyo {
 
+    /**
+     * Queue of failed operations attempted prior to [initialize]
+     */
+    private val preInitQueue: Queue<Operation<Unit>> = LinkedList()
```
ajaysubra, line 35: I may be misunderstanding here but are we using this both as a pre init queue as well as an error queue? I saw that when this is passed to exception handling it was called error queue? 

I ask mainly because should this name reflect that and also just trying to understand what exceptions would cause requests added to this queue?

evan-masseau: Yeah it's a little convoluted but the exceptions we catch with safeCall/safeApply are due to not initializing, so the queue gets added to when handlepush is called pre-initialize. 
This was originally a more generic solution that's why I did it that way

sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoUninitializedTest.kt
```diff
@@ -3,39 +3,38 @@ package com.klaviyo.analytics
...
-            every { error(any(), any<Throwable>()) } answers {
-                println(firstArg<String>())
-                secondArg<Throwable>().printStackTrace()
-            }
-        }
 
-        private val mockApiClient = mockk<ApiClient>()
+    private val logger = spyk(LogFixture()).apply {
+        every { error(any(), any<Throwable>()) } answers {
+            println(firstArg<String>())
```
kennyklaviyo, line 24: debug code?

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -64,6 +77,21 @@ object Klaviyo {
         } ?: throw LifecycleException()
 
         UserInfo.startObservers()
+
+        Registry.get<ApiClient>().startService()
+
+        if (preInitQueue.isNotEmpty()) {
+            val operationCount = preInitQueue.count()
+            val operation = if (operationCount == 1) "operation" else "operations"
+
```
kennyklaviyo, line 86: since this is just for logging, no need to be fancy with the message, can simply do `operation(s)` and skip the singular check

evan-masseau: fiiiine

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -319,9 +347,10 @@ object Klaviyo {
             }
         }
 
-        getPushToken()?.let { event[EventKey.PUSH_TOKEN] = it }
+        UserInfo.pushToken.let { event[EventKey.PUSH_TOKEN] = it }
 
-        createEvent(event)
+        Registry.log.verbose("Enqueuing ${event.metric.name} event")
+        Registry.get<ApiClient>().enqueueEvent(event, UserInfo.getAsProfile())
```
kennyklaviyo, line 353: why not keep `createEvent` here?

evan-masseau: If i used`createEvent` or `getPushToken` in this method, those are wrapped in `safeCall` and when they hit an exception, it gets swallowed instead of bubbling up to handlePush's `safeApply`

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -33,83 +33,37 @@ internal object KlaviyoApiClient : ApiClient {
...
-        // Stop the background batching job while offline
-        Registry.networkMonitor.onNetworkChange { isOnline ->
-            if (isOnline) {
-                startBatch(true)
-            } else {
-                stopBatch()
-            }
+        if (getQueueSize() == 0) {
+            // We only need to restore queue from persistent store once
+            restoreQueue()
```
kennyklaviyo, line 48: there isnt a case where we would have something in the queue but still want to restore from persistence? ie, conditionalizing on whether we've restored from persistence already isn't better/more logically correct?

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
kennyklaviyo: ahh ok i see that now

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
evan-masseau: We only want to restore queue once, because after that we always keep the in-memory and persistent queue in sync. Previously we could be sure we only restored queue once, and it always happened first, because it was done from `init`. 
I think this is what you're saying too, but I'm thinking now that it'd be better/clearer here I used a single bool flag or lateinit var to make sure that it only happens once.

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
evan-masseau: So basically, yes it is an error queue, but just for pre-init errors.

kennyklaviyo (approved):

ajaysubra (approved):

kennyklaviyo (approved):

PR #176, opened 2024-06-05
Version 2.4.0 Minor Release

evan-masseau:
2.4.0 will contain two improvements related to opened push event handling
- Opened Push events will trigger queued API requests to send immediately -- should improve timeliness of reporting
- If Opened Push is triggered prior to `initialize`, we'll hold that request in memory until initialized. Allowing for this special case because of the rigid timing for handling the `intent`. All other SDK methods still require initialization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,8 +298,8 @@ override fun onNewIntent(intent: Intent?) {
 
 **Note:** Intent handling may differ depending on your app's architecture. Typically, the intent will be received in 
 your main `Activity`. If you have multiple activities, you need to call `handlePush` from any activity that could
-be the receiver of this intent. Adjust this example to your use-case, ensuring that `Klaviyo.handlePush(intent)` 
-is called when your app is opened from a notification.
+receive this intent. Adjust this example to your use-case, ensuring that `Klaviyo.handlePush(intent)` is called 
+whenever your app is opened from a notification.
 
 #### Deep Linking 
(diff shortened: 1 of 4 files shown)


jordan-griffin (approved):

versions.properties
```diff
@@ -9,11 +9,13 @@
 ####
 #### NOTE: Some versions are filtered by the rejectVersionIf predicate. See the settings.gradle.kts file.
 
+version.androidx.compose.compiler=1.5.10
+
```
evan-masseau, line 13: noope
```suggestion
```

README.md
```diff
@@ -294,8 +296,10 @@ override fun onNewIntent(intent: Intent?) {
 }
 ```
 
-**Note:** Intent handling may differ depending on your app's architecture. Adjust this example to your use-case, 
-ensuring that `Klaviyo.handlePush(intent)` is called when your app is opened from a notification.
+**Note:** Intent handling may differ depending on your app's architecture. Typically, the intent will be received in 
+your main `Activity`. If you have multiple activities, you need to call `handlePush` from any activity that could
+receive this intent. Adjust this example to your use-case, ensuring that `Klaviyo.handlePush(intent)` is called
```
evan-masseau, line 301: Our default push handler uses the default launcher intent, so that should always be the main activity. But I thought we should call this out in case of someone constructing their own notification intents.

PR #177, opened 2024-06-13
Introduce a de-dupe mechanism for push notifications

evan-masseau:
# Description
Similar to `apns-collapse-id` this would allow us to collapse duplicate push notifications on the device itself. In general, we assumed in the past that we'd manage de-duping this purely from the backend, but this would give us extra options, never a bad thing!

# Check List

- [x] Are your changes backwards compatible with previous SDK Versions? YES - tag is a nullable arg
- [ ] Have you tested this change on real device? NOPE
- [ ] Have you added unit test coverage for your changes? NOPE
- [x] Have you verified that your changes are compatible with all Android versions the SDK currently supports? SHOULD!

## Changelog / Code Overview

- Add extension to parse `"notification_tag"` out of the remote data message payload 
- Use that as the tag for notifications (just like stock FCM)

## Test Plan

- [ ] Test this with the sample / test app. Can easily do it with a dummy payload, don't need to update server code to be able to test. 

## Related Issues/Tickets

CHNL-9595

### DIFF
--- a/sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
+++ b/sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
@@ -29,6 +29,7 @@ import com.klaviyo.pushFcm.KlaviyoRemoteMessage.imageUrl
 import com.klaviyo.pushFcm.KlaviyoRemoteMessage.isKlaviyoNotification
 import com.klaviyo.pushFcm.KlaviyoRemoteMessage.notificationCount
 import com.klaviyo.pushFcm.KlaviyoRemoteMessage.notificationPriority
+import com.klaviyo.pushFcm.KlaviyoRemoteMessage.notificationTag
 import com.klaviyo.pushFcm.KlaviyoRemoteMessage.sound
 import com.klaviyo.pushFcm.KlaviyoRemoteMessage.title
 import java.net.URL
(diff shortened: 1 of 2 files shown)


jordan-griffin (approved):

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
```diff
@@ -104,7 +106,7 @@ class KlaviyoNotification(private val message: RemoteMessage) {
 
         NotificationManagerCompat
             .from(context)
-            .notify(generateId(), notification.build())
+            .notify(message.notificationTag, generateId(), notification.build())
```
evan-masseau, line 109: Needs fixing: `generateId()` returns the time in milliseconds. `notify` dedupes based on tag _and_ id, so because ID is already (pretty much) unique, this will never dedupe. Instead we should just use a constant value for ID (like 0), and just rely on tag. Caveat -- we'd need to autogenerate a tag (timestamp should suffice) if `notificationTag` is null.

evan-masseau: BTW I recall this worked in my original testing, but I think thats because I was sending the notifications near simultaneously from my local, so the device probably did `generateId` twice within a millisecond!

dan-peluso (approved):
might be slightly cheating since I added code but might as well approve

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
```diff
@@ -104,7 +106,7 @@ class KlaviyoNotification(private val message: RemoteMessage) {
 
         NotificationManagerCompat
             .from(context)
-            .notify(generateId(), notification.build())
+            .notify(message.notificationTag ?: generateId().toString(), 0, notification.build())
```
evan-masseau, line 109: 👍 can't approve my own PR, but approved lol

PR #179, opened 2024-08-19
setProfileAttribute should accept any Serializable value not just String

dan-peluso:
# Description
Changing the signature of the setAttribute to accept any Serializable

I currently having this take the form of 'attempt to cast to a String, if that fails add a null value' but I can see an argument against that. Do we want to support being able to add non-string types for our main profile keys (email, external id, phone number, etc) or should we throw an error here if that fails? Looking for input.

# Check List

## Changelog / Code Overview
- updating the `setProfileAttribute` to take in any serializable rather than only String
- updating `setAttribute` on the State interface to accept a Serializable (nullable type cast to a string)

## Test Plan
Updated unit tests

## Related Issues/Tickets
CHNL-7939

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
@@ -111,10 +111,26 @@ internal class KlaviyoState : State {
      * Set an individual property or attribute
      */
     override fun setAttribute(key: ProfileKey, value: Serializable) = when (key) {
-        EMAIL -> email = (value as? String)
-        EXTERNAL_ID -> externalId = (value as? String)
-        PHONE_NUMBER -> phoneNumber = (value as? String)
+        EMAIL -> (value as? String)?.let { email = it } ?: run { logCastError(EMAIL, value) }
+        EXTERNAL_ID -> (value as? String)?.let { externalId = it } ?: run {
+            logCastError(
+                EXTERNAL_ID,
+                value
+            )
+        }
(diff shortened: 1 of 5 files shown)


ajaysubra (approved):
This approach makes sense to me generally. Not sure if the android folks have anything else to add.

jordan-griffin (commented):
IMO you've made the right call on what to do with attributes if they are given as non-string types. Maaaybe phone number could be an int depending on if our APIs with append the preceeding + to the number or not. But overall I dont think we'd want anything but strings on these primary identifiers.
This way we open up the options for serializable values on our other potential profile keys but leave identifiers clean.

jordan-griffin (approved):

kennyklaviyo (approved):

sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
```diff
@@ -109,13 +110,29 @@ internal class KlaviyoState : State {
...
-        PHONE_NUMBER -> phoneNumber = value
+    override fun setAttribute(key: ProfileKey, value: Serializable) = when (key) {
+        EMAIL -> (value as? String)?.let { email = it } ?: run { logCastError(EMAIL, value) }
+        EXTERNAL_ID -> (value as? String)?.let { externalId = it } ?: run {
+            logCastError(
+                EXTERNAL_ID,
+                value
+            )
+        }
+        PHONE_NUMBER -> (value as? String)?.let { phoneNumber = it } ?: run {
```
evan-masseau, line 121: Man there's always more to learn about Kotlin, how did i not know about `run`

evan-masseau (approved):
You might have a record # of approvals on this lol

PR #180, opened 2024-08-20
Concurrent network observer fix

dan-peluso:
# Description
The observers in `KlaviyoNetworkMonitor` were not stored in a thread-safe structure, so when broadcasting a change we sometimes get a `ConcurrentModificationException` if the list is being modified.

# Check List

## Changelog / Code Overview
- updating the `networkChangeObservers` type from a list to a Collections.synchronizedList
- adding a kotlinx coroutine test dependency on the core module

## Test Plan
Created a unit test for this scenario

## Related Issues/Tickets
CHNL-11317

### DIFF
--- a/sdk/analytics/build.gradle
+++ b/sdk/analytics/build.gradle
@@ -19,6 +19,7 @@ android {
 dependencies {
     implementation project(":sdk:core")
     testImplementation project(":sdk:fixtures")
+    testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:1.8.0"
 }
 
 afterEvaluate {
--- a/sdk/analytics/build.gradle
+++ b/sdk/analytics/build.gradle
@@ -19,7 +19,7 @@ android {
 dependencies {
     implementation project(":sdk:core")
     testImplementation project(":sdk:fixtures")
-    testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:1.8.0"
+    testImplementation KotlinX.coroutines.test
 }
 
 afterEvaluate {
--- a/sdk/core/build.gradle
+++ b/sdk/core/build.gradle
@@ -2,6 +2,8 @@ import static de.fayard.refreshVersions.core.Versions.versionFor
 
 dependencies {
     testImplementation project(":sdk:fixtures")
+    // coroutine testing
+    testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:1.8.0"
 }
 description = "Core featureset of the Klaviyo SDK including SDK configuration, session tracking and analytics"
 evaluationDependsOn(":sdk")
--- a/sdk/core/build.gradle
+++ b/sdk/core/build.gradle
@@ -3,7 +3,7 @@ import static de.fayard.refreshVersions.core.Versions.versionFor
 dependencies {
     testImplementation project(":sdk:fixtures")
     // coroutine testing
-    testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:1.8.0"
+    testImplementation KotlinX.coroutines.test
 }
 description = "Core featureset of the Klaviyo SDK including SDK configuration, session tracking and analytics"
 evaluationDependsOn(":sdk")
--- a/sdk/core/src/main/java/com/klaviyo/core/networking/KlaviyoNetworkMonitor.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/networking/KlaviyoNetworkMonitor.kt
@@ -6,6 +6,7 @@ import android.net.Network
 import android.net.NetworkCapabilities
 import android.net.NetworkRequest
 import com.klaviyo.core.Registry
+import java.util.Collections
 
 /**
  * Service for monitoring the application lifecycle and network connectivity
@@ -23,7 +24,9 @@ internal object KlaviyoNetworkMonitor : NetworkMonitor {
     /**
      * List of registered network change observers
      */
-    private var networkChangeObservers = mutableListOf<NetworkObserver>()
+    private val networkChangeObserversConcurrent = Collections.synchronizedList(
+        mutableListOf<NetworkObserver>()
+    )
 
     /**
      * Callback object to register with system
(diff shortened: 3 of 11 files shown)


sdk/core/src/test/java/com/klaviyo/core/networking/KlaviyoNetworkMonitorTest.kt
```diff
@@ -130,6 +134,7 @@ internal class KlaviyoNetworkMonitorTest : BaseTest() {
         netCallbackSlot.captured.onLost(mockk())
 
         assertEquals(6, callCount)
+        expectedNetworkConnection = true
```
dan-peluso, line 137: The reason for this is pretty interesting - the assertion for this test will still run even **after** the test case finishes, which I believe exposes a pitfall in our `set listener` design here. Without this reset to `true`, my new unit test would fail `assert(it == expectedNetworkConnection)` since we start the test by adding a connection. Changing the expectation to `true` essentially resets this to the same state as the start.

kennyklaviyo (approved):

sdk/core/build.gradle
```diff
@@ -2,6 +2,8 @@ import static de.fayard.refreshVersions.core.Versions.versionFor
 
 dependencies {
     testImplementation project(":sdk:fixtures")
+    // coroutine testing
+    testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:1.8.0"
```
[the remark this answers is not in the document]
evan-masseau: na this is fine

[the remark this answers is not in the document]
evan-masseau: I mean we can update it to the refresh versions syntax, but location is fine. I never remember the syntax, but there's a gradle command that'll do it for you.

sdk/core/src/main/java/com/klaviyo/core/networking/KlaviyoNetworkMonitor.kt
```diff
@@ -23,7 +24,9 @@ internal object KlaviyoNetworkMonitor : NetworkMonitor {
     /**
      * List of registered network change observers
      */
-    private var networkChangeObservers = mutableListOf<NetworkObserver>()
+    private val networkChangeObserversConcurrent = Collections.synchronizedList(
```
evan-masseau, line 27: cool, figured there'd be some data structure for this.

sdk/core/src/test/java/com/klaviyo/core/networking/KlaviyoNetworkMonitorTest.kt
evan-masseau: If I follow that correctly, this basically serves the purpose of resetting the test class's expectations prior to the next test? Would it work to put this in a setup or teardown function instead?

sdk/core/src/test/java/com/klaviyo/core/networking/KlaviyoNetworkMonitorTest.kt
```diff
@@ -167,4 +172,25 @@ internal class KlaviyoNetworkMonitorTest : BaseTest() {
         netCallbackSlot.captured.onAvailable(mockk())
         assertEquals(5, callCount)
     }
+
+    @Test()
+    fun `Concurrent modification exception doesn't get thrown on concurrent observer access`() = runTest {
```
evan-masseau, line 177: Its very cool you were able to repro this, considering how unreliable concurrency issues can be.

evan-masseau (changes requested):
I approve of the approach, however I'm going to "request changes" because there are some other observer lists in the codebase I'd like you to give the same `synchronizedList` treatment juuust to be safe.
- com.klaviyo.analytics.networking.KlaviyoApiClient#apiObservers
- com.klaviyo.analytics.state.KlaviyoState#stateObservers
- com.klaviyo.core.model.SharedPreferencesDataStore#storeObservers
- com.klaviyo.core.lifecycle.KlaviyoLifecycleMonitor#activityObservers

FYI Some of these are really only used by our test app. And i don't think you need to write a unit test case for all of them either.

sdk/core/src/test/java/com/klaviyo/core/networking/KlaviyoNetworkMonitorTest.kt
dan-peluso: I ended up changing this to instead remove the observer at the end of the test - it is a bit tricky since we declare the count in the test locally but modify the singleton network monitor. I think it makes sense that at the end of the individual test we remove the observer.

dan-peluso:
> I approve of the approach, however I'm going to "request changes" because there are some other observer lists in the codebase I'd like you to give the same `synchronizedList` treatment juuust to be safe.
> 
> * com.klaviyo.analytics.networking.KlaviyoApiClient#apiObservers
> * com.klaviyo.analytics.state.KlaviyoState#stateObservers
> * com.klaviyo.core.model.SharedPreferencesDataStore#storeObservers
> * com.klaviyo.core.lifecycle.KlaviyoLifecycleMonitor#activityObservers
> 
> FYI Some of these are really only used by our test app. And i don't think you need to write a unit test case for all of them either.

reflected in the most recent commit

sdk/analytics/build.gradle
```diff
@@ -19,6 +19,7 @@ android {
 dependencies {
     implementation project(":sdk:core")
     testImplementation project(":sdk:fixtures")
+    testImplementation KotlinX.coroutines.test
```
evan-masseau, line 22: @dan-peluso I had to run `./gradlew refreshVersionsMigrate --mode=VersionsPropertiesOnly` to figure out the right syntax, figured I'd just push a commit once i'd figured it out though.

sdk/core/src/main/java/com/klaviyo/core/networking/KlaviyoNetworkMonitor.kt
```diff
@@ -23,7 +24,9 @@ internal object KlaviyoNetworkMonitor : NetworkMonitor {
     /**
      * List of registered network change observers
      */
-    private var networkChangeObservers = mutableListOf<NetworkObserver>()
```
evan-masseau, line 26: One more unbelievably nitpicky detail, but lets just keep the name `networkChangeObservers` without `Concurrent` at the end, like you did with the other 4.

evan-masseau (approved):

PR #181, opened 2024-08-23
Retry on 503

dan-peluso:
# Description
We are now returning 503 from push-tokens endpoint. These can be retried. We should add support for this within the Android SDK.

Since our SDK does not handle this error response accordingly, we’re potentially throwing out retriable requests and losing data. In case of a webserver unavailability this would prevent customers from losing out on data from their mobile apps.

# Check List

## Changelog / Code Overview
- sending 503 error codes through the retry flow
- we already have logic for the exponential backoff in the case of no `retry-after` header
- creating a unit test for a 503 response with no header being retried

## Test Plan
We should test this by mocking a 503 response and checking that we retry with exponential backoff

## Related Issues/Tickets
CHNL-8144

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
@@ -6,6 +6,7 @@ import java.io.BufferedReader
 import java.io.IOException
 import java.io.InputStreamReader
 import java.net.HttpURLConnection
+import java.net.HttpURLConnection.HTTP_UNAVAILABLE
 import java.net.URL
 import java.util.UUID
 import javax.net.ssl.HttpsURLConnection
@@ -62,6 +63,7 @@ internal open class KlaviyoApiRequest(
         const val HTTP_ACCEPTED = HttpURLConnection.HTTP_ACCEPTED
         const val HTTP_MULT_CHOICE = HttpURLConnection.HTTP_MULT_CHOICE
         const val HTTP_RETRY = 429 // oddly not a const in HttpURLConnection
+        const val HTTP_UNAVAILABLE = HttpURLConnection.HTTP_UNAVAILABLE
 
         // JSON keys for persistence
         const val TYPE_JSON_KEY = "request_type"
@@ -343,7 +345,7 @@ internal open class KlaviyoApiRequest(
 
         status = when (responseCode) {
             in successCodes -> Status.Complete
-            HTTP_RETRY -> {
+            HTTP_RETRY, HTTP_UNAVAILABLE -> {
                 if (attempts < Registry.config.networkMaxAttempts) {
                     Status.PendingRetry
                 } else {
@@ -351,7 +353,7 @@ internal open class KlaviyoApiRequest(
                 }
             }
             // TODO - Special handling of unauthorized i.e. 401 and 403?
-            // TODO - Special handling of server errors 500 and 503?
+            // TODO - Special handling of server error 500?
             else -> Status.Failed
         }
 
--- a/sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
+++ b/sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
@@ -113,7 +113,8 @@ internal class KlaviyoApiClientTest : BaseTest() {
 
     private fun mockRequest(
         uuid: String = "uuid",
-        status: KlaviyoApiRequest.Status = KlaviyoApiRequest.Status.Complete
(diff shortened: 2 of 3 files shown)


versions.properties
```diff
@@ -27,6 +27,10 @@ version.android.buildTools=34.0.0
 
 # Project dependencies
 
+version.kotlinx.coroutines=1.8.1
+##             # available=1.9.0-RC
+##             # available=1.9.0-RC.2
+
```
kennyklaviyo, line 33: doesnt seem like these changes are related?

ajaysubra (approved):
Neat. Looks like android handles not having retry after header and falls back to exponential back off.

versions.properties
dan-peluso: I don't think they are but I'm a bit unfamiliar with this tool - do we just manually remove these update lines?

evan-masseau: Hmm not sure where this woulda come from. Could this have bled over from your other branch about concurrent modification? Because this PR doesn't seem to have anything to do with coroutines

kennyklaviyo: these are ok to have but shouldn't be bundled with these changes to 503s since they're unrelated, there should simply be another PR that'll introduce coroutines to the versions file

dan-peluso: Got it - I reverted that change

kennyklaviyo (approved):

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -6,6 +6,7 @@ import java.io.BufferedReader
 import java.io.IOException
 import java.io.InputStreamReader
 import java.net.HttpURLConnection
+import java.net.HttpURLConnection.HTTP_UNAVAILABLE
```
kennyklaviyo, line 9: is this actually used? we already import `HttpURLConnection` and we reference `HTTP_UNAVAILABLE` from that import

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -343,15 +345,15 @@ internal open class KlaviyoApiRequest(
 
         status = when (responseCode) {
             in successCodes -> Status.Complete
-            HTTP_RETRY -> {
+            HTTP_RETRY, HTTP_UNAVAILABLE -> {
```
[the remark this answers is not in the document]
kennyklaviyo: but we want to reference the const val we declare above

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
kennyklaviyo: i think this can be removed

dan-peluso: Yeah I can get rid of it

kennyklaviyo (approved):

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
```diff
@@ -507,6 +508,60 @@ internal class KlaviyoApiClientTest : BaseTest() {
...
+    fun `503 results in a retry using exponential backoff`() {
+        // same as above test but with a 503 instead of a 429
+        // First unsent request, which we will retry till max attempts (calculated by exponential)
+        // note that we are not sending the retry-after header on a 503
+        val request1 = mockRequest("uuid-retry", KlaviyoApiRequest.Status.Unsent, 503)
+
+        every { request1.state } answers {
+            when (request1.attempts) {
+                0 -> KlaviyoApiRequest.Status.Unsent.name
+                20 -> KlaviyoApiRequest.Status.Failed.name
```
evan-masseau, line 521: Where does 20 come from?

evan-masseau (approved):
Looks good, and appreciate for the well-commented test. Only question was where 20 retries is defined, I thought the # of retries defaulted to 50 (not that it has to be 50, just want to understand).

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
dan-peluso: Was just testing some arbitrary values, figured this wasn't explicitly testing the limits so having it attempt less than 50 was fine. If you want I can revert this tho

evan-masseau: Ooh sorry i read this wrong, this is the mocked response. Yeah this is fine!

Issue #182, opened 2024-08-28
The initial sdk crashed. Procedure


HuangBiaoHb:
### Checklist

- [X] I have read the [contributing guidelines](https://github.com/klaviyo/klaviyo-android-sdk/blob/master/.github/CONTRIBUTING.md)
- [ ] I have determined whether this bug is also reproducible in a vanilla Android project
- [ ] If possible, I've reproduced the issue using the `master` branch or latest release of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/klaviyo/klaviyo-android-sdk/issues) or [pull request](https://github.com/klaviyo/klaviyo-android-sdk/pulls).

### Description

```markdown
The Klaviyo.initialize method crashes, Should I specify the kotlin version?

My version of kotlin: 1.7.20
My version of gradel: 8.3
```

### Expected behavior

```markdown
Can initialize normally
```

### Actual behavior

```markdown
crashes
 java.lang.ClassNotFoundException: Didn't find class "kotlin.enums.EnumEntriesKt" on path: DexPathList[[dex file "/data/data/com.renpho.health/code_cache/.overlay/base.apk/classes23.dex", zip file "/data/app/~~6Up3Lg5KenJ-kNQBvf88oQ==/com.renpho.health-2tg64saQQ6rPFm49PH2jdA==/base.apk"],nativeLibraryDirectories=[/data/app/~~6Up3Lg5KenJ-kNQBvf88oQ==/com.renpho.health-2tg64saQQ6rPFm49PH2jdA==/lib/arm64, /data/app/~~6Up3Lg5KenJ-kNQBvf88oQ==/com.renpho.health-2tg64saQQ6rPFm49PH2jdA==/base.apk!/lib/arm64-v8a, /system/lib64, /system/system_ext/lib64]]
                                                                                                    │ 	at dalvik.system.BaseDexClassLoader.findClass(BaseDexClassLoader.java:259)
                                                                                                    │ 	at java.lang.ClassLoader.loadClass(ClassLoader.java:379)
                                                                                                    │ 	at java.lang.ClassLoader.loadClass(ClassLoader.java:312)
                                                                                                    │  Caused by: java.lang.NoClassD
                                                                                                    └────────────────────────────────────────────────────────────────────────────────────────────────────────────────
2024-08-28 16:21:17.620 25398-25398 CrashHandler            com.renpho.health                    E   
[...]

PR #183, opened 2024-09-03
Updating readme with min kotlin version

dan-peluso:
# Description
We should specify a min verision that correlates to our SDK kotlin version. I tested it with Kotlin 1.9.0 (note this is a previous version to our internal 1.9.22) and it worked with the SDK.

# Check List

## Changelog / Code Overview
- updating documentation

## Test Plan
 - tested with a new app running on Kotlin 1.9.0 and was able to send a push / event / instantiate
 - Evan tested with a new app running Kotlin 1.8.0 and was able to send a push / event / instantiate

## Related Issues/Tickets
https://github.com/klaviyo/klaviyo-android-sdk/issues/182

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,11 @@ send them timely push notifications via [FCM (Firebase Cloud Messaging)](https:/
 > ⚠️ **We support Android API level 23 and above** ⚠️
 
 ## Installation
+
+### Requirements
+
+- Kotlin **1.9.0** or later
+
 1. Include the [JitPack](https://jitpack.io/#klaviyo/klaviyo-android-sdk) repository in your project's build file
    <details>
       <summary>Kotlin DSL</summary>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,6 +17,11 @@ send them timely push notifications via [FCM (Firebase Cloud Messaging)](https:/
 > ⚠️ **We support Android API level 23 and above** ⚠️
 
 ## Installation
+
+### Requirements
+
+- Kotlin **1.9.0** or later
```
ajaysubra, line 23: 🔍(nit pick)  Should this be like a info bar instead? Now it seems like the below section which formerly was installation is now requirements.

dan-peluso: I kinda felt like this would be a sub-category for installation, but I can make it into a new section if you prefer. Where would you like to see it?

kennyklaviyo: theres an info block above this actually that about the Android API version, maybe we can join those together? probably also make the wording for the Android API match this

ajaysubra: Sorry I wasn't clear. Kenny is right, that is what I meant.

kennyklaviyo (approved):

evan-masseau (approved):

PR #184, opened 2024-09-03
CHNL-6996 Proguard docs and consumer rules

dan-peluso:
# Description
https://github.com/klaviyo/klaviyo-react-native-sdk/issues/138

Related to the above issue - basically we have the ability to add some rules in our SDK to safeguard obfuscation. Adding these rules and additional documentation for troubleshooting.

# Check List

## Changelog / Code Overview
- adding consumer proguard rule for modules that depend on each other

## Test Plan
- test with RN release build

## Related Issues/Tickets
CHNL-6996

### DIFF
--- a/README.md
+++ b/README.md
@@ -483,6 +483,17 @@ the following metadata tag to your manifest file.
 </manifest>
 ```
 
+#### Proguard / R8 Issues
+
+If you notice issues in the release build of your apps, you can try to manually add a couple rules
+to your `proguard-rules.pro` to prevent obfuscation:
+```
+-keep class com.klaviyo.analytics.** { *; }
+-keep class com.klaviyo.core.** { *; }
+-keep class com.klaviyo.push-fcm.** { *; }
+```
+
+
 ## Contributing
 See the [contributing guide](.github/CONTRIBUTING.md) to learn how to contribute to the Klaviyo Android SDK.
 We welcome your feedback in the [issues](https://github.com/klaviyo/klaviyo-android-sdk/issues) section of our public GitHub repository.
--- a/sdk/analytics/consumer-rules.pro
+++ b/sdk/analytics/consumer-rules.pro
@@ -1,0 +1,1 @@
+-keep class com.klaviyo.core.** { *; }
--- a/sdk/build.gradle
+++ b/sdk/build.gradle
@@ -25,6 +25,7 @@ subprojects {
             testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
             versionCode versionFor(project, "version.klaviyo.versionCode") as Integer
             versionName versionFor(project, "version.klaviyo.versionName") as String
+            consumerProguardFiles("consumer-rules.pro")
         }
 
         buildFeatures {
--- a/sdk/push-fcm/consumer-rules.pro
+++ b/sdk/push-fcm/consumer-rules.pro
@@ -1,0 +1,2 @@
+-keep class com.klaviyo.core.** { *; }
+-keep class com.klaviyo.analytics.** { *; }


sdk/analytics/consumer-rules.pro
```diff
@@ -0,0 +1 @@
+-keep class com.klaviyo.core.** { *; }
```
kennyklaviyo, line 1: should this be:
```
-keep class com.klaviyo.analytics.** { *; }
```

sdk/push-fcm/consumer-rules.pro
```diff
@@ -0,0 +1,2 @@
+-keep class com.klaviyo.core.** { *; }
+-keep class com.klaviyo.analytics.** { *; }
```
kennyklaviyo, line 2: and this is:
```
-keep class com.klaviyo.push-fcm.** { *; }
```

README.md
```diff
@@ -483,6 +483,17 @@ the following metadata tag to your manifest file.
 </manifest>
 ```
 
+#### Proguard / R8 Issues
+
+If you notice issues in the release build of your apps, you can try to manually add a couple rules
+to your `proguard-rules.pro` to prevent obfuscation:
+```
+-keep class com.klaviyo.analytics.** { *; }
+-keep class com.klaviyo.core.** { *; }
```
kennyklaviyo, line 492: and you need to also add this to core?

sdk/build.gradle
```diff
@@ -25,6 +25,7 @@ subprojects {
             testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
             versionCode versionFor(project, "version.klaviyo.versionCode") as Integer
             versionName versionFor(project, "version.klaviyo.versionName") as String
+            consumerProguardFiles("consumer-rules.pro")
```
kennyklaviyo, line 28: and each module with consumer-rules would need this line?

sdk/analytics/consumer-rules.pro
dan-peluso: I was under the impression that you need to declare the modules you _depend_ on, not the module that you are writing the proguard rule in. So if we wanted to protect against obfuscating the core library when a customer is using the analytics module, we need to basically keep any of our dependencies. I could be wrong about that though.
https://medium.com/android-news/proguard-r8-in-the-world-of-modularity-f599650b4553

README.md
dan-peluso: Since core has no dependencies (not including tests) I think we are okay here (since there's nothing it could accidentally obfuscate)

sdk/build.gradle
dan-peluso: I believe we only need to do this for the root level build.gradle, it does need to be under the `defaultConfig`

kennyklaviyo (approved):

dan-peluso:
Verified the fix using this branch on a version of the react native SDK that referenced this Android SDK:
| before | after |
| -- | -- |
| image | image |

README.md
evan-masseau: I'm like 99% that they shouldn't need `-keep class com.klaviyo.core.** { *; }` because they shouldn't have any reference to core in their code. I guess if they're troubleshooting, maybe better to be safe?

dan-peluso: Yeah I think just because we do expose it this wouldn't be the worst idea, but certainly can remove it if you'd like

sdk/analytics/consumer-rules.pro
evan-masseau: I was going to say the same thing as Kenny. But then I reread that article. And I think Dan has it right -- and this finally clarifies for me why this only happened to apps dependent upon the react native SDK, and I couldn't reproduce it in a straight Android app that depended on the android SDK modules directly.

In the case of react native, using those event constants represents a transitive dependency, and since we don't explicitly reference the event constants in the react native SDK code, the shrinker thinks it can remove them. We reference those constants dynamically via reflection so the shrinker doesn't see it. So they are missing at run time.

So I think it is by virtue of the fact that your proguard consumer rule in the push-fcm module instruct the consumer to keep all analytics classes that the react native test app winds up knowing to keep the event constants when shrunk. 

In that sense, for robustness these proguard rules technically belong in the react native SDK repo too, though it'd be sorta duplicative as long as the RN SDK is importing push-fcm.

README.md
evan-masseau: Having re-read the article, I feel pretty confident no one should ever need any of these three rules based if they are consuming the Android SDK directly.
BUT this is the troubleshooting section -- so its good to cover our bases. So I say keep all three like you have it!

sdk/build.gradle
evan-masseau: Yeah I think that's right since this section applies to all the subprojects

sdk/push-fcm/consumer-rules.pro
evan-masseau: See above thread

evan-masseau (approved):
Nice.
Now that I _think_ I understand the issue better, I suggest we also add consumer-proguard rules to the RN SDK with these rules: 
```
-keep class com.klaviyo.analytics.** { *; }
-keep class com.klaviyo.push-fcm.** { *; }
```

dan-peluso:
> consumer-proguard rules

https://klaviyo.atlassian.net/browse/CHNL-12127 made this to track

PR #185, opened 2024-09-10
Exposing klaviyo SDK name and version privately

ajaysubra:
# Description
Exposing SDK name and version so that this can be set from react native. This would aid in better debugging on our end and also know the usage of react native. 

## Test Plan
Tested setting these using the android test app like so,
```kotlin
            val config = Registry.configBuilder
                .apiKey(companyId)
                .sdkName("Klaviyo SDK Test App")
                .sdkVersion("2.0.0")
                .applicationContext(context)
                .baseUrl(it)
                .build()
```

and then looking at the events and making sure that the sdk name and version was what was set above.

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/DeviceProperties.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/DeviceProperties.kt
@@ -50,9 +50,8 @@ internal object DeviceProperties {
     val sdkVersion: String
         get() = KlaviyoConfig.sdkVersion
 
-    val sdkName: String by lazy {
-        KlaviyoConfig.sdkName
-    }
+    val sdkName: String
+        get() = KlaviyoConfig.sdkName
 
     val backgroundDataEnabled: Boolean by lazy {
         !activityManager.isBackgroundRestrictedCompat()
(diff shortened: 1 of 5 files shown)


ndurell (approved):

sdk/analytics/src/main/java/com/klaviyo/analytics/DeviceProperties.kt
```diff
@@ -48,11 +48,11 @@ internal object DeviceProperties {
     }
 
     val sdkVersion: String by lazy {
-        BuildConfig.VERSION
+        KlaviyoConfig.sdkVersion
```
evan-masseau, line 51: might not want these to be lazy anymore, i think if you were to re-init and change the version number or sdk name for some reason, it would not get reflected here.

dan-peluso: ^ if this is a concern you could do something like this:
```suggestion
val sdkVersion: String
        get() = KlaviyoConfig.sdkVersion
```

ajaysubra: Thanks makes sense. Thanks @evan-masseau updated it to be non-lazy. Thanks for the code sample @dan-peluso

evan-masseau (approved):

dan-peluso (approved):

PR #186, opened 2024-09-10
CHNL-3990 Remove identifiers from state if format issues

dan-peluso:
# Description
- if we get a 400 from the backend due to a bad email / phone number, we need to clear that bad value from the local state
- adding some logic to deserialize the error body of a request
- adding a way to clear email and phone number fields to the State (internal to SDK only)
# Check List

## Test Plan
Tested locally and in production endpoints by using an ill-formatted email and phone number

## Related Issues/Tickets
CHNL-3990

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/ErrorResponse.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/ErrorResponse.kt
@@ -1,0 +1,39 @@
+package com.klaviyo.analytics.networking.requests
+
+/**
+ * For parsing Error responses from the backend when an HTTP request fails
+ */
+
+internal data class ErrorResponse(
+    val errors: List<KlaviyoError>
+) {
+    companion object {
+        // Error body constants
+        const val ERRORS = "errors"
+        const val ID = "id"
+        const val STATUS = "status"
+        const val TITLE = "title"
+        const val DETAIL = "detail"
+        const val SOURCE = "source"
+        const val POINTER = "pointer"
+        const val INVALID_INPUT_TITLE = "Invalid input."
+    }
+}
+
+internal data class KlaviyoError(
+    val id: String? = null,
+    val status: Int? = null,
+    val title: String? = null,
+    val detail: String? = null,
+    val source: KlaviyoErrorSource? = null
+)
+
+internal data class KlaviyoErrorSource(
+    val pointer: String? = null
+) {
+    companion object {
+        // current path objects from the backend
+        const val EMAIL_PATH = "/data/attributes/email"
+        const val PHONE_NUMBER_PATH = "/data/attributes/phone_number"
+    }
+}
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/JSONUtil.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/JSONUtil.kt
@@ -4,6 +4,9 @@ import org.json.JSONObject
 
 internal object JSONUtil {
 
+    /**
+     * Using this util since built-in optString gets scared with a null default value
+     */
     fun JSONObject.getStringNullable(key: String): String? =
         if (has(key) && !isNull(key)) getString(key) else null
 }
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/JSONUtil.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/JSONUtil.kt
@@ -1,0 +1,9 @@
+package com.klaviyo.analytics.networking.requests
+
+import org.json.JSONObject
+
+internal object JSONUtil {
+
+    fun JSONObject.getStringNullable(key: String): String? =
+        if (has(key) && !isNull(key)) getString(key) else null
+}
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/state/StateSideEffects.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/state/StateSideEffects.kt
@@ -8,15 +8,13 @@ import com.klaviyo.analytics.model.Profile
 import com.klaviyo.analytics.model.ProfileKey
 import com.klaviyo.analytics.networking.ApiClient
 import com.klaviyo.analytics.networking.requests.ApiRequest
-import com.klaviyo.analytics.networking.requests.ErrorResponse
 import com.klaviyo.analytics.networking.requests.KlaviyoApiRequest
 import com.klaviyo.analytics.networking.requests.KlaviyoApiRequest.Companion.HTTP_BAD_REQUEST
-import com.klaviyo.analytics.networking.requests.KlaviyoErrorResponseDecoder
+import com.klaviyo.analytics.networking.requests.KlaviyoErrorResponse
 import com.klaviyo.analytics.networking.requests.KlaviyoErrorSource
 import com.klaviyo.analytics.networking.requests.PushTokenApiRequest
 import com.klaviyo.core.Registry
 import com.klaviyo.core.config.Clock
-import org.json.JSONObject
 
 internal class StateSideEffects(
     private val state: State = Registry.get<State>(),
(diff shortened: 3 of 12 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/ErrorResponse.kt
```diff
@@ -0,0 +1,39 @@
...
+    companion object {
+        // Error body constants
+        const val ERRORS = "errors"
+        const val ID = "id"
+        const val STATUS = "status"
+        const val TITLE = "title"
+        const val DETAIL = "detail"
+        const val SOURCE = "source"
+        const val POINTER = "pointer"
+        const val INVALID_INPUT_TITLE = "Invalid input."
```
dan-peluso, line 19: I'm using this to determine if we have an input error vs some other type of 400 error, feels a little shady but if we can guarantee this in every response I think it should be okay. I just don't want to look for an email / phone error on every 400 we get

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/JSONUtil.kt
```diff
@@ -0,0 +1,9 @@
+package com.klaviyo.analytics.networking.requests
+
+import org.json.JSONObject
+
+internal object JSONUtil {
+
+    fun JSONObject.getStringNullable(key: String): String? =
```
dan-peluso, line 7: Maybe we already have something like this? We don't do a ton with responses currently so my thought was this could be helpful for dealing with malformed responses

sdk/analytics/src/main/java/com/klaviyo/analytics/state/StateSideEffects.kt
```diff
@@ -110,7 +116,29 @@ internal class StateSideEffects(
...
+            request.responseBody?.let { responseBody ->
+                val error = KlaviyoErrorResponseDecoder.fromJson(JSONObject(responseBody))
+                error.errors.find { it.title == ErrorResponse.INVALID_INPUT_TITLE }?.let { inputError ->
+                    when (inputError.source?.pointer) {
+                        KlaviyoErrorSource.EMAIL_PATH -> {
+                            (Registry.get<State>() as? KlaviyoState)?.resetEmail()
+                            Registry.log.warning("Invalid email - resetting email state to null")
+                        }
+                        KlaviyoErrorSource.PHONE_NUMBER_PATH -> {
+                            (Registry.get<State>() as? KlaviyoState)?.resetPhoneNumber()
```
dan-peluso, line 129: This also feels a bit gross, but I don't want to add anything to the public State interface that would allow users to do this.

sdk/analytics/src/main/java/com/klaviyo/analytics/state/StateSideEffects.kt
```diff
@@ -110,7 +116,29 @@ internal class StateSideEffects(
...
+                            Registry.log.warning(
+                                "Invalid phone number - resetting phone number state to null"
+                            )
+                        }
+                        else -> {
+                            Registry.log.warning("Input error: ${inputError.detail}")
```
dan-peluso, line 135: Another potential debate - do we wan to directly expose backend responses like this to the logs? I actually think here it wouldn't be a bad idea, since we do format it nicely for the other issues and I think it would be helpful for context. If we are cool with this I might just log the inputError.detail for the above cases as well, since they add more details than I did above

sdk/analytics/src/test/java/com/klaviyo/analytics/state/SideEffectTests.kt
```diff
@@ -209,25 +215,102 @@ class SideEffectTests : BaseTest() {
...
+    }
+
     @Test
     fun `Other API failures do not affect push state`() {
         StateSideEffects(stateMock, apiClientMock)
 
         capturedApiObserver.captured(
             mockk<ProfileApiRequest>().apply {
                 every { status } returns KlaviyoApiRequest.Status.Failed
+                every { responseCode } returns 412
```
dan-peluso, line 306: Since we now check for response code in the `afterApiRequest` logic flow, I had to add this

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/JSONUtil.kt
evan-masseau: The only JSON decoding utils we have are related to decoding [network requests](https://github.com/klaviyo/klaviyo-android-sdk/blob/4a6f806064d1e24f1f4be1cefcdc87852b87cfa9/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequestDecoder.kt#L14) or [profile](https://github.com/klaviyo/klaviyo-android-sdk/blob/99a60a65bcb8d20cea5c81014ed3c69e892dc1bc/sdk/analytics/src/main/java/com/klaviyo/analytics/state/PersistentObservableProfile.kt#L42) objects from the persistent store. More tailor made, and simply catch JSONExceptions to detect unexpected JSON formats. 

So yeah this is a good addition

sdk/analytics/src/main/java/com/klaviyo/analytics/state/StateSideEffects.kt
evan-masseau: Yeah there's nothing private about that, I think we also log the whole API response on error elsewhere.

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/requests/KlaviyoErrorResponseDecoderTest.kt
```diff
@@ -0,0 +1,142 @@
...
+                  "meta": {}
+                }
+              ]
+            }
+        """.trimIndent()
+        assertEquals(errorResponse, KlaviyoErrorResponseDecoder.fromJson(JSONObject(errorJson)))
+    }
+
+    @Test
+    fun `Builds error response with some spooky nulls`() {
```
evan-masseau, line 47: 👻

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/requests/KlaviyoErrorResponseDecoderTest.kt
```diff
@@ -0,0 +1,142 @@
...
+                  },
+                  "links": {},
+                  "meta": {}
+                }
+              ]
+            }
+        """.trimIndent()
+        assertEquals(errorResponse, KlaviyoErrorResponseDecoder.fromJson(JSONObject(errorJson)))
+    }
+}
```
evan-masseau, line 142: How about some unhappy path tests, such as a test against an empty response body, or malformed JSON?

evan-masseau (commented):
Looks great! 
Will approve with tests against unexpected server responses like empty string, different format of json, or straight up malformed JSON.

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/requests/KlaviyoErrorResponseDecoderTest.kt
dan-peluso: Definitely can add more here!

evan-masseau: >    fun `Error json except it was written by somebody who does not know what json is`() {
yess

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequestTest.kt
```diff
@@ -399,4 +416,123 @@ internal class KlaviyoApiRequestTest : BaseApiRequestTest<KlaviyoApiRequest>() {
         assertEquals(null, get.body)
         assertEquals(0, get.attempts)
     }
+
+    @Test
+    fun `Malformed error response body`() {
```
evan-masseau, line 421: ah i see now 👍

evan-masseau (approved):
Sweet! We're catching up to you @ajaysubra

PR #161, opened 2024-05-01
Staging branch for next major release 3.0.0

evan-masseau:
Individual PRs to this branch should get reviewed like they were going in to master, and we will conduct final review of the whole set of changes together prior to 3.0.0 release.

PR #188, opened 2024-10-01
Automatically update push permission state on app resume

evan-masseau:
# Description
Adds automatic tracking of permission state change, by listening for `resume` events from the host app and re-checking permission state and device metadata against the latest push state data.

TODO:
- [x] Add migration notes and readme updates
- [x] Check with test app
- [x] Compile and test edge cases / extended use cases.
- [x] Check on performance implications

# Check List

## Changelog / Code Overview

- Attach a state side effect listener to activity lifecycle events
- On resume, if there is a push token in state, refresh overall push state
- Detach listeners on re-initialize to prevent dupes

## Test Plan

- [x] Quick test in an emulator
- [x] Unit tests
- [ ] Extended app testing

## Related Issues/Tickets

#161 - Required for version 3.0.0
CHNL-961

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
@@ -69,6 +69,7 @@ object Klaviyo {
         Registry.get<ApiClient>().startService()
 
         Registry.register<State>(KlaviyoState())
+        Registry.getOrNull<StateSideEffects>()?.detach()
         Registry.register<StateSideEffects>(StateSideEffects())
 
         Registry.get<State>().apiKey = apiKey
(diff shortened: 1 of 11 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -69,6 +69,7 @@ object Klaviyo {
         Registry.get<ApiClient>().startService()
 
         Registry.register<State>(KlaviyoState())
+        Registry.getOrNull<StateSideEffects>()?.detach()
```
evan-masseau, line 72: Because we didn't detach previous listeners, we risked accumulating multiple identical side effect listeners 😬

jordan-griffin (approved):

PR #191, opened 2024-10-08
[CHNL-12521] Looking for react-native strings to determine sdk name and version for config

dan-peluso:
# Description
Adding some logic to the Config so we can see if the SDK is being used in the react-native SDK or if it's being implemented natively. Will need to be merged along with [this](https://github.com/klaviyo/klaviyo-react-native-sdk/pull/180) to get the full functionality. tl;dr, this is the android version of [this](https://github.com/klaviyo/klaviyo-swift-sdk/pull/204/files)

# Check List

## Changelog / Code Overview

I removed Ajay's code that allows for these to be publicly exposed in the core config builder, since we wouldn't want someone manually building the Klaviyo object to change the thing we're trying to differentiate. There's also some error logging if we fail to use reflection. While reflection is never an ideal option, we need to get the id for the res item. Since this changes depending on how the app is built we need to search for the key name I've declared in the react-native SDK.

## Test Plan
Tested this on both android and react native test apps, and both sent up the proper config value as part of our requests. An interesting note on android was that I was unable to override the string resource, which is a good thing for design integrity. 

## Related Issues/Tickets
CHNL-12052
CHNL-12521

### DIFF
--- a/sdk/build.gradle
+++ b/sdk/build.gradle
@@ -48,15 +48,11 @@ subprojects {
                 proguardFiles getDefaultProguardFile("proguard-android-optimize.txt"), "proguard-rules.pro"
                 buildConfigField "String", "KLAVIYO_SERVER_URL", "\"${serverTarget}\""
                 buildConfigField "String", "KLAVIYO_API_REVISION", "\"${apiRevision}\""
-                buildConfigField "String", "VERSION", "\"${readXmlValue('src/main/res/values/strings.xml','klaviyo_sdk_version_override', project(':sdk:core'))}\""
-                buildConfigField "String", "NAME", "\"${readXmlValue('src/main/res/values/strings.xml','klaviyo_sdk_name_override', project(':sdk:core'))}\""
             }
             debug {
                 debuggable = true
                 buildConfigField "String", "KLAVIYO_SERVER_URL", "\"${serverTarget}\""
                 buildConfigField "String", "KLAVIYO_API_REVISION", "\"${apiRevision}\""
-                buildConfigField "String", "VERSION", "\"${readXmlValue('src/main/res/values/strings.xml','klaviyo_sdk_version_override', project(':sdk:core'))}\""
-                buildConfigField "String", "NAME", "\"${readXmlValue('src/main/res/values/strings.xml','klaviyo_sdk_name_override', project(':sdk:core'))}\""
             }
         }
 
--- a/sdk/build.gradle
+++ b/sdk/build.gradle
@@ -25,7 +25,7 @@ subprojects {
             targetSdkVersion versionFor(project, "version.android.targetSdk") as Integer
             testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
             versionCode versionFor(project, "version.klaviyo.versionCode") as Integer
-            versionName getVersionName()
+            versionName readXmlValue('src/main/res/values/strings.xml','klaviyo_sdk_version_override', project(':sdk:core'))
             consumerProguardFiles("consumer-rules.pro")
         }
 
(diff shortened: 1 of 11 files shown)


sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
```diff
@@ -116,11 +129,41 @@ object KlaviyoConfig : Config {
...
+            defaultValue
+        } else {
+            applicationContext.getManifestInt(key, defaultValue)
+        }
+
+    /**
+     * Helper to grab string resource from application using SDK - uses reflection
+     */
+    @SuppressLint("DiscouragedApi")
+    private fun getReactNativeSdkProperty(key: String): String? =
```
kennyklaviyo, line 143: not really a fan of how we're doing react native specific stuff, wondering if we can declare resource values in this android sdk side and in the react native sdk, override it using the same key

kennyklaviyo: and im wondering what level of validation we need to do since technically, developers would also be able to override these values

dan-peluso: I can remove the react-native specific naming here for sure. And yeah I agree it's a bit concerning that we're reflecting a resource value that could be overidden - in my research there's not many other ways we can get the value since the dependency route is Android SDK -> React Native SDK -> Client App. I tried using Manifest values but was unsuccessful since we can only attach meta data to an application (which we don't have in the react native SDK) and gradle BuildConfig values cannot be read from the depending sdk. If its any consolation, I did try and override on the Android test app and was not able to, which maybe is a sign that this would only be an issue for react-native clients who add this string resource to their android built folder. 

I like the idea of adding some backend checks around this so we only add config values for strings we expect (ex. android, react-native) so avoid any malicious play here and default to platform if we see anything funky. Would be cool to get another backend ticket in as well, definitely trying to learn more there. For now, I'll update this to be less react-native specific and maybe we can discuss post-hackathon?

kennyklaviyo: if we add `res/values/strings.xml` to the android sdk with:
```
<resources>
    <string name="klaviyo_sdk_name">android</string>
</resources>
```
and in the react native side declare the same thing, that'll override it right?

then in the android sdk side, theres no need for reflection, it already knows that R.string.klaviyo_sdk_name should exist

sdk/build.gradle
```diff
@@ -48,15 +48,15 @@ subprojects {
                 proguardFiles getDefaultProguardFile("proguard-android-optimize.txt"), "proguard-rules.pro"
                 buildConfigField "String", "KLAVIYO_SERVER_URL", "\"${serverTarget}\""
                 buildConfigField "String", "KLAVIYO_API_REVISION", "\"${apiRevision}\""
-                buildConfigField "String", "VERSION", "\"${versionFor(project, "version.klaviyo.versionName")}\""
-                buildConfigField "String", "NAME", "\"${klaviyoAndroidSDKName}\""
+                buildConfigField "String", "VERSION", "\"${readXmlValue('src/main/res/values/strings.xml','klaviyo_sdk_version_override', project(':sdk:core'))}\""
```
evan-masseau, line 51: aren't we just going to read this from the resource file at runtime? doesn't need to be a build field anymore?

dan-peluso: Good callout, I can remove these

evan-masseau (approved):
Honestly, I think this turned out pretty clean in the end. Thanks for persevering on this Dan.

evan-masseau:
Ooh just a little nit: I believe you've left behind some unused imports in gradle `import static de.fayard.refreshVersions.core.Versions.versionFor`, noticed while taking a look at the test app composite build situation

PR #192, opened 2024-10-17
[CHNL-13778] Fix network monitor over-broadcasting

dan-peluso:
# Description

We were seeing an issue manifest in the rate limit retry logic where requests that should have waited `retry-header` amount of time would send off regardless. This was happening because we would retry the top request on the queue when a network change gets broadcasted. In Android land, this capability is quite sensitive and can lead to us sending many requests out that have a 429 and ignoring the retry time value. This PR removes the broadcast on a network capability change. 

# Check List

## Changelog / Code Overview
- removing internal network broadcast relaying of the `ConnectivityManager.NetworkCallback.onCapabilityChanged` message. 

## Test Plan
- tested on a physical device to ensure the header retry value was being respected
- tested on an emulator and ensured the same thing

## Related Issues/Tickets
[CHNL-13778]

[CHNL-13778]: https://klaviyo.atlassian.net/browse/CHNL-13778?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sdk/core/src/main/java/com/klaviyo/core/networking/KlaviyoNetworkMonitor.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/networking/KlaviyoNetworkMonitor.kt
@@ -39,11 +39,6 @@ internal object KlaviyoNetworkMonitor : NetworkMonitor {
 
         override fun onUnavailable() = broadcastNetworkChange()
 
-        override fun onCapabilitiesChanged(
-            network: Network,
-            networkCapabilities: NetworkCapabilities
-        ) = broadcastNetworkChange()
-
         override fun onLinkPropertiesChanged(
             network: Network,
             linkProperties: LinkProperties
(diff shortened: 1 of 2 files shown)


sdk/core/src/main/java/com/klaviyo/core/networking/KlaviyoNetworkMonitor.kt
```diff
@@ -39,11 +39,6 @@ internal object KlaviyoNetworkMonitor : NetworkMonitor {
 
         override fun onUnavailable() = broadcastNetworkChange()
 
-        override fun onCapabilitiesChanged(
-            network: Network,
-            networkCapabilities: NetworkCapabilities
-        ) = broadcastNetworkChange()
-
         override fun onLinkPropertiesChanged(
```
evan-masseau, line 42: is `onLinkPropertiesChanged` needed either?

evan-masseau: https://developer.android.com/reference/android/net/LinkProperties
Maybe more granular than we need, though maybe not as flappy as capabilities at least

evan-masseau: Yeah, actually something like a proxy change would be a good event to trigger the network queue to flush.

evan-masseau (approved):

evan-masseau:
This was a great find 👍

PR #194, opened 2024-10-23
CHNL-13253 IAM Basic Webview + Module Scaffolding

dan-peluso:
# Description
Adding a new module for IAM, creating a new class `KlaviyoWebView` and singleton `InAppMessaging` for showing a basic webview. Currently this is hardcoded to the IAMTest file added to this module's assets, but we will update this to support a backend-sent file soon. 

Note the bridge interactions and event sending! This is crucial infra for how we communicate with the form in the future.

Almost all commits here are from @evan-masseau and @kennyklaviyo, this PR mostly serves to initialize the feature branch. 

# Check List

## Test Plan
- adding a new page to the test app for usage of this module

## Related Issues/Tickets
CHNL-13253


sdk/messaging/build.gradle
```diff
@@ -0,0 +1,39 @@
+project.description = 'In-app messaging functionality for the Klaviyo SDK suite'
+evaluationDependsOn(':sdk')
+
+def ext = rootProject.ext
+
+android {
+    namespace 'com.klaviyo.messaging'
```
jordan-griffin, line 7: Could do `namespace "${klaviyoGroupId}.messaging"` here

sdk/messaging/src/main/java/com/klaviyo/messaging/InAppMessaging.kt
```diff
@@ -0,0 +1,26 @@
...
+import java.io.BufferedReader
+
+object InAppMessaging {
+
+    fun triggerInAppMessage(activity: Activity) {
+        val rootView = activity.getRootViewGroup()
+        val webView = KlaviyoWebView()
+        val html = Registry.config.applicationContext
+            .assets
+            .open("IAMTest.html")
```
jordan-griffin, line 15: Obviously just is just the initial work to build out this new module, but should this testing logic be merged at all? Seems like something that should be maybe held in another branch - not the feature branch we're building up for this concept

settings.gradle
```diff
@@ -11,10 +11,13 @@ dependencyResolutionManagement {
...
+include ":sdk:core", ":sdk:analytics", ":sdk:push-fcm", ":sdk:fixtures", ":sdk:messaging"
 
 refreshVersions {
     rejectVersionIf {
         candidate.stabilityLevel.isLessStableThan(current.stabilityLevel)
     }
 }
+include ':sdk'
+include ':sdk:core', ':sdk:analytics', ':sdk:push-fcm', ':sdk:fixtures'
+include ':sample'
```
jordan-griffin, line 23: whats this?

settings.gradle
```diff
@@ -11,10 +11,13 @@ dependencyResolutionManagement {
...
-include ":sdk:core", ":sdk:analytics", ":sdk:push-fcm", ":sdk:fixtures"
+include ":sdk:core", ":sdk:analytics", ":sdk:push-fcm", ":sdk:fixtures", ":sdk:messaging"
 
 refreshVersions {
     rejectVersionIf {
         candidate.stabilityLevel.isLessStableThan(current.stabilityLevel)
     }
 }
+include ':sdk'
+include ':sdk:core', ':sdk:analytics', ':sdk:push-fcm', ':sdk:fixtures'
```
jordan-griffin, line 22: Are these two lines needed when we have the `include` updated above on line 14 already?

versions.properties
```diff
@@ -9,6 +9,8 @@
 ####
 #### NOTE: Some versions are filtered by the rejectVersionIf predicate. See the settings.gradle.kts file.
 
+version.com.google.gms..google-services=4.4.2
```
jordan-griffin, line 12: Extra `.` here?

sdk/messaging/build.gradle
```diff
@@ -0,0 +1,39 @@
...
+
+afterEvaluate {
+    publishing {
+        publications {
+            // Creates a Maven publication called "release".
+            release(MavenPublication) {
+                from components[ext.publishBuildVariant]
+                groupId = klaviyoGroupId
+                artifactId = 'messaging'
+                version = readXmlValue('src/main/res/values/strings.xml','klaviyo_sdk_version_override', project(":sdk:core"))
```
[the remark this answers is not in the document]
dan-peluso: All the modules set versions like this to support overwriting versioning for clients (React native). We use a core-level string to keep the module versions aligned.

sdk/messaging/src/main/java/com/klaviyo/messaging/InAppMessaging.kt
dan-peluso: Yeah this is definitely a difference between the ios and Android branches for doing this - personally I figured since it was already written there wasn't much reason to remove. Since we will be just continuing to add to this branch I do like how this shows what it looks like to render a page (specifically one that handles events). If you want though I can yank this and add to a new PR - or remove entirely.

settings.gradle
dan-peluso: A little sneaky leftover from the sample app that was a part of the internal sdk feature branch. I'll get these outta here

versions.properties
dan-peluso: This file is strange - it gets auto generated but I'm still not sure why this was added. Going to test and see if I can still get a build to work without this. The double dots might be necessary unfortunately.

sdk/messaging/src/main/java/com/klaviyo/messaging/InAppMessaging.kt
jordan-griffin: That's all pretty fair sounding to me. I'm like 50/50, not sure if we should have testing code set into our feature branch unless we're really sure we'll clean out all the files later but totally get why this is here. In the current state of things we have no other way to easily test this stuff in the initial stages and separating the branch out just creates more overhead in our future work as we constantly aim to re-test our changes

jordan-griffin: All that to say, I would not block on this conversation at all!

dan-peluso:
@evan-masseau One thing to point out on this is that we might need to come up with a clever way to expose the user-agent to the new module, since that's just become internal only in our 3.0.0 update and we'll definitely need it for IAM

dan-peluso:
> > @evan-masseau One thing to point out on this is that we might need to come up with a clever way to expose the user-agent to the new module, since that's just become internal only in our 3.0.0 update and we'll definitely need it for IAM
> 
> @dan-peluso I don't remember why we made it private, can we just expose it again? Seems harmless if read only

@evan-masseau Made it the only public field of the DeviceProperties object

evan-masseau (approved):

PR #196, opened 2024-11-04
CHNL-13255 Add wrapper view for hosting form webview

dan-peluso:
# Description
Creates a dialog we can wrap the webview we'll interact with around. I have a video showing the functionality below, but note this PR does *not* address the following open issues. 
- How do we determine if the content has fully loaded? Onsite team knows about this and are coming up with a way to communicate that, for now we're using the documentReady method which klaviyo.js does not respect.
- How do we communicate a button press closing the dialog? Our current bridge functionality looks for the view to be removed from the tree, which explains why the dialog closes _after_ the webview has already been collapsed. We need a way of hooking into the button press action rather than the removal of the button id from the tree.

[webview-as-dialog.webm](https://github.com/user-attachments/assets/3077491e-68d5-4d42-aca6-ba24831ae6c3)

# Check List

^ we might have an issue supporting devices below Android API 17 but our SDK does not support those anyway. 

## Changelog / Code Overview
- making a dialog for the webview, and extrapolating the webview client code into its own class for better separation of intents

### DIFF
--- a/sdk/messaging/src/main/java/com/klaviyo/messaging/InAppMessaging.kt
+++ b/sdk/messaging/src/main/java/com/klaviyo/messaging/InAppMessaging.kt
@@ -1,15 +1,15 @@
 package com.klaviyo.messaging
 
 import android.app.Activity
-import android.view.ViewGroup
+import android.webkit.JavascriptInterface
 import com.klaviyo.core.Registry
 import java.io.BufferedReader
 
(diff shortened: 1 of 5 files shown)


dan-peluso:
We opted to not do this since we can launch a fullscreen webview much easier than handling the wrapper, and the styling looks great on it

PR #197, opened 2024-11-08
[CHNL-14635] Fetch Full Forms

dan-peluso:
# Description
- adding new request for 'full-forms' and logging the response

A few things to consider
 - This response will be going to the `klaviyo.js loadData()` call, so we don't need to serialize more than the top 3 fields here (and probably wouldn't want to given how we don't care about the details)
 - For now we're using the web forms endpoint, so we have to take only the first item on the list of forms
 - We still need to define that function in the klaviyo.js updates, so for now this PR is just creating the request and logging with some basic serialization, I will integrate once klaviyo.js gets updated

# Check List

## Related Issues/Tickets
CHNL-14635

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
@@ -7,6 +7,7 @@ import com.klaviyo.analytics.model.Event
 import com.klaviyo.analytics.model.EventMetric
 import com.klaviyo.analytics.model.Profile
 import com.klaviyo.analytics.networking.requests.EventApiRequest
+import com.klaviyo.analytics.networking.requests.FormApiRequest
 import com.klaviyo.analytics.networking.requests.KlaviyoApiRequest
 import com.klaviyo.analytics.networking.requests.KlaviyoApiRequest.Status
 import com.klaviyo.analytics.networking.requests.KlaviyoApiRequestDecoder
@@ -115,6 +116,10 @@ internal object KlaviyoApiClient : ApiClient {
 
     override fun offApiRequest(observer: ApiObserver) {
         apiObservers -= observer
+    }
+
+    override fun getActiveForms() {
+        enqueueRequest(FormApiRequest())
     }
 
     private fun broadcastApiRequest(request: KlaviyoApiRequest) {
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
@@ -119,6 +119,7 @@ internal object KlaviyoApiClient : ApiClient {
     }
 
     override fun getActiveForms() {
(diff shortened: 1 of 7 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/FullFormsApiResponseDecoder.kt
```diff
@@ -0,0 +1,26 @@
...
+                json.getJSONArray(FullsFormsResponse.FULL_FORMS)[0]
+            )
+            put(
+                FullsFormsResponse.FORM_SETTINGS,
+                json.getJSONObject(FullsFormsResponse.FORM_SETTINGS)
+            )
+            put(
+                FullsFormsResponse.DYNAMIC_INFO_CONFIG,
+                json.getJSONObject(FullsFormsResponse.DYNAMIC_INFO_CONFIG)
+            )
```
njparadis, line 24: (question): Are these global settings/config info or do we need to key into them somehow to get details for the specific form we want to display?

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -117,6 +118,10 @@ internal object KlaviyoApiClient : ApiClient {
         apiObservers -= observer
     }
 
+    override fun getActiveForms() {
+        enqueueRequest(FormApiRequest())
+    }
+
```
njparadis, line 124: (question, non-blocking): I see we log something out before enqueuing the request in some of the other methods in here. Do we want to do that for `getActiveForms` too?

njparadis (commented):
LGTM, just one question around the shape of the response from the API. Do you think we need any tests around that stuff? Logic seems straightforward enough to me but maybe an example response would be helpful documentation for our future selves 👍

njparadis (approved):

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/FullFormsApiResponseDecoder.kt
dan-peluso: I'm pretty certain we don't need to key into these. The three fields here are what we need to send as args to the 'loadForm' Klaviyo.js call, all this decoder really does it take the top of the list of full forms sent. I'm not further serializing anything to bc we're really just relaying it all forward.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
dan-peluso: Definitely not a bad idea, I can add that here

dan-peluso:
> LGTM, just one question around the shape of the response from the API. Do you think we need any tests around that stuff? Logic seems straightforward enough to me but maybe an example response would be helpful documentation for our future selves 👍

I was planning to add those once we have more stucture defined around what we actually do with that response. For now since we can't really pass it to klaviyo.js I figured this ticket can be limited to setting up the new request type and creating some of the infrastructure for when we do need to format for passing to klaviyo.js

PR #201, opened 2024-12-26
[CHNL-15530] Message Bus Support

dan-peluso:
# Description
Mostly logic added to the bridge.js (which I am certainly in support of making a different repo so we can import the same file for every native client we have) to support listening for the bus to be created, and then listening for each of the messages and relaying to the kotlin webview client.

# Check List

## Test Plan
Testing this requires some decent setup:
1 - get app running with onsite modules

2 - get fender running [(using this branch)](https://github.com/klaviyo/fender/pull/36364) with onsite dependencies as well - should be able to publish a form in the webapp and then make it live

3 - change the HTML code to point to a local `:8080` onsite instead of production, and make sure to run the following adb commands if you're running a network proxy
`adb reverse tcp:8080 tcp:8080`
`adb reverse tcp:4001 tcp:4001`

You should now be able to see the locally created form from the android test app in the 'messaging' menu. Check the logs after loading and ensure that a `inAppFormsDataLoaded` has come through.

## Related Issues/Tickets
https://klaviyo.atlassian.net/browse/CHNL-15530

### DIFF
--- a/sdk/messaging/src/main/assets/bridge.js
+++ b/sdk/messaging/src/main/assets/bridge.js
@@ -46,6 +46,27 @@
                         WebViewBridge.postMessageToNative("close", {});
                     });
                 });
+
+                // add a listener to the message bridge init function, so we only listen to the bus if it's ready
+                window.addEventListener('inAppFormsInitMessage', function(event) {
+                    const supportedEventTypes = event.detail.eventTypes;
+                    console.log('Supporting message types:', supportedEventTypes);
+
+                    const messageBus = window['inAppFormsMessageBus']?.messageBus;
+                    if (!messageBus) {
+                        console.error('Message bus not found');
+                        return;
+                    } else {
+                        console.log('Found the message bus!');
+                    }
+
+                    supportedEventTypes.forEach(eventType => {
+                        console.log('Registering listener for eventType: ', eventType);
+                        messageBus.addEventListener(eventType.toString(), function(subEvent) {
+                            WebViewBridge.postMessageToNative(subEvent.type, subEvent.detail);
+                        });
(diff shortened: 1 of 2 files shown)


sdk/messaging/src/main/assets/bridge.js
```diff
@@ -46,6 +46,27 @@
                         WebViewBridge.postMessageToNative("close", {});
                     });
                 });
+
+                // add a listener to the message bridge init function, so we only listen to the bus if it's ready
+                window.addEventListener('inAppFormsInitMessage', function(event) {
+                    const supportedEventTypes = event.detail.eventTypes;
+                    console.log('Supporting message types:', supportedEventTypes);
+
+                    const messageBus = window['inAppFormsMessageBus']?.messageBus;
```
evan-masseau, line 55: TIL you can use `?` for optional chaining in JS nowadays

evan-masseau: Oof this is annoying, but this syntax may not be supporting in an iOS webview on our minimum supported iOS version, ios 13, see 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#browser_compatibility

evan-masseau:
@dan-peluso I've created a [repo](https://github.com/klaviyo/klaviyo-in-app-js-bridge/tree/master) for the bridge js code, and I included the changes you have here. So this PR will get superceded by [this one](https://github.com/klaviyo/klaviyo-android-sdk/pull/204), adding the git submodule with the new repo.

evan-masseau:
Closing this bc similar code will just live in fender.

PR #203, opened 2025-01-03
Update android-master.yml to address deprecated action

jordan-griffin:
# Description

The `upload-artifact@v2` action is deprecated. Its been deprecated since June and Github quickly deprecated the V3 as well. The V4 is available and we should adopt that to avoid downtime on out Github action for master builds.

Moving from V2/V3 -> V4 I followed [this migration guide](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md). Since artifacts are no longer mutable I reckon we could get away with marking the artifact as an overwrite so it deletes the old and creates new. Seemed more straightforward to me than building a unique name for each build.
Note: I read some things about `overwrite` causing issues when running parallel jobs but felt this wouldn't likely be an issue for us yet.

Starting Jan 9th, `upload-artifact` will experience brownouts with full disablement by end of month so that's why I thought I'd cut this now.

_Note: I haven't tested this in any way yet. It's just a suggestion for now._

# Check List

### DIFF
--- a/.github/workflows/android-master.yml
+++ b/.github/workflows/android-master.yml
@@ -77,8 +77,9 @@ jobs:
 
       # Stores the generated AAR to an outputs folder
       - name: Store AAR
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v4
         with:
           name: sdk
           path: app/build/outputs/aar/debug/aar-debug.aar
+          overwrite: true
 


.github/workflows/android-master.yml
```diff
@@ -77,8 +77,9 @@ jobs:
 
       # Stores the generated AAR to an outputs folder
       - name: Store AAR
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v4
```
evan-masseau, line 80: tbh i'm not even sure what this does 🤦

jordan-griffin: Hm nevermind, the artifact IS meant to be there. But I dont see it in our builds, now I'm wondering why https://github.com/actions/upload-artifact?tab=readme-ov-file#where-does-the-upload-go

jordan-griffin: Ah I see some github chatter that this was a known issue with `upload-artifact` v3, meaning if we keep this step here but update it to v4 currently, the artifact should always append to the action's run in the UI so we can download for potential debugging on failed master runs.
Minor but I think worth keeping in

njparadis (approved):

dan-peluso (approved):

PR #204, opened 2025-01-03
Git submodule for bridge JS

evan-masseau:
Webview's html and the bridge js are sourced from a git submodule now.
Adjust some code for updated bridge JS stuff. 

# Description

# Check List


sdk/messaging/src/main/java/com/klaviyo/messaging/KlaviyoWebView.kt
```diff
@@ -31,20 +31,22 @@ import org.json.JSONObject
...
             .assets
-            .open("bridge.js")
+            .open("klaviyo-in-app-js-bridge/bridge.js")
             .bufferedReader()
             .use(BufferedReader::readText)
             .apply {
                 val opts = JSONObject()
-                    .put("bridgeName", JS_BRIDGE_NAME)
+                    .put("messageHandlerName", JS_BRIDGE_NAME)
+                    .put("klaviyoJsUrl", "http://local-klaviyo.com:8080/onsite/js/klaviyo.js")
```
njparadis, line 45: should this point to the production URL?

njparadis (approved):

sdk/messaging/src/main/java/com/klaviyo/messaging/KlaviyoWebView.kt
njparadis: Sounds good!

evan-masseau:
Decided against this approach after all

PR #207, opened 2025-01-14
[CHNL-16219] Updating IAF branch with local setup + removing unused code

dan-peluso:
# Description
- working state for local IAF development
- note that closing will *not* actually close the webview since we are no longer creating that on the SDK side. We will soon be adding that to fender.

# Check List

## Related Issues/Tickets
https://klaviyo.atlassian.net/browse/CHNL-16219


sdk/messaging/src/main/java/com/klaviyo/messaging/KlaviyoWebView.kt
```diff
@@ -56,38 +39,18 @@ class KlaviyoWebView : WebViewClient(), WebViewCompat.WebMessageListener {
...
         it.settings.javaScriptEnabled = true
 
-        if (USE_NEW_FEATURES && isFeatureSupported(WEB_MESSAGE_LISTENER)) {
+        if (isFeatureSupported(WEB_MESSAGE_LISTENER)) {
             Registry.log.verbose("$WEB_MESSAGE_LISTENER Supported")
             WebViewCompat.addWebMessageListener(
                 it,
                 JS_BRIDGE_NAME,
-                setOf(Registry.config.baseUrl),
+                setOf("http://a.local-klaviyo.com:8080"),
```
evan-masseau, line 47:
```suggestion
                //TODO - sort out how to toggle between local and production. 
                setOf("http://a.local-klaviyo.com:8080"),
```
For viz to any other reviewers, we're temporarily just pointing to local to reduce confusion, plan to sort out how to toggle this from the test app shortly.

sdk/messaging/src/main/java/com/klaviyo/messaging/KlaviyoWebView.kt
```diff
@@ -126,7 +89,7 @@ class KlaviyoWebView : WebViewClient(), WebViewCompat.WebMessageListener {
     }
 
     fun loadHtml(html: String) = webView.loadDataWithBaseURL(
-        Registry.config.baseUrl,
+        "http://a.local-klaviyo.com:8080",
```
evan-masseau, line 92:
```suggestion
        //TODO - sort out how to toggle between local and production. 
        "http://a.local-klaviyo.com:8080",
```

versions.properties
```diff
@@ -9,6 +9,8 @@
 ####
 #### NOTE: Some versions are filtered by the rejectVersionIf predicate. See the settings.gradle.kts file.
 
+version.androidx.compose.compiler=1.5.10
+
```
evan-masseau, line 13:
```suggestion
```

I think we'd wanted to avoid committing this?

evan-masseau: I'm actually not sure if that really matters. I guess my one nit would be if we do commit it, lets put it lower down with other androidx deps

evan-masseau (approved):
just added todo comments for clarity, merge away

versions.properties
dan-peluso: ah my bad - this is the one that comes up every time I build

PR #208, opened 2025-01-24
Update SystemClock.kt

Musharaf13:
Updated the SDK to ensure that timestamps are always sent in English, regardless of the user's device language settings. This resolves an issue where the API fails if an Arabic timestamp is sent due to the user's device being set to Arabic.

# Description
Updated the SDK to ensure that timestamps are always sent in English, regardless of the user's device language settings. This resolves an issue where the API fails if an Arabic timestamp is sent due to the user's device being set to Arabic.

# Check List

 Are you changing anything with the public API?
No, this change is internal and does not affect the public API.
 Are your changes backwards compatible with previous SDK Versions?
Yes, the change maintains compatibility with previous SDK versions.
 Have you tested this change on a real device?
Yes, the changes have been tested on a real device to ensure proper functionality.
 Have you added unit test coverage for your changes?
Unit tests have been added to verify that the timestamp is always sent in English.
 Have you verified that your changes are compatible with all Android versions the SDK currently supports?
Yes, the changes have been verified to work across all supported Android versions.

## Changelog / Code Overview
Modified: The SystemClock implementation now explicitly uses Locale.ENGLISH in the SimpleDateFormat instance to ensure timestamps are consistently formatted in English.
Reason: The API fails when Arabic-formatted timestamps are sent, as the system defaults to the device's language settings. This update ensures consistent API behavior regardless of the user's locale.
Supporting Material:

## Test Plan
Tested the updated SystemClock implementation by simulating device language settings in Arabic and verifying that the timestamp is sent in English.
Verified that the API accepts the updated timestamp format without errors.
Confirmed the change works across different Android versions and devices.

### DIFF
--- a/sdk/core/src/main/java/com/klaviyo/core/config/SystemClock.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/config/SystemClock.kt
@@ -3,14 +3,16 @@ package com.klaviyo.core.config
 import android.annotation.SuppressLint
 import java.text.SimpleDateFormat
 import java.util.Date
+import java.util.Locale
 import java.util.TimeZone
 import java.util.Timer
 import kotlin.concurrent.schedule
 
 internal object SystemClock : Clock {
 
+    // Explicitly set Locale.ENGLISH to ensure consistent formatting
     @SuppressLint("SimpleDateFormat")
-    private val format = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ").apply {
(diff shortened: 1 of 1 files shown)


evan-masseau (approved):
Thank you for bringing this to our attention and applying a fix!

dan-peluso:
Going to try and get a unit test in for this today. Thanks for opening this!

dan-peluso:
I didn't have push permissions in your fork so I made a new PR with a unit test

https://github.com/klaviyo/klaviyo-android-sdk/pull/210

@Musharaf13

kennyklaviyo (approved):

PR #212, opened 2025-01-27
[CHNL-16656] Enqueue Aggregate Events from IAF

dan-peluso:
# Description
- support for new API request
- support for deserializing new event type
- unit tests for the mapper 

note that we are adding to the interface of the public ApiClient; however, this is only available through the `core` module which is not a recommended import for our users. Currently we tell users to send requests through the `Klaviyo.` object, which I have not included a route for here.

# Check List

- [ish] Are you changing anything with the public API?

## Test Plan
- tested on local with mocked responses from the response google doc Evan opened

## Related Issues/Tickets
https://klaviyo.atlassian.net/browse/CHNL-16656

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
@@ -6,6 +6,8 @@ import android.os.Looper
 import com.klaviyo.analytics.model.Event
 import com.klaviyo.analytics.model.EventMetric
 import com.klaviyo.analytics.model.Profile
+import com.klaviyo.analytics.networking.requests.AggregateEventApiRequest
+import com.klaviyo.analytics.networking.requests.AggregateEventPayload
 import com.klaviyo.analytics.networking.requests.EventApiRequest
 import com.klaviyo.analytics.networking.requests.KlaviyoApiRequest
 import com.klaviyo.analytics.networking.requests.KlaviyoApiRequest.Status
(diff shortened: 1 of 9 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -66,6 +68,11 @@ internal object KlaviyoApiClient : ApiClient {
         enqueueRequest(PushTokenApiRequest(token, profile))
     }
 
+    override fun enqueueAggregateEvent(payload: AggregateEventPayload) {
+        Registry.log.verbose("Enqueuing Aggregate Event request")
+        enqueueRequest(AggregateEventApiRequest(payload))
+    }
+
     override fun enqueueUnregisterPushToken(apiKey: String, token: String, profile: Profile) {
         enqueueRequest(UnregisterPushTokenApiRequest(apiKey, token, profile))
```
evan-masseau, line 77: unrelated: mind adding this while you're here? 
```suggestion
        Registry.log.verbose("Enqueuing unregister token request")
        enqueueRequest(UnregisterPushTokenApiRequest(apiKey, token, profile))
```

jordan-griffin (approved):
This all looks pretty solid!

evan-masseau (approved):
no notes (well except that unrelated note)

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
dan-peluso: makes sense to me

PR #213, opened 2025-01-28
[CHNL-16704] Deeplink Support for IAF

dan-peluso:
# Description
 
Supporting Deeplinks for IAF messages, according to the contract specc'd out in @evan-masseau s doc. The deeplink logic is very similar to the one we have for adding an intent to deeplinking notifications, but I did have to change around our implementation of the webview to support the `addFlags(Intent.FLAG_ACTIVITY_SINGLE_TOP)` line. This can only be used when launching an activity from an activity (noticeably _not_ from an application which we were using as our primary context before). We already almost certainly want some sort of Activity reference for the API anyhow, so I'm just forwarding this to the klaviyo webview class we already have. I added some safety around the activity reference, since this is a super common android memory leak I'd like to avoid. Hence, the weak reference which will be garbage collected once the webview is destroyed. 

# Check List

[deeplink-and-aggregate-events.webm](https://github.com/user-attachments/assets/1459abc8-d946-42ee-85b1-12c02054854e)

## Related Issues/Tickets
https://klaviyo.atlassian.net/browse/CHNL-16704

### DIFF
--- a/sdk/messaging/src/main/java/com/klaviyo/messaging/InAppMessaging.kt
+++ b/sdk/messaging/src/main/java/com/klaviyo/messaging/InAppMessaging.kt
@@ -9,15 +9,15 @@ object InAppMessaging {
 
     fun triggerInAppMessage(activity: Activity) {
         val rootView = activity.getRootViewGroup()
-        val webView = KlaviyoWebView()
-        val html = Registry.config.applicationContext
+        val webView = KlaviyoWebView(activity)
+        val html = activity
             .assets
             .open("IAMTest.html")
             .bufferedReader()
             .use(BufferedReader::readText)
-            .replace("KLAVIYO_PUBLIC_KEY_PLACEHOLDER", Registry.config.apiKey)
-            .replace("SDK_NAME", Registry.config.sdkName)
-            .replace("SDK_VERSION", Registry.config.sdkVersion)
+            .replace(IAF_PUBLIC_KEY_PLACEHOLDER, Registry.config.apiKey)
+            .replace(IAF_SDK_NAME_PLACEHOLDER, Registry.config.sdkName)
+            .replace(IAF_SDK_VERSION_PLACEHOLDER, Registry.config.sdkVersion)
 
         webView.loadHtml(html)
         webView.addTo(rootView)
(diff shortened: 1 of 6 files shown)


kennyklaviyo (approved):

sdk/messaging/src/main/java/com/klaviyo/messaging/KlaviyoWebView.kt
```diff
@@ -130,9 +136,26 @@ class KlaviyoWebView : WebViewClient(), WebViewCompat.WebMessageListener {
                 is KlaviyoWebFormMessageType.AggregateEventTracked -> {
                     Registry.get<ApiClient>().enqueueAggregateEvent(messageType.payload)
                 }
+                is KlaviyoWebFormMessageType.DeepLink -> {
```
evan-masseau, line 139: nit: If its more than just a function call, maybe a when case should be pulled into a separate function for readability.

sdk/messaging/src/main/java/com/klaviyo/messaging/KlaviyoWebView.kt
```diff
@@ -21,19 +22,24 @@ import com.klaviyo.analytics.Klaviyo
...
-class KlaviyoWebView : WebViewClient(), WebViewCompat.WebMessageListener {
+class KlaviyoWebView(
+    activity: Activity
+) : WebViewClient(), WebViewCompat.WebMessageListener {
     companion object {
         private const val MIME_TYPE = "text/html"
         private const val JS_BRIDGE_NAME = "Klaviyo"
     }
 
+    private val activityRef = WeakReference(activity)
```
evan-masseau, line 35: Errr, Is this something we should be doing with the application context that we store in `Registry.config` too?

sdk/messaging/src/main/java/com/klaviyo/messaging/InAppMessaging.kt
```diff
@@ -9,15 +9,15 @@ object InAppMessaging {
 
     fun triggerInAppMessage(activity: Activity) {
```
evan-masseau, line 10: oh neat we already were asking for activity here

evan-masseau (approved):

sdk/messaging/src/main/java/com/klaviyo/messaging/KlaviyoWebView.kt
dan-peluso: Haha yeah so this definitely exposes quite the rabbit hole of things we could fix with our current Registry method of exposing things across the application. I'd say there's a chance it could, but this probably warrants an investigative story. Thankfully, there's tons of great tools that we can use for android application/activity leaks, because this is a popular issue. But to answer your question, I can't say for sure if we need this immediately.

PR #214, opened 2025-02-03
fix: prevent crash when connection issue product a response without body

Shywim:
# Description

# Check List

## Changelog / Code Overview

- errorStream only handles HTTP error with a body, when there is a connection issue, there is no body, so no stream is returned
- This makes the app crash

image

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
@@ -378,7 +378,9 @@ internal open class KlaviyoApiRequest(
             else -> connection.errorStream
         }
 
-        responseBody = BufferedReader(InputStreamReader(stream)).use { it.readText() }
+        if (stream != null) {
+            responseBody = BufferedReader(InputStreamReader(stream)).use { it.readText() }
+        }
 
         return status
     }


evan-masseau:
Thanks for the report @Shywim! Its odd how `stream` compiles to `InputStream!` and yet clearly can be null.
Wondering if you have any info on either: 
- Volume that you're seeing of this crash report
- Reproduction steps

Although I can take a guess at the repro steps: seems like this would occur if network connection is lost between starting to send the request (because we check network connection before sending) and before a response from the server can be received (i.e. leading to a timeout). 
Next steps on our end will be to write a reproducing unit test and merge your fix. I also think before we release it, we may need a tweak here such that the network request remains queued and gets retried later. I think as written right now, this would lead to the request being dropped, causing data loss.

Shywim:
> Thanks for the report @Shywim! Its odd how `stream` compiles to `InputStream!` and yet clearly can be null.

I am not very familiar with Kotlin, but I came to understand that `Type!` means that the type could or couldn't be null, which comes from missing @Nullable or @NotNull annotations on Java code (I did not find the info in Kotlin docs, but here: https://stackoverflow.com/a/27462091). It felt confusing to me, since everywhere else I know, `!` would mean "never null" 

> Wondering if you can provide additional info on either:
> 
>     * Volume that you're seeing of this crash report
> 
>     * Reproduction steps
> 
> 
> Although I can take a guess at the repro steps: seems like this would occur if network connection is lost between starting to send the request (because we check network connection before sending) and before a response from the server can be received (i.e. leading to a timeout).

Yes sorry I was quite expeditious, but this is exactly the pattern we see from our logs.
It impacts around 3% (~350) of our Android users. Also I am not sure how to make 100% reproduction steps, generally for these kind of issues I debug them in the train to have a unreliable network while not being totally disconnected 😅

evan-masseau:
> When the type ends with ! it means that this is a platform type and compiler does not enforce null-safety for it
That makes sense, thank you. The other puzzling thing here is why the timeout exception isn't thrown a little earlier in the stack. But I'll try out some network conditioner tools I've used in the past and see what I can repro, failing that I'll have to look up my local train lines 🤣 

In the mean time, I'm going to get your PR merged and build it in jitpack so that you can pull this fix into your app. I'll follow up in a minute with instructions how to access that artifact until we have put together an official release including your patch.

ajaysubra (approved):

dan-peluso (approved):

evan-masseau:
@Shywim We'll be working to get this into an upcoming patch release shortly, but in the mean time if you want to test that your fix made it in and is working, install these artifacts in your gradle file (they use the commit from master that I just merged)

```
com.github.klaviyo.klaviyo-android-sdk:push-fcm:fb5dd9510b
com.github.klaviyo.klaviyo-android-sdk:analytics:fb5dd9510b
```

evan-masseau:
@Shywim thanks again for contributing, just wanted to let you know this got officially released in version 3.2.0 today!

PR #215, opened 2025-02-03
Null input stream unit test

dan-peluso:
# Description
BOY this was annoying to write

verified before/after the fix to see the failure. Thanks @Shywim for putting up the original pr:

https://github.com/klaviyo/klaviyo-android-sdk/pull/214

# Check List

### DIFF
--- a/sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
+++ b/sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
@@ -728,13 +728,6 @@ internal class KlaviyoApiClientTest : BaseTest() {
 
         // Mock network connection and config
         every { Registry.networkMonitor.isNetworkConnected() } returns true
-        every { Registry.config.networkTimeout } returns 1000
-        every { Registry.config.baseUrl } returns "http://localhost"
-        every { Registry.config.networkMaxAttempts } returns 3
-        every { Registry.config.networkMaxRetryInterval } returns 30000
-        every { Registry.config.networkFlushIntervals } returns longArrayOf(10000, 20000, 30000)
-        every { Registry.config.networkJitterRange } returns 1..5
-
         // Mock URL and URLConnection
         val urlMock = mockk<URL>()
         val connectionMock = mockk<HttpURLConnection>()
--- a/sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
+++ b/sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
@@ -20,14 +20,17 @@ import com.klaviyo.core.networking.NetworkMonitor
 import com.klaviyo.core.networking.NetworkObserver
 import com.klaviyo.fixtures.BaseTest
 import io.mockk.every
+import io.mockk.just
 import io.mockk.mockk
 import io.mockk.mockkConstructor
 import io.mockk.mockkObject
+import io.mockk.runs
 import io.mockk.slot
 import io.mockk.spyk
 import io.mockk.unmockkConstructor
 import io.mockk.unmockkObject
 import io.mockk.verify
+import java.net.HttpURLConnection
 import java.net.URL
 import kotlinx.coroutines.Dispatchers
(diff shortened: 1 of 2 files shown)


ajaysubra (approved):

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
```diff
@@ -712,4 +717,50 @@ internal class KlaviyoApiClientTest : BaseTest() {
...
+        every { urlMock.protocol } returns "http"
+        every { connectionMock.inputStream } returns null
+        every { connectionMock.errorStream } returns null
+        every { connectionMock.responseCode } returns HttpURLConnection.HTTP_INTERNAL_ERROR
+        every { connectionMock.headerFields } returns emptyMap()
+        every { connectionMock.setRequestProperty(any(), any()) } returns Unit
+        every { connectionMock.requestMethod = any() } returns Unit
+        every { connectionMock.readTimeout = any() } returns Unit
+        every { connectionMock.connectTimeout = any() } returns Unit
+        every { connectionMock.disconnect() } just runs
```
evan-masseau, line 753: lol `just runs` haven't seen this one before and makes me chuckle.

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
```diff
@@ -712,4 +717,50 @@ internal class KlaviyoApiClientTest : BaseTest() {
...
+        every { connectionMock.connect() } just runs
```
evan-masseau, line 754: Not at all the point of the unit test, but I'm puzzled about the fact that `connect()` didn't raise the `SocketTimeoutException` for us to catch earlier. Its even in the screenshot of the bug report.

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
```diff
@@ -712,4 +717,50 @@ internal class KlaviyoApiClientTest : BaseTest() {
...
+    fun `response body handles null stream correctly`() {
+        // Create a KlaviyoApiRequest with required arguments
+        val request = KlaviyoApiRequest("mockPath", RequestMethod.GET)
+
+        // Spy on that specific instance
+        val spiedRequest = spyk(request)
+
+        // Mock network connection and config
+        every { Registry.networkMonitor.isNetworkConnected() } returns true
+        every { Registry.config.networkTimeout } returns 1000
```
evan-masseau, line 731: nit: not obligatory but in case you didn't see, there is a shared mockConfig setup in the BaseTest class that should cover these mocks

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
dan-peluso: haha yeah I'm also a big fan of this one

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
dan-peluso: Yeah it's a bit concerning

evan-masseau: I wonder if its a platform thing, like some android version that doesn't throw that exception properly

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
dan-peluso: Nice, yeah at a certain point I was just throwing every mock I could to get this to pass. I can revisit this and see what I can remove

evan-masseau: I think you'd just need to add one: 

`every { networkTimeout } returns 1000`

PR #216, opened 2025-02-04
IAF handshake

evan-masseau:
# Description
See handshake section of the native bridge spec doc.

# Check List


sdk/messaging/src/test/java/com/klaviyo/messaging/FormsUtilityTest.kt
```diff
@@ -166,7 +166,9 @@ class FormsUtilityTest : BaseTest() {
         """.trimIndent()
 
         val expectedAggBody =
-            JSONObject("{\"metric_group\":\"signup-forms\",\"events\":[{\"log_to_metrics_service\":true,\"metric\":\"stepSubmit\",\"log_to_statsd\":true,\"event_details\":{\"page_url\":\"http://localhost:4001/onsite/js/\",\"first_referrer\":\"http://localhost:4001/onsite/js/\",\"action_type\":\"Submit Step\",\"form_version_id\":3,\"form_id\":\"64CjgW\",\"device_type\":\"DESKTOP\",\"form_type\":\"POPUP\",\"referrer\":\"http://localhost:4001/onsite/js/\",\"submitted_fields\":{\"sms_consent\":true,\"consent_method\":\"Klaviyo Form\",\"consent_form_version\":3,\"step_name\":\"Email Opt-In\",\"consent_form_id\":\"64CjgW\",\"source\":\"Local Form\",\"email\":\"local@local.com\",\"sent_identifiers\":{}},\"hostname\":\"localhost\",\"step_number\":1,\"form_version_c_id\":\"1\",\"step_name\":\"Email Opt-In\",\"is_client\":true,\"href\":\"http://localhost:4001/onsite/js/\",\"cid\":\"ODZjYjJmMjUtNjliMC00ZGVlLTllM2YtNDY5YTlmNjcwYmUz\"},\"metric_service_event_name\":\"submitted_form_step\",\"log_to_s3\":true}]}")
```
evan-masseau, line 169: This must have been ktlint

sdk/messaging/src/test/java/com/klaviyo/messaging/InAppFormsHtmlTests.kt
```diff
@@ -0,0 +1,41 @@
+package com.klaviyo.messaging
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class InAppFormsHtmlTests {
+    @Test
+    fun `Verify IAF handshake`() {
```
evan-masseau, line 8: A dumb test, but tests like this serve as extra validation that we don't mistakenly change important static data

versions.properties
```diff
@@ -9,85 +9,50 @@
 ####
 #### NOTE: Some versions are filtered by the rejectVersionIf predicate. See the settings.gradle.kts file.
 
-# Project versioning, run the following gradle command to update version numbers automatically:
```
evan-masseau, line 12: Unrelated but has been bothering me: I organized and alphabetized the versions, and I added the _version_ property for `version.androidx.compose.compiler=1.5.10` that's always causing annoyance in composite test app builds. This does NOT add compose as a dependency, only sets a version property for when the SDK + test app are used in composite mode.

versions.properties
```diff
@@ -9,85 +9,50 @@
...
 # Project dependencies
-
-version.kotlinx.coroutines=1.9.0
-
-version.org.jetbrains.dokka..versioning-plugin=1.9.10
-##                                 # available=1.9.20
-
+version.androidx.activity=1.6.1
+version.androidx.appcompat=1.6.1
+version.androidx.compose.compiler=1.5.10
```
evan-masseau, line 32: This does NOT add compose as a dependency, only sets a version property for when the SDK + test app are used in composite mode.

versions.properties
evan-masseau: I really shouldn't cross purposes of PRs, i'll pull this out

sdk/messaging/src/main/java/com/klaviyo/messaging/Utils.kt
```diff
@@ -50,6 +50,11 @@ internal fun decodeWebviewMessage(webMessage: String): KlaviyoWebFormMessageType
                 )
             }
         }
+        IAF_MESSAGE_TYPE_ABORT -> {
```
evan-masseau, line 53: Wanted to be able to test aborting while I'm implementing the fender side, so i threw this in, doesn't have to be the permanent implementation

sdk/messaging/src/main/assets/InAppFormsTemplate.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+<head data-sdk-name="SDK_NAME"
+      data-sdk-version="SDK_VERSION"
+      data-native-bridge-name="BRIDGE_NAME"
+      data-native-bridge-handshake='BRIDGE_HANDSHAKE'
+>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Klaviyo In-App Form Test</title>
```
kennyklaviyo, line 10: name to `Form Template`?

sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
```diff
@@ -95,6 +95,9 @@ object KlaviyoConfig : Config {
 
     override var baseUrl: String = BuildConfig.KLAVIYO_SERVER_URL
         private set
+
```
evan-masseau, line 98: We will need to put a few setters in the builder class as well.

dan-peluso: Do we have a ticket for that or will it be part of this PR?

evan-masseau: It could be done as part of the ticket adding test-app support for being able to set this.

sdk/messaging/src/main/java/com/klaviyo/messaging/InAppMessaging.kt
```diff
@@ -10,14 +10,17 @@ object InAppMessaging {
     fun triggerInAppMessage(activity: Activity) {
         val rootView = activity.getRootViewGroup()
         val webView = KlaviyoWebView(activity)
+        val klaviyoJsUrl = "${Registry.config.baseCdnUrl}/onsite/js/klaviyo.js?env=in-app&company_id=${Registry.config.apiKey}"
```
dan-peluso, line 13: nice

sdk/messaging/src/test/java/com/klaviyo/messaging/InAppFormsHtmlTests.kt
dan-peluso: No test is a dumb test!

kennyklaviyo (approved):

PR #217, opened 2025-02-05
Organized the versions.properties file for better readability 

evan-masseau:
and added a compose.compiler version (not dependency) to alleviate local dev headaches.

# Description
I organized and alphabetized the versions, and I added the version property for `version.androidx.compose.compiler=1.5.10` that's always causing annoyance in composite test app builds.

There's no version updates in here. I did take out the commented future versions tho, I find those make it harder to read and don't need to be committed (they're useful when you're working on upgrading packages, but as a general rule I prefer to delete them before committing)

I would like to merge this prior to rebasing the IAF feature branch onto master to catch up (something I intend to do soon)

# Check List

### DIFF
--- a/versions.properties
+++ b/versions.properties
@@ -9,56 +9,50 @@
 ####
 #### NOTE: Some versions are filtered by the rejectVersionIf predicate. See the settings.gradle.kts file.
 
-# Project versioning, run the following gradle command to update version numbers automatically:
-# ./gradlew bumpVersion --nextVersion=X.Y.Z
+# Project versioning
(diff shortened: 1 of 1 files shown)


evan-masseau:
Not the most important thing going right now, but it was buggin me.

dan-peluso (approved):
RIP changing the compose compiler version for every build I did locally, good times were had

evan-masseau:
> RIP changing the compose compiler version for every build I did locally, good times were had

Every freakin' time.

PR #219, opened 2025-02-07
Exposing kv pairs properties

jasper-hsu-kyvo:
# Description

Added two new extension properties to `RemoteMessage`s:

1. `hasKlaviyoKeyValuePairs` which is a boolean (`true` if message payloads contain kv-pairs, `false` otherwise)
2. `keyValuePairs` optional `Map<String,String>` which is the dictionary of kv-pairs itself (deserialized from json)

Also updated the README to reflect these changes.

# Check List

No.
- [X] Are your changes backwards compatible with previous SDK Versions?
- [X] Have you tested this change on real device?
- [X] Have you added unit test coverage for your changes?
How can I do this?

## Changelog / Code Overview

- Added extension properties for key-value pairs
- Added new branching logic for silent pushes, and an overridable handler method

## Test Plan

- Tested on the test app - made standard and silent pushes to it, with and without key-value pairs, and then would use extension properties to access the data.
- Also ensured that the log was being made (for the new conditional logic using the new `onSilentPushMessageReceived` method)

### DIFF
--- a/README.md
+++ b/README.md
@@ -374,6 +374,9 @@ deep links in your app.
     [preview push notification](https://help.klaviyo.com/hc/en-us/articles/18011985278875) 
     containing a deep link from the Klaviyo push editor. 
 
+#### Custom Data
+Klaviyo messages can also include custom key-value pairs (custom data) for both standard and silent push notifications. You can access these key-value pairs using the extension property `RemoteMessage.keyValuePairs` and check for their presence with the boolean extension property `RemoteMessage.hasKlaviyoKeyValuePairs`. This enables you to extract additional information from the push payload and handle it appropriately - for instance, but triggering background processing, logging analytical events, or dynamically updating app content.
+
 For additional resources on deep linking, refer to 
 [Android developer documentation](https://developer.android.com/training/app-links/deep-linking).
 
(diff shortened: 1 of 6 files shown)


sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoRemoteMessage.kt
```diff
@@ -135,6 +136,32 @@ object KlaviyoRemoteMessage {
     val RemoteMessage.notificationCount: Int
         get() = this.data[KlaviyoNotification.NOTIFICATION_COUNT_KEY]?.toInt() ?: 1
 
+    /**
+     * Determine if the message is bearing key-value pairs
+     */
+    val RemoteMessage.hasKlaviyoKeyValuePairs: Boolean
+        get() = this.isKlaviyoMessage && (this.data[KlaviyoNotification.KEY_VALUE_PAIRS_KEY]?.let { true } ?: false)
```
evan-masseau, line 143: `this.data.containsKey"KlaviyoNotification.KEY_VALUE_PAIRS_KEY")` probs cleaner.

jasper-hsu-kyvo: Wouldn't I have to wrap everything in an if/else then too? (the let statement no longer works)

evan-masseau: don't need the let i mean, it should just return a bool of whether that key is present

evan-masseau: same as `val RemoteMessage.isKlaviyoMessage: Boolean get() = this.data.containsKey("_k")` elsewhere in this file

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoRemoteMessage.kt
```diff
@@ -135,6 +136,32 @@ object KlaviyoRemoteMessage {
...
+
+    /**
+     * Parse out the key-value pairs into a string:string map
+     */
+    val RemoteMessage.keyValuePairs: Map<String,String>?
```
evan-masseau, line 148: Looks like we have a gap in our test coverage, this file has no corresponding test file. Would you mind adding a file in the same dir as `KlaviyoPushServiceTest` to cover your changes, and we'll backfill tests for other stuff in the future.

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoPushService.kt
```diff
@@ -43,10 +44,33 @@ open class KlaviyoPushService : FirebaseMessagingService() {
     override fun onMessageReceived(message: RemoteMessage) {
         super.onMessageReceived(message)
 
-        if (message.isKlaviyoNotification) {
-            KlaviyoNotification(message).displayNotification(this)
+        if (message.isKlaviyoMessage) {
+            if (message.isKlaviyoNotification) {
+                KlaviyoNotification(message).displayNotification(this)
+            } else {
```
evan-masseau, line 50: Ooh wait, correct me if I'm wrong, but a notification could also be bearing key/value pairs right? So this logic can't be exclusively if/else? Rather it'd need to be (pseudocode)
```
if (message.isKlaviyoNotification){
  //display
}
if (message.hasKeyValuePairs) {
  //handle k/v pairs
}
``` 
?

evan-masseau: Also, if that isn't the case _now_ but it _could be_ the case in the future, we should still design the SDK for that eventuality i believe.

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoPushService.kt
```diff
@@ -43,10 +44,33 @@ open class KlaviyoPushService : FirebaseMessagingService() {
...
+                // Handle silent push notifications
+                onSilentPushMessageReceived(message)
+            }
         } else {
             Registry.log.info("Ignoring non-Klaviyo RemoteMessage")
         }
     }
+
+    /**
+     * Called when a silent push notification is received.
```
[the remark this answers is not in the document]
jasper-hsu-kyvo: README changes included in this PR

dan-peluso (approved):
King kotlin!

README.md
```diff
@@ -417,47 +420,65 @@ You may either subclass `KlaviyoPushService` or invoke the necessary Klaviyo SDK
...
-            if (!message.isKlaviyoNotification) {
-                // Handle non-Klaviyo messages
+            if (!message.isKlaviyoMessage) {
+                TODO("Handle non-Klaviyo messages")
             }
         }
-    }
-    ```
-
-2. Subclass `FirebaseMessagingService` and invoke Klaviyo SDK methods directly
```
evan-masseau, line 427: total nit, but did this whole `2.` section get mistakenly indented?

evan-masseau (approved):

PR #224, opened 2025-02-21
Removed silent push and custom data from README

jasper-hsu-kyvo:
# Description

# Check List


njparadis (approved):
Are these features not making it into the next release?

evan-masseau (approved):

evan-masseau:
> Are these features not making it into the next release?

@njparadis correct, we're leaving in the code, omitting the documentation

PR #225, opened 2025-02-21
IAF Feature Branch

evan-masseau:
# Description
Feature branch for all in-app forms work. As we approach release, this is a good place to leave overall comments and review the whole module!

# Check List


sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
```diff
@@ -95,6 +95,9 @@ object KlaviyoConfig : Config {
 
     override var baseUrl: String = BuildConfig.KLAVIYO_SERVER_URL
         private set
+
+    override var baseCdnUrl: String = BuildConfig.KLAVIYO_CDN_URL
```
evan-masseau, line 99: TODO: add to config builder to test app can set it.

sdk/analytics/src/main/java/com/klaviyo/analytics/DeviceProperties.kt
```diff
@@ -11,60 +11,60 @@ import com.klaviyo.core.config.getPackageInfoCompat
 import com.klaviyo.core.model.fetchOrCreate
 import java.util.UUID
 
-internal object DeviceProperties {
+object DeviceProperties {
```
jordan-griffin, line 14: total nit: Was this internal object made accessible just so we could `userAgent`? I guess I see why we popped user agent in here since its constructed from so many values tied directly to this class but just had me wondering if this should be moved elsewhere out of the object

README.md
```diff
@@ -55,8 +55,8 @@ send them timely push notifications via [FCM (Firebase Cloud Messaging)](https:/
       ```kotlin
       // build.gradle.kts
       dependencies {
-          implementation("com.github.klaviyo.klaviyo-android-sdk:analytics:3.0.3")
-          implementation("com.github.klaviyo.klaviyo-android-sdk:push-fcm:3.0.3")
+          implementation("com.github.klaviyo.klaviyo-android-sdk:analytics:3.2.0-alpha.2")
```
ajaysubra, line 58: do we want to replace with actual release versions or is that a separate PR?

sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
```diff
@@ -2,21 +2,26 @@ package com.klaviyo.core.lifecycle
...
-internal object KlaviyoLifecycleMonitor : LifecycleMonitor, Application.ActivityLifecycleCallbacks {
+internal object KlaviyoLifecycleMonitor : LifecycleMonitor, Application.ActivityLifecycleCallbacks, ComponentCallbacks {
 
     private var activeActivities = 0
 
     private val activityObservers = Collections.synchronizedList(
         mutableListOf<ActivityObserver>()
     )
 
+    override var currentActivity: Activity? by WeakReferenceDelegate()
```
jordan-griffin, line 23: nit: Is a weak reference even necessary? Seems we're clearing out the activity on pause/stop just fine. I get a weak reference isnt enough on its own cause we cant guarantee garbage collection will run in time but I'm just a little confused on if we need this reference to be any different given the logic below.

Honestly it likely affects nothing so its whatever

sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
```diff
@@ -70,5 +84,13 @@ internal object KlaviyoLifecycleMonitor : LifecycleMonitor, Application.Activity
         // Warning: onActivityDestroyed is unreliable, I'm not even going to try to broadcast it
     }
 
+    override fun onConfigurationChanged(newConfig: Configuration) {
+        broadcastEvent(ActivityEvent.ConfigurationChanged(newConfig))
+    }
+
+    override fun onLowMemory() {
```
jordan-griffin, line 91: What was this intended for?

sdk/forms/src/main/assets/InAppFormsTemplate.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html lang="en">
+<head data-sdk-name="SDK_NAME"
+      data-sdk-version="SDK_VERSION"
+      data-native-bridge-name="BRIDGE_NAME"
+      data-native-bridge-handshake='BRIDGE_HANDSHAKE'
+>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0, viewport-fit=cover"/>
+    <meta name="referrer" content="same-origin" />
```
ajaysubra, line 10: would adding a comment on the context of why this is a here a good idea?

sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
```diff
@@ -70,5 +84,13 @@ internal object KlaviyoLifecycleMonitor : LifecycleMonitor, Application.Activity
         // Warning: onActivityDestroyed is unreliable, I'm not even going to try to broadcast it
     }
 
+    override fun onConfigurationChanged(newConfig: Configuration) {
```
jordan-griffin, line 87: nit: Not technically an activity lifecycle callback (right?), should we move this below the `//endregion` comment that denotes our block for activity lifecycle callbacks in this object?

README.md
jordan-griffin, line 58: I assume we'll need to bump this to a non-alpha release version before we merge everything in?

ajaysubra (approved):

sdk/forms/src/main/java/com/klaviyo/forms/BridgeMessage.kt
```diff
@@ -0,0 +1,163 @@
...
+
+                IAF_MESSAGE_TYPE_SHOW -> Show(
+                    formId = jsonData.optString("formId").ifEmpty { "" }
+                )
+
+                IAF_MESSAGE_TYPE_AGGREGATE_EVENT -> AggregateEventTracked(
+                    payload = jsonData
+                )
+
+                IAF_MESSAGE_TYPE_PROFILE_EVENT -> ProfileEvent(
```
jordan-griffin, line 113: Profile events werent in the MVP right? With this code bridged, whats the current functionality whenever a profile event is triggered? Or is it that the onsiteJS bridge is where the profile-level event handling is missing so this will never fire for now?

sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebView.kt
```diff
@@ -0,0 +1,62 @@
...
+    ) : super(context) {
+        settings.userAgentString = DeviceProperties.userAgent
+        settings.javaScriptEnabled = true
+        settings.domStorageEnabled = true
+        setBackgroundColor(Color.TRANSPARENT)
+
+        if (BuildConfig.DEBUG) {
+            setWebContentsDebuggingEnabled(BuildConfig.DEBUG)
+            // Disable webview resources cache when debugging:
+            // settings.cacheMode = WebSettings.LOAD_NO_CACHE
```
jordan-griffin, line 31: nit: should we control through some env vars or something instead of just leaving commented out?

sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebViewDelegate.kt
```diff
@@ -0,0 +1,244 @@
...
+    /**
+     * Initialize a webview instance, with protection against duplication
+     * and initialize klaviyo.js for in-app forms with handshake data injected in the document head
+     */
+    fun initializeWebView() = webView ?: KlaviyoWebView().also { webView ->
+        this.webView = webView
+
+        val klaviyoJsUrl = Uri.parse(Registry.config.baseCdnUrl)
+            .buildUpon()
+            .path("onsite/js/klaviyo.js")
```
jordan-griffin, line 80: the biggest of all nits: make this a const? Similar to how we treat paths in our request classes as part of the analytics module

sdk/analytics/src/main/java/com/klaviyo/analytics/DeviceProperties.kt
evan-masseau: Good point. Might also be able to do this via registry

jordan-griffin (approved):
Looks great. Took a nice long read, everything I left were nits to consider but I dont think anything was blocking. Just make sure we have the SDK sem version correct

sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
evan-masseau: Agreed, but basically just erring on the side of caution

sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
evan-masseau: I think it's part of the configuration change interface

README.md
evan-masseau: Yep, I expect to build at least one more alpha SDK/test app, then we'll bump version to official 3.2.0

evan-masseau: We can do it in this PR, we'll have a few more last minute additions though, like version bump, readme, possibly bug fixes from our next test app build

sdk/forms/src/main/assets/InAppFormsTemplate.html
```diff
@@ -0,0 +1,16 @@
...
+    <title>Klaviyo In-App Form Template</title>
+    <link rel="stylesheet" type="text/css" href="https://static-forms.klaviyo.com/fonts/api/v1/in-app-web-fonts/websafe_fonts.css" crossorigin/>
```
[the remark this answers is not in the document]
evan-masseau: I didn't bother bc it doesn't require any variables or anything dynamic

sdk/forms/src/main/java/com/klaviyo/forms/BridgeMessage.kt
evan-masseau: IAF editor only supports close and deep link, neither of which trigger a profile event yet. But by supporting it in the SDK we leave the option open for a fast follow of adding some profile events via a fender update.

sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebView.kt
evan-masseau: I had it controlled thru the DEBUG build flag, but depending on what you're doing, sometimes the cache is useful even if debugging. So I just left this as a breadcrumb to remember it's a tool available to us. I am fine with deleting it before release tho

sdk/forms/src/main/java/com/klaviyo/forms/InAppForms.kt
```diff
@@ -0,0 +1,18 @@
+package com.klaviyo.forms
+
+import com.klaviyo.analytics.Klaviyo
+import com.klaviyo.core.Registry
+import com.klaviyo.core.safeApply
+
+/**
+ * Initialize in-app forms listeners
```
evan-masseau, line 8: TODO we should update this comment block when we finish drafting the README.

sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebViewDelegate.kt
jordan-griffin: Thats fair!

sdk/analytics/src/main/java/com/klaviyo/analytics/DeviceProperties.kt
jordan-griffin: Ha thats a good point, but honestly a better solution than what I was thinking when I commented. Nice!

sdk/forms/src/main/java/com/klaviyo/forms/BridgeMessage.kt
jordan-griffin: Ah perfect thanks for clarifying that.

jordan-griffin: Only concern is once we support profile level events in the future, are we confident that the logic here would lead to some bugs popping up in this upcoming SDK release? Like the code for supporting profile level events is there now so its forward thinking, but we've no guarantee that it'll work and it may be more confusing for users if that was the case (even if we did not advertise it as a feature for this version)

evan-masseau: Oh no, we've already tested profile event creation e2e. I made a fender branch with a profile level event triggered by deep link, asset-sourced that, and verified that the profile event made it to the backend. 
FWIW if we did find a bug like that, we would be able to use our handshake data to selectively exclude a buggy SDK implementation for even one message bridge type!

evan-masseau:
@dan-peluso care to throw your approval on here too so we can all be to blame when i merge this :D

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/AggregateEventApiRequest.kt
```diff
@@ -0,0 +1,26 @@
...
+    companion object {
+        private const val PATH = "onsite/track-analytics"
+    }
+
+    override val type: String = "Create Aggregate Event"
+    override var query: Map<String, String> = mapOf(
+        COMPANY_ID to Registry.config.apiKey
+    )
+
+    override val successCodes: IntRange get() = HTTP_ACCEPTED..HTTP_ACCEPTED
```
dan-peluso, line 21: late stage ask ik but just want to be sure our endpoint is truly this specific regarding the success codes, wouldn't want to retry something we don't need to

evan-masseau: yep, 202 is the "accepted" status code, our public APIs use this because the data is validated synchronously, but then inserted into queues to do full processing async.

evan-masseau: We could probs be more lax, like if the server did respond with a 200 instead of 202. But I think we'd just log an error message and otherwise do the same thing as a success -- clear the request from the queue and move on to the next.

sdk/core/src/main/java/com/klaviyo/core/DeviceProperties.kt
```diff
@@ -71,54 +70,25 @@ internal object DeviceProperties {
     }
 
     val userAgent: String by lazy {
-        val sdkAgent = "klaviyo-${sdkName.replace("_","-")}"
+        val sdkAgent = "klaviyo-${sdkName.replace("_", "-")}"
         "$applicationLabel/$appVersion ($applicationId; build:$appVersionCode; $platform $osVersion) $sdkAgent/$sdkVersion"
     }
 
+    val environment: String by lazy {
```
dan-peluso, line 77: Must have missed this one get added earlier - I'm a little concerned we have too many sources for what build flavor we're using. I can think of 3 off the top of my head:

1. DeviceProperties.environment
2. BuildConfig(.debug == true)
3. Registry.config.isDebugBuild

Additionally I feel like the environment is less of a 'device' property and more a build/config thing. Any thoughts?

sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
dan-peluso: did we ever do this?

sdk/forms/src/main/assets/InAppFormsTemplate.html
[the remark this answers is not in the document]
dan-peluso: praying this doesn't jinx us and we need to change the bucket path or something lol

sdk/core/src/main/java/com/klaviyo/core/DeviceProperties.kt
evan-masseau: I'm pretty sure we used `Registry.config.applicationContext.applicationInfo.flags and ApplicationInfo.FLAG_DEBUGGABLE` to make sure that we were looking at the host app's build info, not the SDK's. This is pre-existing, just turned it into a property for consistency (see below). It is used to detect the host app's environment for app metadata that we track on the backend. Its been useful in the past for support requests, although its less important on android. On iOS there are different push notification sending endpoints depending on environment, sandbox and production. 

As for `Registry.config.isDebugBuild` and `BuildConfig.DEBUG`... I do prefer not to access `BuildConfig.DEBUG` directly, opting instead for using `Registry.config.isDebugBuild` but its not the end of the world to me

sdk/forms/src/main/assets/InAppFormsTemplate.html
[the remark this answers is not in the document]
evan-masseau: meh it'd just 404 silently in the background

versions.properties
```diff
@@ -13,7 +13,7 @@
 #  NOTE: semantic version of the SDK lives in strings.xml
 #  run the following gradle command to update version numbers automatically:
 #  ./gradlew bumpVersion --nextVersion=X.Y.Z
-version.klaviyo.versionCode=22
+version.klaviyo.versionCode=28
```
dan-peluso, line 16: lol

versions.properties
```diff
@@ -47,6 +47,8 @@ version.kotlin=1.9.22
 
 version.kotlinx.coroutines=1.9.0
 
+version.leakcanary=2.14
```
dan-peluso, line 50: 🔥

evan-masseau: 🤫

sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
evan-masseau: yes, its below:
```
        override fun baseCdnUrl(baseCdnUrl: String): Config.Builder = apply {
            this.baseCdnUrl = baseCdnUrl
        }
```

sdk/core/src/main/java/com/klaviyo/core/DeviceProperties.kt
dan-peluso: That makes sense, thanks for clarifying

dan-peluso (approved):
Nothing blocking from me
:shipit:

sdk/core/src/main/java/com/klaviyo/core/DeviceProperties.kt
evan-masseau: you actually already found the reason that I prefer not to use `BuildConfig.DEBUG` directly -- because its more annoying to mock in tests. So I think your tests PR already fixes some uses.

sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
dan-peluso: oh sorry I meant added to the test app

evan-masseau: oh lol that makes a lot more sense -- yes i updated the test app's settings page last week to have inputs for cdn url and assetSource!

sdk/core/src/main/java/com/klaviyo/core/DeviceProperties.kt
dan-peluso: Yeah I'm down to avoid the BuildConfig directly, but I was more confused about this one vs the isDebugEnabled call. But the underlying app vs the sdk makes sense, sometimes I forget we live in SDK land

PR #229, opened 2025-02-24
Tweak Weak References usage

evan-masseau:
# Description
- New utility to use `WeakReference` as a delegated property
- Added a little more safety to how WebView is referenced to avoid mistakenly reinstantiating. 

# Check List

### DIFF
--- a/sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebViewDelegate.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebViewDelegate.kt
@@ -122,26 +122,34 @@ internal class KlaviyoWebViewDelegate : WebViewClient(), WebViewCompat.WebMessag
     }
 
     private fun show() = webView?.let { webView ->
-        activity?.let {
-            it.window.decorView
-                .findViewById<ViewGroup>(android.R.id.content)
-                .addView(webView)
-            webView.post { webView.visibility = View.VISIBLE }
+        activity?.window?.decorView?.let { decorView ->
+            decorView.post {
+                decorView.findViewById<ViewGroup>(android.R.id.content).addView(webView)
+                webView.visibility = View.VISIBLE
+            }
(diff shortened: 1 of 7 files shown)


sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebViewDelegate.kt
```diff
@@ -36,33 +36,32 @@ internal class KlaviyoWebViewDelegate() : WebViewClient(), WebViewCompat.WebMess
...
+
     init {
         WebView.setWebContentsDebuggingEnabled(BuildConfig.DEBUG)
     }
 
-    private var internalWebview: WeakReference<WebView> = WeakReference(null)
-
-    @get:SuppressLint("SetJavaScriptEnabled")
-    private val webView: WebView
-        get() = internalWebview.get() ?: WebView(Registry.config.applicationContext).also {
```
evan-masseau, line 47: `internalWebview.get() ?: WebView(Registry.config.applicationContext)` using this raised a mild concern that if internalWebview was GC'd we could instantiate a new webview at an unintentional moment.

dan-peluso: All here for the extra safety, but I'm not sure I fully understand what could lead to the internal webview being garbage collected before we'd want it to be.

evan-masseau: Then lets do it for the safety.

evan-masseau: Its not that I've observed an issue, but the pattern of defaulting to creating a new instance even when accessing this property from places where we _expect_ to have an existing instance already could lead to unintended consequences.

dan-peluso (approved):
LGTM

PR #230, opened 2025-02-25
Close on orientation change

dan-peluso:
# Description
- we attach the webview to the current screen on the show message, but if a new activity is launched (like on a config change) that screen is lost in the ether
- we need to change on not just a config change, but specifically when the orientation is updated
- because that callback gets emitted when the app first started, we locally store the orientation and only close the form if the new orientation is different and the forms is already visible to the user

# Check List

### DIFF
--- a/sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebViewDelegate.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebViewDelegate.kt
@@ -12,7 +12,6 @@ import android.webkit.WebResourceResponse
 import android.webkit.WebView
 import android.webkit.WebViewClient
 import androidx.core.content.ContextCompat.startActivity
-import androidx.core.view.isVisible
 import androidx.webkit.JavaScriptReplyProxy
 import androidx.webkit.WebMessageCompat
 import androidx.webkit.WebViewCompat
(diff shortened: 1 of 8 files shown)


evan-masseau (approved):
Let's go for it, I'll test it out tomorrow but feel free to merge to feat branch

sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebViewDelegate.kt
```diff
@@ -38,8 +40,28 @@ internal class KlaviyoWebViewDelegate : WebViewClient(), WebViewCompat.WebMessag
 
     private var webView: WebView? by WeakReferenceDelegate()
 
+    private var orientation: Int? = null
```
evan-masseau, line 43: I think there's a flaw in the logic here: initializing with `null`, in combination with guarding for null below `orientation?.let { oldOrientation` means on the first rotation the webview doesn't get destroyed (still disappears of course)

dan-peluso: Thanks for patching - yeah I was noticing an initial emission of config change when I would open the app, which is why I first set it to null and then overwrite. Any reason for removing the `isVisible` check? Thought it was a decent extra precaution

PR #231, opened 2025-02-26
mistakenly have the same enum in two places!

evan-masseau:
# Description

# Check List


ajaysubra:
would have never happened on iOS since we don't have namespace!

ajaysubra (approved):

PR #233, opened 2025-02-26
IAF - Structural refactors

evan-masseau:
# Description
Took an hour to go through all the IAF code and applied a lot of refactors to tidy up. I think a lot of this developed ad-hoc and it was due for a higher level review. This contains only a few minor functional changes, on the whole its just code organization / refactors. 

# Check List


sdk/forms/src/main/java/com/klaviyo/forms/BridgeMessage.kt
```diff
@@ -0,0 +1,163 @@
+package com.klaviyo.forms
```
evan-masseau, line 1: Renamed this to BridgeMessage, more in line with how our naming conventions have evolved. And I merged the Utils into this file as a companion factory method.

sdk/forms/src/main/java/com/klaviyo/forms/BridgeMessage.kt
```diff
@@ -0,0 +1,163 @@
...
+    data class Close(
+        val formId: String
+    ) : BridgeMessage()
+
+    data class Abort(
+        val reason: String
+    ) : BridgeMessage()
+
+    companion object {
+        private const val IAF_MESSAGE_DATA_KEY = "data"
```
evan-masseau, line 43: I moved everything out of `Constants` file and eliminated single-use constants. Generally best to avoid constants files -- [reference for anyone interested](https://www.hacklewayne.com/the-constants-file-is-an-anti-pattern-so-is-the-interface-folder-placement-by-functionality-not-technical-concerns)

sdk/forms/src/main/java/com/klaviyo/forms/InAppForms.kt
```diff
@@ -3,24 +3,16 @@ package com.klaviyo.forms
...
+/**
+ * Initialize in-app forms listeners
+ * Once registered, an in-app form may appear within 10s
+ */
 fun Klaviyo.registerForInAppForms(): Klaviyo = safeApply {
+    // Ensure we only ever register one delegate instance
     if (!Registry.isRegistered<KlaviyoWebViewDelegate>()) {
         Registry.register<KlaviyoWebViewDelegate>(KlaviyoWebViewDelegate())
     }
-    val klaviyoJsUrl =
```
evan-masseau, line 12: Felt odd to house this logic in the top level API method

sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebView.kt
```diff
@@ -0,0 +1,62 @@
+package com.klaviyo.forms
```
evan-masseau, line 1: Extracted WebView-specific stuff from the Delegate class.

sdk/forms/src/main/java/com/klaviyo/forms/InAppForms.kt
```diff
@@ -3,24 +3,16 @@ package com.klaviyo.forms
 import com.klaviyo.analytics.Klaviyo
 import com.klaviyo.core.Registry
 import com.klaviyo.core.safeApply
-import java.io.BufferedReader
 
+/**
+ * Initialize in-app forms listeners
```
evan-masseau, line 8: We'll update this more when we draft the readme.

sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebViewDelegate.kt
```diff
@@ -59,102 +68,159 @@ internal class KlaviyoWebViewDelegate : WebViewClient(), WebViewCompat.WebMessag
...
+            Registry.log.info("Redirect URL to external browser: ${request.url}")
             val intent = Intent(Intent.ACTION_VIEW).apply {
                 data = request.url
                 addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
             }
             activity?.let {
                 startActivity(it, intent, null)
             } ?: run {
-                Registry.log.error("Unable to launch external browser, null activity reference")
+                Registry.log.warning("Unable to launch external browser - null activity reference")
```
evan-masseau, line 130: pretty trivial, but `error` and above get output from production builds, so I try to stay at warning or lower unless there's a compelling reason.

sdk/forms/src/main/java/com/klaviyo/forms/BridgeMessage.kt
```diff
@@ -0,0 +1,163 @@
...
+
+        /**
+         * Parse [Event] properties for a [ProfileEvent] message
+         */
+        private fun JSONObject.getEventProperties(): Map<EventKey, Serializable> {
+            val map = mutableMapOf<EventKey, Serializable>()
+            val propertyMap = getJSONObject("properties")
+            propertyMap.keys().forEach {
+                try {
+                    map[EventKey.CUSTOM(it)] = propertyMap.get(it) as Serializable
```
evan-masseau, line 144: This is one small functional change that I made here: this was parsing only strings, we should be able to support any `Serializable` (and if it somehow isn't serializable, this will throw)

ajaysubra (approved):

PR #235, opened 2025-03-03
3.2.0 - In-App Forms Readme Updates

evan-masseau:
# Description

# Check List

### DIFF
--- a/README.md
+++ b/README.md
@@ -472,7 +472,8 @@ notification intent that the Klaviyo SDK requires in order to track opens when y
 
 [In-app forms](https://klaviyo.slack.com/archives/C064QGU3BCG/p1741024810256069) are messages displayed to mobile app 
 users while they are actively using your app. You can create new in-app forms in a drag-and-drop editor in the 
-Sign-Up Forms tab in Klaviyo.
+Sign-Up Forms tab in Klaviyo. Follow the instructions in this section to integrate forms with your app. The SDK will
+display forms according to targeting and behavior settings and collect delivery and engagement analytics automatically.
 
 ### Prerequisites
 - Version 3.2.0 or higher
--- a/README.md
+++ b/README.md
@@ -503,15 +503,16 @@ You can call `registerForInAppForms()` any time after initializing with your com
 your app's UI a form can appear. It is safe to register multiple times per application session. 
 The SDK will internally prevent multiple forms appearing at once.
 
-Consider how often you want to register for forms. For example, registering from an activity lifecycle event is advisable 
+Consider how often you want to register for forms. For example, registering from a lifecycle event is advisable 
 so that the user has multiple opportunities to see your messaging if they are browsing your app for a prolonged period.
 However, be advised the form will be shown as soon as it is ready, so you may still need to condition this based on the
 user's context within your application. Future versions of this product will provide more control in this regard.
 
 | Callback                 | Description                                                                                            |
 |--------------------------|--------------------------------------------------------------------------------------------------------|
-| `Activity.onCreate()`    | If you only want forms to appear over a specific activity, checks every time this activity is created. |
+| `Application.onCreate()` | Show a form _only_ upon initial app launch.                                                            |
 | `Application.onResume()` | Anytime the app is foregrounded, check for forms and show if available.                                |
+| `Activity.onCreate()`    | If you only want forms to appear over a specific activity, checks every time this activity is created. |
 
 **Note:** At this time, when device orientation changes any currently visible form is closed and will not be re-displayed automatically.
 
--- a/README.md
+++ b/README.md
@@ -465,6 +465,37 @@ extensions such as `import com.klaviyo.pushFcm.KlaviyoRemoteMessage.body` to acc
 We also provide an `Intent.appendKlaviyoExtras(RemoteMessage)` extension method, which attaches the data to your
 notification intent that the Klaviyo SDK requires in order to track opens when you call `Klaviyo.handlePush(intent)`.
 
+## In App Forms
+
+### Prerequisites
+- Using version 3.2.0 or higher
+- You need to have Klaviyo `analytics` imported and the company ID must be set
+
+### Setup
+`registerForInAppForms` can be chained with your initialization code to allow for form display.
+```kotlin
+Klaviyo
+    .initialize("KLAVIYO_PUBLIC_API_KEY", applicationContext)
+    .registerForInAppForms()
+```
+You can also call it publicly anywhere you can access the Klaviyo forms module:
+```kotlin
+Klaviyo.registerForInAppForms() // make sure you only call this after initializing Klaviyo company ID
+```
+
+### Behavior
+
+Once you call the `registerForInAppForms` function, the web view will persist unattached to UI until a form is ready to be shown.
+This allows you to choose *when* forms will start to overlay on your App UI. It also means that wherever you decide to call this will impact the behavior.
+Consider how often you want to register for forms when you call this, for example:
+* `Activity.onCreate()` -> If you only want forms to appear over a specific activity, checks every time this activity is created.
+* `Application.onResume()` -> Anytime the app is foregrounded, check for forms and show if available.
+
+The web view will listen for forms to display with the same timeout as other requests (10 seconds).
+If no form is available to show, the web view will be removed from memory and there will need to be another call to `registerForInAppForms`
+to fetch available form. This is why we recommend registering on a lifecycle function. Forms will show automatically, and are dismissed by pressing the close button or tapping outside the content of the dialog.
+If an orientation configuration change is detected, we close the web view and remove it from memory.
+
 ## Troubleshooting
 The SDK contains logging at different levels from `verbose` to `assert`. By default, the SDK logs at the `error` level
 in a production environment and at the `warning` level in a debug environment. You can change the log level by adding 
--- a/README.md
+++ b/README.md
@@ -404,7 +404,7 @@ notification intent that the Klaviyo SDK requires in order to track opens when y
 [In-app forms](https://help.klaviyo.com/hc/en-us/articles/34567685177883) are messages displayed to mobile app 
 users while they are actively using your app. You can create new in-app forms in a drag-and-drop editor in the 
 Sign-Up Forms tab in Klaviyo. Follow the instructions in this section to integrate forms with your app. The SDK will
-display forms according to targeting and behavior settings and collect delivery and engagement analytics automatically.
+display forms according to their targeting and behavior settings and collect delivery and engagement analytics automatically.
 
 ### Prerequisites
 - Version 3.2.0 or higher
@@ -430,7 +430,7 @@ Klaviyo.registerForInAppForms()
 ### Behavior
 
 Once `registerForInAppForms()` is called, the SDK will load form data for your account and display no more than one form 
-within 10 seconds, based on form targeting and behavior settings.
+within 15 seconds, based on form targeting and behavior settings.
 
 You can call `registerForInAppForms()` any time after initializing with your company ID to control when and where in 
 your app's UI a form can appear. It is safe to register multiple times per application session. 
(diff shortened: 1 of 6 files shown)


sdk/forms/src/main/java/com/klaviyo/forms/InAppForms.kt
```diff
@@ -1,13 +1,15 @@
...
 import com.klaviyo.core.Registry
 import com.klaviyo.core.safeApply
 
 /**
- * Initialize in-app forms listeners
- * Once registered, an in-app form may appear within 10s
+ * Load in-app forms data and display a form to the user if applicable based on the forms
+ * configured in your Klaviyo account. Note [Klaviyo.initialize] must be called first
  */
+@UiThread
```
evan-masseau, line 12: FYI - this is just an annotation, doesn't actually enforce anything. But it is a typical pattern to decorate a UI method this way and leave it to the caller to call from the right thread -- but we can revisit if necessary post MVP

README.md
```diff
@@ -465,6 +468,54 @@ extensions such as `import com.klaviyo.pushFcm.KlaviyoRemoteMessage.body` to acc
...
+    .registerForInAppForms()
+
+// ... or any time thereafter
+Klaviyo.registerForInAppForms()
+```
+
+### Behavior
+
+Once `registerForInAppForms()` is called, the SDK will load form data for your account and display no more than one form 
+within 10 seconds, based on form targeting and behavior settings.
```
evan-masseau, line 501: > no more than one form within 10 seconds

Is doing some heavy lifting to indicate we're not doing an ongoing listener without phrasing that in a negative way...

README.md
```diff
@@ -465,6 +468,54 @@ extensions such as `import com.klaviyo.pushFcm.KlaviyoRemoteMessage.body` to acc
...
+advisable so that the user has multiple opportunities to see your messaging if they are browsing your app for a prolonged period.
+However, be advised that this will be shown as soon as the form is ready in the SDK, so you may still need to condition 
+this based on the user's context within your application. Future versions of this product will provide more control in this regard.
+
+| Callback                 | Description                                                                                            |
+|--------------------------|--------------------------------------------------------------------------------------------------------|
+| `Activity.onCreate()`    | If you only want forms to appear over a specific activity, checks every time this activity is created. |
+| `Application.onResume()` | Anytime the app is foregrounded, check for forms and show if available.                                |
+
+**Note:** At this time, when device orientation changes any currently visible form is closed and will not be re-displayed automatically.
```
evan-masseau, line 517: Possibly just remove this as overly detailed? https://github.com/klaviyo/klaviyo-swift-sdk/pull/301#discussion_r1977931403

README.md
```diff
@@ -465,6 +468,54 @@ extensions such as `import com.klaviyo.pushFcm.KlaviyoRemoteMessage.body` to acc
...
+within 10 seconds, based on form targeting and behavior settings.
+
+You can call `registerForInAppForms()` any time after initializing with your company ID to control when and where in 
+your app's UI a form can appear. It is safe to register multiple times per application session. The SDK will internally
+prevent multiple forms appearing at once.
+
+Consider how often you want to register for forms. For example, registering from an activity lifecycle event is 
+advisable so that the user has multiple opportunities to see your messaging if they are browsing your app for a prolonged period.
+However, be advised that this will be shown as soon as the form is ready in the SDK, so you may still need to condition 
+this based on the user's context within your application. Future versions of this product will provide more control in this regard.
```
evan-masseau, line 510: [Related thread on iOS readme](https://github.com/klaviyo/klaviyo-swift-sdk/pull/301/files#r1976843241)

README.md
```diff
@@ -465,6 +468,53 @@ extensions such as `import com.klaviyo.pushFcm.KlaviyoRemoteMessage.body` to acc
...
 
+## In App Forms
+
+[In-app forms](https://klaviyo.slack.com/archives/C064QGU3BCG/p1741024810256069) are messages displayed to mobile app 
+users while they are actively using your app. You can create new in-app forms in a drag-and-drop editor in the 
+Sign-Up Forms tab in Klaviyo.
+
+### Prerequisites
+- Version 3.2.0 or higher
+- Klaviyo `analytics` and `forms` packages
```
dan-peluso, line 479: feels redundant to say forms is a prereq for in app forms. Do we mean they need to import the module?

README.md
dan-peluso: Added my comment there but I do think it's a particularly relevant detail for Android

sdk/forms/src/main/java/com/klaviyo/forms/InAppForms.kt
dan-peluso: I like the decorator

README.md
evan-masseau: It is pretty hand hold y, but I want to keep it because it explicitly calls out analytics as required. The import is covered in the code snippet, unless I misunderstood

evan-masseau: I added the same over-explanation to the push setup section above btw for consistency

README.md
evan-masseau: agree, lets keep

README.md
```diff
@@ -465,6 +399,126 @@ extensions such as `import com.klaviyo.pushFcm.KlaviyoRemoteMessage.body` to acc
 We also provide an `Intent.appendKlaviyoExtras(RemoteMessage)` extension method, which attaches the data to your
 notification intent that the Klaviyo SDK requires in order to track opens when you call `Klaviyo.handlePush(intent)`.
 
+## In App Forms
+
+[In-app forms](https://klaviyo.slack.com/archives/C064QGU3BCG/p1741024810256069) are messages displayed to mobile app
```
evan-masseau, line 404:
```suggestion
[In-app forms](https://help.klaviyo.com/hc/en-us/articles/34567685177883) are messages displayed to mobile app 
```

README.md
```diff
@@ -190,6 +192,8 @@ Klaviyo.createEvent(event)
 - A [Firebase account](https://firebase.google.com/docs/android/setup) for your Android app
 - Familiarity with [Firebase](https://firebase.google.com/docs/cloud-messaging/android/client) documentation
 - Configure [Android push](https://help.klaviyo.com/hc/en-us/articles/14750928993307) in your Klaviyo account settings
+- Klaviyo `analytics` and `push-fcm` packages
```
dan-peluso, line 195: Why does the forms module require the push fcm library?

evan-masseau: it doesn't, this is the push section of the readme

README.md
```diff
@@ -190,6 +192,8 @@ Klaviyo.createEvent(event)
 - A [Firebase account](https://firebase.google.com/docs/android/setup) for your Android app
 - Familiarity with [Firebase](https://firebase.google.com/docs/cloud-messaging/android/client) documentation
```
dan-peluso, line 193: Similarly confused about these as forms prereqs (maybe I'm missing something here?)

README.md
```diff
@@ -465,6 +399,126 @@ extensions such as `import com.klaviyo.pushFcm.KlaviyoRemoteMessage.body` to acc
...
+        <!-- ... -->
+        <application>
+            <!-- ... -->
+            <activity>
+                <!-- ... -->
+                <intent-filter android:label="@string/filter_view_example_gizmos">
+                    <action android:name="android.intent.action.VIEW" />
+                    <category android:name="android.intent.category.DEFAULT" />
+                    <category android:name="android.intent.category.BROWSABLE" />
+                    <!-- Accepts URIs formatted "example://host.com” -->
```
dan-peluso, line 474: This could be extra hand holdy - but is it worth having some images that correspond to the web builder for this schema? Something like 'type in this schema in your push / forms deeplinks' so they know they need the colon and slashes

README.md
```diff
@@ -465,6 +399,126 @@ extensions such as `import com.klaviyo.pushFcm.KlaviyoRemoteMessage.body` to acc
...
+                    <data android:scheme="example" android:host="host.com"/>
+                </intent-filter>
+            </activity>
+        </application>
+    </manifest>
+    ```
+
+2. Read data from incoming intents:
```
dan-peluso, line 482: This made me think - in the forms deeplinks we're pretty explicitly navigating within the same activity using that route. Is this worth mentioning? In our notifications we're attaching it as an intent / click action, but maybe worth differentiating those two experiences

README.md
```diff
@@ -465,6 +399,126 @@ extensions such as `import com.klaviyo.pushFcm.KlaviyoRemoteMessage.body` to acc
...
+
+    ```xml
+    <!-- AndroidManifest.xml -->
+    <manifest>
+        <!-- ... -->
+        <application>
+            <!-- ... -->
+            <activity>
+                <!-- ... -->
+                <intent-filter android:label="@string/filter_view_example_gizmos">
```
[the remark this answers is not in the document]
evan-masseau: oh hahaha its from google's own docs page https://developer.android.com/training/app-links/deep-linking

README.md
evan-masseau: Yeah for that level of detail I often prefer to point folks to the more definitive documentation rather than re-write it myself, so we've got links to Klaviyo's help center docs on deep linking (above) and more to your point, google's docs on deep links below (https://developer.android.com/training/app-links/deep-linking)

README.md
evan-masseau: in fact, the tiniest of nits, but it makes me wanna put something in the `core` package for creating those intents with one extension function or something.

dan-peluso: >  except of course one is from tapping a notification, other is sent from SDK immediately
Yeah I guess this is what I wanted to highlight, that the triggering action comes from the webview and is not being attached to internal code like the notification is. Just trying to add the expectation that whatever you set on the portal will be called immediately when that webview button is clicked, rather than when a notification is clicked into. Could be a little over explainy, but I do think the experiences differ slightly.

dan-peluso (approved):

README.md
evan-masseau: Hm yeah I think i see what you mean, like this language:

> When the app is opened from a deep link, the intent that started the activity contains data for the deep link.

Is specific to a notification because that is actually opening the app, so maybe we should tweak that to be broader.

README.md
```diff
@@ -465,6 +399,126 @@ extensions such as `import com.klaviyo.pushFcm.KlaviyoRemoteMessage.body` to acc
...
+                </intent-filter>
+            </activity>
+        </application>
+    </manifest>
+    ```
+
+2. Read data from incoming intents:
+
+    When the app is opened from a deep link, the intent that started the activity contains data for the deep link.
+    You can parse the URI from the intent's data property and use it to navigate to the appropriate part of your app.
```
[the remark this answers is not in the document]
evan-masseau: @dan-peluso ☝️  ?

README.md
evan-masseau: See below

README.md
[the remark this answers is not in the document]
dan-peluso: LGTM

README.md
```diff
@@ -465,6 +399,126 @@ extensions such as `import com.klaviyo.pushFcm.KlaviyoRemoteMessage.body` to acc
 We also provide an `Intent.appendKlaviyoExtras(RemoteMessage)` extension method, which attaches the data to your
 notification intent that the Klaviyo SDK requires in order to track opens when you call `Klaviyo.handlePush(intent)`.
 
+## In App Forms
+
+[In-app forms](https://help.klaviyo.com/hc/en-us/articles/34567685177883) are messages displayed to mobile app 
+users while they are actively using your app. You can create new in-app forms in a drag-and-drop editor in the 
+Sign-Up Forms tab in Klaviyo. Follow the instructions in this section to integrate forms with your app. The SDK will
+display forms according to targeting and behavior settings and collect delivery and engagement analytics automatically.
```
mark-piana, line 407: Slight tweak, but I feel like this reads a little better - "according to **their** targeting and behavior settings"

evan-masseau: I think iOS has the same verbiage, i can get both patched.

evan-masseau:
```suggestion
display forms according to their targeting and behavior settings and collect delivery and engagement analytics automatically.
```

README.md
evan-masseau:
```suggestion
within 15 seconds, based on form targeting and behavior settings.
```

PR #238, opened 2025-03-05
Patches related to webview.destroy

evan-masseau:
While testing the new abort condition, I noticed that webviews seemed to be accumulating according to chrome device inspector. LeakCanary didn't detect this as a leak for some reason, but it does appear to be accumulating in memory, and simply moving where we call `destroy` from, to be every time rather than only if the view had been attached, fixes the issue in that chrome devices list. 

And while reading up on `webview.destroy` I came across [this method](https://developer.android.com/develop/ui/views/layout/webapps/managing-webview) that is called if the webview's javascript context crashes or gets GC'd. It seems like a very unlikely situation but we may as well handle it gracefully rather than crash the host app.

### DIFF
--- a/sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebViewDelegate.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebViewDelegate.kt
@@ -120,6 +120,16 @@ internal class KlaviyoWebViewDelegate : WebViewClient(), WebViewCompat.WebMessag
         super.onReceivedHttpError(view, request, errorResponse)
     }
 
+    override fun onPageFinished(view: WebView?, url: String?) {
+        super.onPageFinished(view, url)
+        // When an assetSource is specified, log whether we actually got it
+        Registry.config.assetSource?.let { expected ->
+            webView?.evaluateJavascript("window.klaviyoModulesObject?.assetSource") { actual ->
+                Registry.log.info("Actual Asset Source: $actual. Expected $expected")
+            }
+        }
+    }
+
     /**
      * Intercept page navigation and redirect to an external browser application
      */
@@ -229,10 +239,8 @@ internal class KlaviyoWebViewDelegate : WebViewClient(), WebViewCompat.WebMessag
                 Registry.log.verbose("Clear IAF WebView reference")
                 this.webView = null
                 webView.visibility = View.GONE
-                webView.parent?.let {
-                    (it as ViewGroup).removeView(webView)
-                    webView.destroy()
-                }
+                webView.parent?.let { it as ViewGroup }?.removeView(webView)
+                webView.destroy()
             }
         } ?: run {
             Registry.log.warning("Unable to close IAF - null activity reference")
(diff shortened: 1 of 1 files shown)


sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebViewDelegate.kt
```diff
@@ -120,6 +120,16 @@ internal class KlaviyoWebViewDelegate : WebViewClient(), WebViewCompat.WebMessag
         super.onReceivedHttpError(view, request, errorResponse)
     }
 
+    override fun onPageFinished(view: WebView?, url: String?) {
+        super.onPageFinished(view, url)
+        // When an assetSource is specified, log whether we actually got it
+        Registry.config.assetSource?.let { expected ->
```
evan-masseau, line 126: and this is just a handy utility that confirms whether asset source loaded as expected in logcat

ajaysubra (approved):
Neat!

sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebViewDelegate.kt
```diff
@@ -120,6 +121,25 @@ internal class KlaviyoWebViewDelegate : WebViewClient(), WebViewCompat.WebMessag
...
+            webView?.evaluateJavascript("window.klaviyoModulesObject?.assetSource") { actual ->
+                Registry.log.info("Actual Asset Source: $actual. Expected $expected")
```
dan-peluso, line 129: Do we want to go with debug for this? Super nit

evan-masseau: sure

sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebViewDelegate.kt
```diff
@@ -179,11 +199,9 @@ internal class KlaviyoWebViewDelegate : WebViewClient(), WebViewCompat.WebMessag
      * Handle a [BridgeMessage.Show] message by displaying the webview before the form animates in
      */
     private fun show() = webView?.let { webView ->
-        activity?.window?.decorView?.let { decorView ->
-            decorView.post {
-                decorView.findViewById<ViewGroup>(android.R.id.content).addView(webView)
-                webView.visibility = View.VISIBLE
-            }
+        activity?.window?.decorView?.post { decorView ->
+            decorView.findViewById<ViewGroup>(android.R.id.content).addView(webView)
```
dan-peluso, line 203: Is removing the 'decorView.post' intentional here? Noticed we keep it in the close function

evan-masseau: consolidated it with an extension function is all, .post is on the line above now

dan-peluso (approved):
LGTM one small ask could you add the timer cancel on abort to this? I know it's technically a different thing but while you're here

sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebViewDelegate.kt
dan-peluso: Sorry I could have sworn it was gone but I might have been looking at the red, my b

PR #239, opened 2025-03-05
[CHNL-18328] IAF WebView Delegate Unit Testing

dan-peluso:
# Description
Testing everything thing we can without launching UI

|before|after|
|--|--|
|image | image |

# Check List

## Related Issues/Tickets
https://klaviyo.atlassian.net/browse/CHNL-18328

### DIFF
--- a/sdk/fixtures/src/main/java/com/klaviyo/fixtures/BaseTest.kt
+++ b/sdk/fixtures/src/main/java/com/klaviyo/fixtures/BaseTest.kt
@@ -1,9 +1,13 @@
 package com.klaviyo.fixtures
 
+import android.app.Activity
 import android.content.Context
 import android.content.pm.ApplicationInfo
 import android.content.pm.PackageManager
+import android.content.res.AssetManager
 import android.os.Build
+import android.view.View
+import android.view.ViewGroup
 import com.klaviyo.core.Registry
 import com.klaviyo.core.config.Config
 import com.klaviyo.core.lifecycle.LifecycleMonitor
@@ -13,6 +17,7 @@ import io.mockk.mockk
 import io.mockk.mockkObject
 import io.mockk.spyk
 import io.mockk.unmockkObject
+import java.io.ByteArrayInputStream
 import java.lang.reflect.Field
 import java.lang.reflect.Method
 import java.lang.reflect.Modifier
@@ -58,14 +63,38 @@ abstract class BaseTest {
                 }
             }
         }
+
+        val htmlString = """
+            <!DOCTYPE html>
+            <html lang="en">
+            <head data-sdk-name="SDK_NAME"
+                  data-sdk-version="SDK_VERSION"
+                  data-native-bridge-name="BRIDGE_NAME"
+                  data-native-bridge-handshake='BRIDGE_HANDSHAKE'
+            >
+                <meta charset="UTF-8">
+                <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0, viewport-fit=cover"/>
+                <meta name="referrer" content="same-origin" /> <!--  This meta tag protects @imported fonts from being blocked by CORS  -->
+                <title>Klaviyo In-App Form Template</title>
+                <link rel="stylesheet" type="text/css" href="https://static-forms.klaviyo.com/fonts/api/v1/in-app-web-fonts/websafe_fonts.css" crossorigin/>
+                <script type="text/javascript" src="KLAVIYO_JS_URL"></script>
+            </head>
+            <body></body>
+            </html>
+        """.trimIndent()
     }
 
     protected val mockApplicationInfo = mockk<ApplicationInfo>()
     protected val mockPackageManager = mockk<PackageManager>()
+    protected val mockAssets = mockk<AssetManager> {
+        every { open("InAppFormsTemplate.html") } returns
+            ByteArrayInputStream(htmlString.encodeToByteArray())
+    }
 
     protected val mockContext = mockk<Context>().apply {
         every { applicationInfo } returns mockApplicationInfo
         every { packageManager } returns mockPackageManager
+        every { assets } returns mockAssets
     }
 
     protected val debounceTime = 5
@@ -80,10 +109,22 @@ abstract class BaseTest {
         every { networkJitterRange } returns 0..0
         every { baseUrl } returns "https://test.fake-klaviyo.com"
         every { apiRevision } returns "1234-56-78"
+        every { baseCdnUrl } returns "https://decent.cdn.url.com"
+        every { assetSource } returns null
+        every { sdkName } returns "klaviyo-android-sdk"
+        every { sdkVersion } returns "4.20.69"
+    }
+    protected val mockContentView: ViewGroup = mockk(relaxed = true)
+    protected val mockDecorView: View = mockk(relaxed = true) {
+        every { findViewById<ViewGroup>(any()) } returns mockContentView
+    }
+    protected val mockActivity: Activity = mockk(relaxed = true) {
+        every { window.decorView } returns mockDecorView
     }
     protected val mockLifecycleMonitor = mockk<LifecycleMonitor>().apply {
         every { onActivityEvent(any()) } returns Unit
         every { offActivityEvent(any()) } returns Unit
+        every { currentActivity } returns mockActivity
     }
     protected val mockNetworkMonitor = mockk<NetworkMonitor>()
     protected val spyDataStore = spyk(InMemoryDataStore())
--- a/sdk/fixtures/src/main/java/com/klaviyo/fixtures/BaseTest.kt
+++ b/sdk/fixtures/src/main/java/com/klaviyo/fixtures/BaseTest.kt
@@ -1,13 +1,9 @@
 package com.klaviyo.fixtures
 
-import android.app.Activity
 import android.content.Context
 import android.content.pm.ApplicationInfo
 import android.content.pm.PackageManager
-import android.content.res.AssetManager
 import android.os.Build
-import android.view.View
-import android.view.ViewGroup
 import com.klaviyo.core.Registry
 import com.klaviyo.core.config.Config
 import com.klaviyo.core.lifecycle.LifecycleMonitor
@@ -17,7 +13,6 @@ import io.mockk.mockk
 import io.mockk.mockkObject
 import io.mockk.spyk
 import io.mockk.unmockkObject
-import java.io.ByteArrayInputStream
 import java.lang.reflect.Field
 import java.lang.reflect.Method
 import java.lang.reflect.Modifier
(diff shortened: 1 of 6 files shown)


sdk/fixtures/src/main/java/com/klaviyo/fixtures/BaseTest.kt
```diff
@@ -58,14 +63,38 @@ abstract class BaseTest {
                 }
             }
         }
+
+        val htmlString = """
```
evan-masseau, line 67: Prefer to move this and its setup stuff to the one file that is using it. `BaseTest` is already on its way to getting a bit bloated, I prefer to only put properties and setup stuff in here if its used by multiple test files.

sdk/fixtures/src/main/java/com/klaviyo/fixtures/BaseTest.kt
```diff
@@ -80,10 +109,22 @@ abstract class BaseTest {
...
+        every { assetSource } returns null
+        every { sdkName } returns "klaviyo-android-sdk"
+        every { sdkVersion } returns "4.20.69"
+    }
+    protected val mockContentView: ViewGroup = mockk(relaxed = true)
+    protected val mockDecorView: View = mockk(relaxed = true) {
+        every { findViewById<ViewGroup>(any()) } returns mockContentView
+    }
+    protected val mockActivity: Activity = mockk(relaxed = true) {
+        every { window.decorView } returns mockDecorView
```
evan-masseau, line 122: Same with these props, since they're only used in one test file, move them to that file plz

sdk/forms/src/test/java/com/klaviyo/forms/KlaviyoWebViewDelegateTest.kt
```diff
@@ -0,0 +1,380 @@
...
+import org.junit.Before
+import org.junit.Test
+
+internal class KlaviyoWebViewDelegateTest : BaseTest() {
+
+    private companion object {
+        private val slotOnActivityEvent = slot<ActivityObserver>()
+    }
+
+    private val mockApiClient: ApiClient = mockk(relaxed = true)
```
evan-masseau, line 44: I have a concern that relaxing every mock runs a risk of papering over an issue or making it easier to miss out on test cases. TBH I haven't reckoned with this much so I'm not sure how strongly I hold this feeling, might need to do a little reading before I form an opinion. Your thoughts?

sdk/forms/src/test/java/com/klaviyo/forms/InAppFormsTest.kt
```diff
@@ -0,0 +1,74 @@
...
+
+        verify { delegate.initializeWebView() }
+    }
+
+    @Test
+    fun `registers a delegate if we don't have one`() {
+        assert(!Registry.isRegistered<KlaviyoWebViewDelegate>())
+        Klaviyo.registerForInAppForms()
+        assert(Registry.isRegistered<KlaviyoWebViewDelegate>())
+    }
```
evan-masseau, line 73: Lets add a test that throwing an exception from within `registerForInAppForms` does not cause a crash. e.g. what it looks like if `MissingConfig` is thrown because they haven't initialized yet.

sdk/forms/src/test/java/com/klaviyo/forms/KlaviyoWebViewDelegateTest.kt
```diff
@@ -0,0 +1,380 @@
...
+        delegate.postMessage("""{"type":"formWillAppear"}""")
+
+        verifyShow()
+    }
+
+    @Test
+    fun `malformed message throws an error`() {
+        val delegate = KlaviyoWebViewDelegate()
+        delegate.initializeWebView()
+        delegate.postMessage("""sawr a warewolf with a chinese menu inhis hands""")
```
evan-masseau, line 144: yes.

sdk/forms/src/test/java/com/klaviyo/forms/KlaviyoWebViewDelegateTest.kt
```diff
@@ -0,0 +1,380 @@
...
+    @Test
+    fun `settings are properly set`() {
+        every {
+            anyConstructed<KlaviyoWebView>()
+                .settings
+        } returns mockSettings
+        val delegate = KlaviyoWebViewDelegate()
+        delegate.initializeWebView()
+
+        verify { mockSettings.javaScriptEnabled = true }
```
evan-masseau, line 219: Oh this is nice, does using the relaxed mock mean that a property actually gets set and saved in a map or something? I feel like setting this property in a strict mock would have thrown unless you explicitly mocked a setter?

sdk/forms/src/test/java/com/klaviyo/forms/KlaviyoWebViewDelegateTest.kt
```diff
@@ -0,0 +1,380 @@
...
+
+        val delegate = KlaviyoWebViewDelegate()
+        delegate.initializeWebView()
+        delegate.postMessage("""{"type":"formDisappeared"}""")
+        verify { spyLog.warning("Unable to close IAF - null activity reference") }
+        verifyClose(true)
+    }
+
+    @Test
+    fun `postMessage handles AggregateEventTracked`() {
```
evan-masseau, line 323: nit: some of these might be a bit duplicative of tests in `BridgeMessageTest`

evan-masseau: Meh i guess that doesn't really matter though. Except maybe nice for DRYness to pull some shared strings out into a fixtures file.

evan-masseau:
Some great tests in here, left some comments. I see no issue merging this prior to release, I also don't think its blocking if you need time tomorrow to respond to the feedback. Merged or not, just knowing that these tests pass is another release confidence booster!

sdk/fixtures/src/main/java/com/klaviyo/fixtures/BaseTest.kt
```diff
@@ -80,10 +109,22 @@ abstract class BaseTest {
         every { networkJitterRange } returns 0..0
         every { baseUrl } returns "https://test.fake-klaviyo.com"
         every { apiRevision } returns "1234-56-78"
+        every { baseCdnUrl } returns "https://decent.cdn.url.com"
```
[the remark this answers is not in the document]
dan-peluso: I'll update those while I'm here

sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebView.kt
```diff
@@ -43,6 +33,18 @@ internal class KlaviyoWebView : WebView {
             null
         )
 
+    private fun configure() = apply {
+        settings.userAgentString = DeviceProperties.userAgent
+        settings.javaScriptEnabled = true
+        settings.domStorageEnabled = true
+        setBackgroundColor(Color.TRANSPARENT)
+        if (Registry.config.isDebugBuild) {
+            setWebContentsDebuggingEnabled(BuildConfig.DEBUG)
```
evan-masseau, line 42:
```suggestion
            setWebContentsDebuggingEnabled(true)
```

evan-masseau: or `Registry.config.isDebugBuild`

sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebView.kt
```diff
@@ -43,6 +33,18 @@ internal class KlaviyoWebView : WebView {
...
+            // Disable webview resources cache when debugging:
+            // settings.cacheMode = WebSettings.LOAD_NO_CACHE
+        }
```
evan-masseau, line 45:
```suggestion
        setWebContentsDebuggingEnabled(Registry.config.isDebugBuild)
        // Disable webview resources cache when debugging:
        // settings.cacheMode = WebSettings.LOAD_NO_CACHE
```

sdk/forms/src/test/java/com/klaviyo/forms/KlaviyoWebViewDelegateTest.kt
```diff
@@ -0,0 +1,380 @@
...
+        every {
+            anyConstructed<KlaviyoWebView>()
+                .settings
+        } returns mockSettings
+        val delegate = KlaviyoWebViewDelegate()
+        delegate.initializeWebView()
+
+        verify { mockSettings.javaScriptEnabled = true }
+        verify { mockSettings.userAgentString = "Mock User Agent" }
+        verify { mockSettings.domStorageEnabled = true }
```
evan-masseau, line 221:
```suggestion
        verify { mockSettings.domStorageEnabled = true }
        verify { mockSettings.cacheMode = null }
```
This would enforce we never mistakenly commit `cacheMode`  bc a test will fail.

sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebView.kt
evan-masseau: or maybe
`settings.cacheMode = if(Registry.config.isDebugBuild) WebSettings.LOAD_NO_CACHE else LOAD_DEFAULT`

evan-masseau: oh wait no don't do that.

evan-masseau: just thought of this bc of that thread about not referencing `BuildConfig.DEBUG` directly

dan-peluso: Trying to add this now, but ngl this is trickier to test. Probably not a good enough reason to avoid this since I agree the code is cleaner, but mockk doesn't like to interact with Android UI function calls. In this case, this is a UI call and setting the test to be _not a debug build_ was able to avoid this call entirely. Would you be okay with me keeping this as is?

evan-masseau: ooh ok yeah thats fine too, maybe just `setWebContentsDebuggingEnabled(true)` then since its already in the condition!

evan-masseau:
Weird, I rebase merged onto master, not sure why that now conflicts.

sdk/forms/src/test/java/com/klaviyo/forms/KlaviyoWebViewDelegateTest.kt
dan-peluso: image
rip

sdk/forms/src/test/java/com/klaviyo/forms/KlaviyoWebViewDelegateTest.kt
dan-peluso: Yeah I think it's reasonable to be cautious, but relaxing mocks of things like API client which is sort of a down stream from the behavior we're checking here I think is fair game. Mostly this test is just a way for us to check the integration of how messages are invoking dependencies, but for something like an API client we don't really care _how_ the result is getting processed and more that it's called at all.

dan-peluso:
> Weird, I rebase merged onto master, not sure why that now conflicts.

I probably need to rebase locally and force push to clean up some of those commits

sdk/forms/src/test/java/com/klaviyo/forms/KlaviyoWebViewDelegateTest.kt
dan-peluso: going to add something like this, which basically ensures we never access that property
```suggestion
        verify { mockSettings.domStorageEnabled = true }
        verify(exactly = 0) { mockSettings.cacheMode }
```

evan-masseau: yaaa that works

sdk/forms/src/test/java/com/klaviyo/forms/InAppFormsTest.kt
dan-peluso: Finally got one working, was trying to get the wrong thing to throw that exception for faaaar too long

sdk/forms/src/test/java/com/klaviyo/forms/KlaviyoWebViewDelegateTest.kt
evan-masseau: oh jeez i was mis reading this as `mockSettings.javaScriptEnabled == true`

evan-masseau: ok cool thx

evan-masseau (approved):

PR #240, opened 2025-03-06
Add webview compatibility notice

evan-masseau:
# Description

# Check List

### DIFF
--- a/README.md
+++ b/README.md
@@ -544,6 +544,16 @@ the following metadata tag to your manifest file.
 </manifest>
 ```
 
+#### WebViews Compatibility
+Klaviyo's in-app forms are powered by [WebViews](https://developer.android.com/reference/android/webkit/WebView).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -544,6 +544,16 @@ the following metadata tag to your manifest file.
 </manifest>
 ```
 
+#### WebViews Compatibility
+Klaviyo's in-app forms are powered by [WebViews](https://developer.android.com/reference/android/webkit/WebView).
+At this time, we require a version of WebView compatible with JavaScript standard ES2015. Older versions will fail
+gracefully without displaying a form to the user.
+
+WebView is a system app that updates independently of the OS (as of API level 21). Therefore, understanding backwards
```
evan-masseau, line 552: Its in the troubleshooting section, so some technical detail seems OK, but IDK if this is _too much_ detail.

dan-peluso: I think this is a good amount of detail

dan-peluso (approved):

PR #241, opened 2025-03-12
[CHNL-12435] Adding new lifecycle listener function

dan-peluso:
# Description
- while porting forms to react native, we found that if the Klaviyo SDK is initialized after the first activity is started, our internal lifecycle listener has no way to access the current activity. This means that we cannot show the in app web form. To protect against this, I'm adding back the lifecycle listener function. In the case of react native, this means that we need to update that documentation to tell users 'hey add this to your Android built application onCreate'. 
- 
# Check List

## Test Plan
Tested on an Android device with the test app, on the react native example app, and the react native test app.

## Related Issues/Tickets
https://klaviyo.atlassian.net/browse/CHNL-18547

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,10 @@ persist data. Upon initialize, the SDK registers listeners for your application'
 to gracefully manage background processes.
 
 `Klaviyo.initialize()` **must** be called before any other SDK methods can be invoked. We recommend initializing from 
-the earliest point in your application code, such as the `Application.onCreate()` method.
+the earliest point in your application code, such as the `Application.onCreate()` method If you are
+unable to provide your Klaviyo API Key in the `Application.onCreate()` method, please call
+`Klaviyo.registerForLifecycleCallbacks()` in the `onCreate()`. You will still need to provide the API
+key eventually for full functionality.
 
 ```kotlin
 // Application subclass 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -84,7 +84,10 @@ persist data. Upon initialize, the SDK registers listeners for your application'
 to gracefully manage background processes.
 
 `Klaviyo.initialize()` **must** be called before any other SDK methods can be invoked. We recommend initializing from 
-the earliest point in your application code, such as the `Application.onCreate()` method.
+the earliest point in your application code, such as the `Application.onCreate()` method If you are
```
evan-masseau, line 87: Wondering if we should make this callout less prominent, since it will hopefully be quite an edge case. Maybe as a note below

evan-masseau: Couldn't write this in a comment suggestion, put it in PR instead :
https://github.com/klaviyo/klaviyo-android-sdk/pull/242

README.md
```diff
@@ -99,6 +102,11 @@ class YourApplication : Application() {
         
         // Initialize is required before invoking any other Klaviyo SDK functionality 
         Klaviyo.initialize("KLAVIYO_PUBLIC_API_KEY", applicationContext)
+        
+        // If you don't have the API key at Application creation, call this before
+        Klaviyo.registerForLifecycleCallbacks(applicationContext)
+        // and this after
+        Klaviyo.initialize("KLAVIYO_PUBLIC_API_KEY", applicationContext)
```
evan-masseau, line 109: nit: felt superfluous to give them this syntax again

evan-masseau (approved):

PR #253, opened 2025-04-25
[CHNL-20017] Web Forms Environment Switching Support

dan-peluso:
# Description
- new core config for setting up web forms environment
- new enum! to support this (in case we ever do add more forms environments) (also bc booleans can be tricky to interpret sometimes)

Test App PR - https://github.com/klaviyo/klaviyo-android-test-app/pull/89
# Check List

[form-env-switching.webm](https://github.com/user-attachments/assets/561c8933-fa42-4a4b-b13b-768cf807a88c)

## Related Issues/Tickets
https://github.com/klaviyo/klaviyo-android-test-app/pull/89

### DIFF
--- a/sdk/core/src/main/java/com/klaviyo/core/config/FormEnvironment.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/config/FormEnvironment.kt
@@ -1,0 +1,14 @@
+package com.klaviyo.core.config
+
+enum class FormEnvironment(val templateName: String) {
+    IN_APP("in-app"),
+    WEB("web");
+
+    companion object {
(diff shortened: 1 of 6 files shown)


sdk/core/src/main/java/com/klaviyo/core/config/FormEnvironment.kt
```diff
@@ -0,0 +1,14 @@
+package com.klaviyo.core.config
+
+enum class FormEnvironment(val templateName: String) {
```
evan-masseau, line 3: 💯

evan-masseau (approved):

PR #256, opened 2025-05-05
[CHNL-20122] SDK Plugin User-Agent Field update

dan-peluso:
# Description

- we want a way to discern if a request is being sent with an app that was built using the expo plugin
- to avoid a backend change, we'll add it to the user agent field
- EXPO apps will add a string resource with the new id, and if the SDK detects those have been set it will add them to the user agent

# Check List

## Test Plan
Need to release this on both android and react native before we can test with expo, but the same resource overriding logic is what we use for the SDK name and version

## Related Issues/Tickets
https://klaviyo.atlassian.net/browse/CHNL-20122

### DIFF
--- a/sdk/core/src/main/java/com/klaviyo/core/DeviceProperties.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/DeviceProperties.kt
@@ -57,7 +57,8 @@ object DeviceProperties {
     }
 
     val notificationPermissionGranted: Boolean
-        get() = NotificationManagerCompat.from(Registry.config.applicationContext).areNotificationsEnabled()
+        get() = NotificationManagerCompat.from(Registry.config.applicationContext)
+            .areNotificationsEnabled()
 
     val applicationId: String by lazy {
         Registry.config.applicationContext.packageName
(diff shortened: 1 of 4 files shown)


sdk/core/src/main/java/com/klaviyo/core/DeviceProperties.kt
```diff
@@ -69,11 +70,26 @@ object DeviceProperties {
...
 
+    val pluginSdkVersion: String? by lazy {
+        Registry.config.applicationContext.getString(R.string.klaviyo_sdk_plugin_version_override)
+            .ifBlank { null }
+    }
+
+    val userAgent: String
+        get() {
+            val pluginString = pluginSdk?.let {
+                "($it/$pluginSdkVersion)"
```
evan-masseau, line 86: extremely nitpicky, but can we guard against this being null/empty better? even if you just put "unknown" or something, I think i'll work better for our regex/detection if we have `klaviyo-expo/SOMETHING` and never just `plugin/`

sdk/fixtures/src/main/java/com/klaviyo/fixtures/DevicePropertiesFixture.kt
```diff
@@ -21,6 +21,8 @@ fun mockDeviceProperties() {
     every { DeviceProperties.deviceId } returns "Mock Device ID"
     every { DeviceProperties.manufacturer } returns "Mock Manufacturer"
     every { DeviceProperties.osVersion } returns "Mock OS Version"
+    every { DeviceProperties.pluginSdk } returns "klaviyo-expo"
```
evan-masseau, line 24: another extreeeme nit, but lets use a mock value like `klaviyo-mock-plugin` or w/e

evan-masseau (approved):
some of the most trivial feedback i've ever left, but LGTM

PR #257, opened 2025-05-06
Forms | Refactor KlaviyoWebViewDelegate into two classes

evan-masseau:
# Description
Splits `KlaviyoWebViewDelegate` up by the two interfaces/superclasses that it implemented to simplify the existing code and set us up for adding more complexity for upcoming features. In updating the tests it already feels like it simplifies some of the test writing. 

# Check List

- [X] Are your changes backwards compatible with previous SDK Versions?

## Changelog / Code Overview

- Split `KlaviyoWebViewDelegate` into `KlaviyoWebViewClient` and `BridgeMessageHandler`
- Split up the existing test coverage, updated mocks to simplify a number of test cases
- Added a couple new tests of previously uncovered methods. 

## Test Plan

- Unit test updates
- IAF testing spreadsheet [Test Sheet](https://docs.google.com/spreadsheets/d/1kgQA13vJzTXReKLz9UihEWkQaz8hGcnezZp6E9J_VoQ/edit?gid=1099883529#gid=1099883529)
- Coverage before and after:

| Before | After |
|--------|--------|
|  old |  new  |
 

## Related Issues/Tickets
Precursor work for 
https://klaviyo.atlassian.net/browse/CHNL-19754
https://klaviyo.atlassian.net/browse/CHNL-19959


jordan-griffin (approved):
I cant see any logical drift, looks like a solid refactor. I like it!

sdk/forms/src/main/java/com/klaviyo/forms/BridgeMessageHandler.kt
```diff
@@ -0,0 +1,115 @@
...
+ * Manages [KlaviyoWebView] to power in-app forms
+ */
+internal class BridgeMessageHandler(
+    private val client: KlaviyoWebViewClient = Registry.get()
+) : WebViewCompat.WebMessageListener {
+    /**
+     * This bridge object is injected into a [KlaviyoWebView] as a global property on the window.
+     * This is the name that will be used to access the bridge from JS, i.e. window.KlaviyoNativeBridge
+     */
+    val name = "KlaviyoNativeBridge"
```
dan-peluso, line 26: Should this be in a private companion object? Just had a scary thought that R8 could burn us here if we're using a constant that could be obfuscated and then aren't able to communicate with the JS bridge. Could also make a new file with these and @keep similar to how we've done for some other stuff

sdk/forms/src/main/java/com/klaviyo/forms/BridgeMessageHandler.kt
```diff
@@ -0,0 +1,115 @@
...
+            is BridgeMessage.Show -> show()
+            is BridgeMessage.AggregateEventTracked -> createAggregateEvent(bridgeMessage)
+            is BridgeMessage.ProfileEvent -> createProfileEvent(bridgeMessage)
+            is BridgeMessage.DeepLink -> deepLink(bridgeMessage)
+            is BridgeMessage.Close -> close()
+            is BridgeMessage.Abort -> abort(bridgeMessage.reason)
+        }
+    } catch (e: Exception) {
+        Registry.log.warning("Failed to relay webview message: $message", e)
+    }.let { }
```
dan-peluso, line 61: Is this because we're putting the`= try {` instead of a regular function here? Might be able to avoid the unnecessary .let at the bottom if we call it without the equals and nest the try catch

sdk/forms/src/main/java/com/klaviyo/forms/BridgeMessageHandler.kt
evan-masseau: Shrinker issues i've encountered in the past were typically when referencing something dynamically or through reflection in such a way that the shrinker can't detect its used, so it strips it out. The most recent example I remember was the event/profile key words that we used in RN but only via reflection, so we had to add those `keep` rules

All that said, i was waffling on putting this in a companion object, so I think i'll do that.

sdk/forms/src/main/java/com/klaviyo/forms/BridgeMessageHandler.kt
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * Handle a [BridgeMessage.DeepLink] message by broadcasting an intent to the host app
+     * similar to how we handle deep links from a notification
+     */
+    private fun deepLink(messageType: BridgeMessage.DeepLink) {
+        Registry.lifecycleMonitor.currentActivity?.startActivity(
+            Intent().apply {
+                data = messageType.route.toUri()
+                action = Intent.ACTION_VIEW
+                `package` = Registry.config.applicationContext.packageName
```
dan-peluso, line 95: How is this different than the 'setPackage' call we had before?

sdk/forms/src/main/java/com/klaviyo/forms/InAppForms.kt
```diff
@@ -11,10 +11,10 @@ import com.klaviyo.core.safeApply
  */
 @UiThread
 fun Klaviyo.registerForInAppForms(): Klaviyo = safeApply {
-    // Ensure we only ever register one delegate instance
-    if (!Registry.isRegistered<KlaviyoWebViewDelegate>()) {
-        Registry.register<KlaviyoWebViewDelegate>(KlaviyoWebViewDelegate())
+    // Ensure we only ever register one KlaviyoWebViewClient instance
+    if (!Registry.isRegistered<KlaviyoWebViewClient>()) {
+        Registry.register<KlaviyoWebViewClient>(KlaviyoWebViewClient())
```
dan-peluso, line 16: Incredibly unrelated but I wonder if we can do some generics hacking to get us to not have to manually add the type here if we already pass a strong type in as the register argument

sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebView.kt
```diff
@@ -22,10 +22,10 @@ internal class KlaviyoWebView : WebView {
 
     constructor(context: Context, attrs: AttributeSet?) : super(context, attrs, 0)
 
-    fun loadTemplate(html: String, delegate: KlaviyoWebViewDelegate) = configure()
-        .setDelegate(delegate)
+    fun loadTemplate(html: String, client: KlaviyoWebViewClient) = configure()
+        .setClient(client)
         .loadDataWithBaseURL(
-            delegate.allowedOrigin.first(),
+            client.nativeBridge.allowedOrigin.first(),
```
dan-peluso, line 28: Do we want this field to be exposed? I wonder if it could make some refactoring tricky down the line if these two are so tightly coupled. Maybe the native bridge deserves its own class as well?

sdk/forms/src/main/java/com/klaviyo/forms/BridgeMessageHandler.kt
evan-masseau: even dumber, it was so that above, `message.data?.let { postMessage(it) } ?: Unit` this could fit on one line

sdk/forms/src/main/java/com/klaviyo/forms/BridgeMessageHandler.kt
evan-masseau: Its the same, its just kotlin magic converting any Java `setXYZ` methods to setter properties

evan-masseau: this one just looks funny bc package is a reserved word so it has to be escaped

sdk/forms/src/main/java/com/klaviyo/forms/InAppForms.kt
evan-masseau: Yeah TBH I'm not sure why it doesn't let me do that. But typically we register an instance under its interface type, not its class type, so I wasn't too worried about "fixing" it.

sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebView.kt
evan-masseau: Not sure I follow this comment. Do you mean pass it in as a separate argument?

dan-peluso: I guess I mean it feels like a bit of a code smell to have functionality in the webview need to access the native bridge field of the client class. And I know unit testing this stuff is difficult anyways given the android sdk dependencies, but my concern is we're coupling the native bridge fields we need in the webview with the webview client. Just wondering if we could decouple that bridge as well.

sdk/forms/src/main/java/com/klaviyo/forms/BridgeMessageHandler.kt
```diff
@@ -0,0 +1,115 @@
...
+                flags = Intent.FLAG_ACTIVITY_SINGLE_TOP
+            }
+        ) ?: run {
+            Registry.log.warning("Unable to open deep link - null activity reference")
```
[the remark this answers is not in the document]
evan-masseau: sure

sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoWebView.kt
evan-masseau: OK, yeah so seems like splitting it into two arguments, instead of referencing the bridge as a property of client addresses your concern.

dan-peluso (approved):

evan-masseau:
Completed regression test cycle finally, good to go

PR #258, opened 2025-05-07
[CHNL-19776] In App Forms Config

dan-peluso:
# Description
- adding a new data class config for in app forms registry
- only a single field atm, session timeout

# Check List

## Related Issues/Tickets
https://klaviyo.atlassian.net/browse/CHNL-19776

### DIFF
--- a/sdk/forms/src/main/java/com/klaviyo/forms/InAppFormsConfig.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/InAppFormsConfig.kt
@@ -3,8 +3,15 @@ package com.klaviyo.forms
 /**
  * Configuration for in app forms
  *
- * @param sessionTimeoutDuration (seconds) timeout for listening to new in app forms to display
+ * @param sessionTimeoutDuration Duration (in seconds) of the period of user inactivity after which the user's app session is terminated. Defaults to 1 Hour.
  */
 data class InAppFormsConfig(
-    val sessionTimeoutDuration: Long = 3600
-)
+    val sessionTimeoutDuration: Long = DEFAULT_SESSION_TIMEOUT
+) {
+    companion object {
+        /***
+         * One hour - meaning we will destroy the forms connection after an hour of inactivity
+         */
+        const val DEFAULT_SESSION_TIMEOUT = 3600L
+    }
+}
--- a/sdk/forms/src/main/java/com/klaviyo/forms/InAppFormsConfig.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/InAppFormsConfig.kt
@@ -6,5 +6,5 @@ package com.klaviyo.forms
  * @param sessionTimeoutDuration (seconds) timeout for listening to new in app forms to display
  */
 data class InAppFormsConfig(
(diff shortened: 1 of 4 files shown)


sdk/forms/src/main/java/com/klaviyo/forms/InAppFormsConfig.kt
```diff
@@ -0,0 +1,10 @@
+package com.klaviyo.forms
+
+/**
+ * Configuration for in app forms
+ *
+ * @param sessionTimeoutDuration (seconds) timeout for listening to new in app forms to display
+ */
+data class InAppFormsConfig(
+    val sessionTimeoutDuration: Long = 3600
```
belleklaviyo, line 9: similar comment to https://github.com/klaviyo/klaviyo-swift-sdk/pull/338#discussion_r2078253226 -- could we define the default value not in the config itself but where its used higher up to keep this more generic

dan-peluso: done

belleklaviyo (approved):

sdk/forms/src/main/java/com/klaviyo/forms/InAppFormsConfig.kt
```diff
@@ -0,0 +1,10 @@
+package com.klaviyo.forms
+
+/**
+ * Configuration for in app forms
+ *
+ * @param sessionTimeoutDuration (seconds) timeout for listening to new in app forms to display
+ */
+data class InAppFormsConfig(
+    val sessionTimeoutDuration: Long
```
evan-masseau, line 9: Lets put a default value on here?

sdk/forms/src/main/java/com/klaviyo/forms/InAppFormsConfig.kt
```diff
@@ -0,0 +1,10 @@
+package com.klaviyo.forms
+
+/**
+ * Configuration for in app forms
+ *
+ * @param sessionTimeoutDuration (seconds) timeout for listening to new in app forms to display
```
evan-masseau, line 6: We gotta come up with a better definition for this. Belle used

> The duration in seconds for which a form session should be extended.

But I don't think that's super intuitive either. How about

> Duration (in seconds) of the period of user inactivity after which an app session is terminated.

sdk/forms/src/main/java/com/klaviyo/forms/InAppFormsConfig.kt
evan-masseau: My thinking is that (when this class supports more things) the developer should be able to flexibly specify  only the config values they care to override without having to set ALL the things.

sdk/forms/src/main/java/com/klaviyo/forms/InAppFormsConfig.kt
evan-masseau: cc @belleklaviyo 
lets workshop this comment and then use consistent language on both platforms

evan-masseau (approved):

PR #263, opened 2025-05-13
IAF | Display in an activity

evan-masseau:
# Description
Move to displaying the form overlay as an activity instead of inserting into host app's view hierarchy.

Happy to start fielding feedback on this one, I've done local testing but haven't documented it all yet. I can re-ping for final review when I've completed the PR writeup.

## Due Diligence


sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoFormsOverlayActivity.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        Registry.get<WebViewClient>().attachWebView(this)
+    }
+
+    override fun finish() {
+        Registry.get<WebViewClient>().detachWebView(this)
+        super.finish()
+    }
```
evan-masseau, line 24: This activity will be a little hole in our test coverage, since unit testing these is nigh impossible. So keeping them skinny and testing the services it relies on instead will at least improve confidence.

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoFormsOverlayActivity.kt
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Presented over the host application to display a Klaviyo form when triggered.
+ *
+ * TODO On back button, close the form within the webview first (for the css animation)
+ */
+internal class KlaviyoFormsOverlayActivity : AppCompatActivity() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
```
dan-peluso, line 17: Truly getting into the weeds on this but I wonder if we even need this super call? Just trying to find ways to make this as lightweight as possible

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoPresentationManager.kt
```diff
@@ -35,23 +34,21 @@ internal class KlaviyoPresentationManager() : PresentationManager {
...
-            Registry.get<WebViewClient>().attachWebView(
-                it
-            )
-        } ?: run {
-            Registry.log.warning("Unable to show IAF - null activity reference")
-        }
+    /**
+     * Launch the overlay activity
+     */
+    override fun present() = if (Registry.lifecycleMonitor.currentActivity !is KlaviyoFormsOverlayActivity) {
```
dan-peluso, line 40: Do we need to do any sort of checking if this activity exists in the stack? We probably only ever want one of these at once, just thinking about if some other service is launching things or this activity is getting abandoned

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoFormsOverlayActivity.kt
```diff
@@ -0,0 +1,36 @@
...
+        Registry.get<WebViewClient>().attachWebView(this)
+    }
+
+    override fun finish() {
```
dan-peluso, line 21: Do we care about detaching in any other state? Like destroy/pause/stop. I know calling finish eventually leads to these, but what about if someone closes the app or it crashes? I'm not sure this is guaranteed to be called

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoFormsOverlayActivity.kt
```diff
@@ -0,0 +1,36 @@
...
+        Registry.get<WebViewClient>().detachWebView(this)
```
dan-peluso, line 22: Also realizing adding this could be a good enough reason to get some lightweight espresso testing in, event just to validate lifecycle behavior and certain registry calls. Definitely not asking for this to be done in this PR, but want to bring it up again.

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
```diff
@@ -99,37 +99,29 @@ internal class KlaviyoWebViewClient(
...
 
     /**
-     * Detach the webview from the activity
+     * Detach the webview from the overlay activity, keeping it in memory
      */
     override fun detachWebView(activity: Activity) = apply {
         webView?.let { webView ->
-            handshakeTimer?.cancel()
-            activity.window?.decorView?.post {
+            activity.runOnUiThread {
```
dan-peluso, line 118: Was removing the handshake timer call intentional here?

sdk/forms/src/test/java/com/klaviyo/forms/webview/KlaviyoWebViewClientTest.kt
```diff
@@ -28,7 +28,7 @@ import org.junit.Assert.assertEquals
 import org.junit.Before
 import org.junit.Test
 
-internal class KlaviyoWebViewClientTest : BaseTest() {
+class KlaviyoWebViewClientTest : BaseTest() {
```
dan-peluso, line 31: I'm kinda curious why we had all these internal in the first place - does having this all be public mess with full SDK unit testing flows?

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoFormsOverlayActivity.kt
evan-masseau: Yeah, checked, and there's enough logic in the super class that we need to call them

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoFormsOverlayActivity.kt
evan-masseau: good to know, will add that to the testing plans

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoFormsOverlayActivity.kt
evan-masseau: Hey google, remind me to google espresso testing

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
evan-masseau: Ah no, I think that was carry over from another experiment, thanks!

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoPresentationManager.kt
evan-masseau: Hm good question, probably good to have more safeguards than just if it is the current activity

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
evan-masseau: oh wait yes-ish. meant to move it to destroy though.

sdk/forms/src/test/java/com/klaviyo/forms/webview/KlaviyoWebViewClientTest.kt
evan-masseau: Doesn't seem to break anything to be internal, but since it is in a different source set it doesn't really matter either way. From git history looks like i just mistakenly set it to internal in a prior refactor. All our other test files are public so I'll just do this for consistency

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoFormsOverlayActivity.kt
evan-masseau: Refactored to put all the presentation logic and attach/detach concerns into the presentation manager.

evan-masseau: And I think I'll be addressing more lifecycle stuff for the app session work coming shortly.

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoPresentationManager.kt
evan-masseau: Adjusted the logic here a bit. Now I have a weak reference to our activity, and tracking presentation state in memory. So that should be more robust than just checking what the current activity is.

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoFormsOverlayActivity.kt
evan-masseau: Made this about as skinny as possible by moving more logic to lifecycle listeners in the presentation manager instead!

sdk/core/src/test/java/com/klaviyo/core/utils/TakeIfTest.kt
```diff
@@ -0,0 +1,35 @@
...
+    fun `takeIfNot runs on non-matching type`() {
+        var itRan = false
+        aThing.takeIfNot<Int>()?.let { itRan = true }
+        assert(itRan) { "takeIfNot should run on non-matching type" }
+    }
+
+    @Test
+    fun `takeIfNot does not run on matching type`() {
+        var itRan = false
+        aThing.takeIfNot<String>()?.let { itRan = true }
```
dan-peluso, line 32: You don't see uniquely named variables like this anymore, good stuff

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoFormsOverlayActivity.kt
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Presented over the host application to display a Klaviyo form when triggered.
+ */
+internal class KlaviyoFormsOverlayActivity : AppCompatActivity() {
+
+    /**
+     * TODO On back button, close the form within the webview first (for the css animation), requires JS injection
+     */
+    override fun onBackPressed() = Registry.get<PresentationManager>().dismiss()
```
dan-peluso, line 15: Does this still kill the activity? Wondering if we need the super call here or if it's covered

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoPresentationManager.kt
```diff
@@ -24,34 +31,45 @@ internal class KlaviyoPresentationManager() : PresentationManager {
...
-            Registry.get<WebViewClient>().detachWebView(
-                it
-            )
-        } ?: run {
-            Registry.log.warning("Unable to dismiss IAF - null activity reference")
-        }
+    /**
+     * Detach the webview from the overlay activity and finish it
+     */
+    override fun dismiss() = overlayActivity?.let { activity ->
```
dan-peluso, line 67: I think this would be covered by the weak reference, but I just want to confirm that this does clear the reference once it's killed

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoPresentationManager.kt
```diff
@@ -24,34 +31,45 @@ internal class KlaviyoPresentationManager() : PresentationManager {
      *
      * TODO handle rotation better, including enum or typealias for orientation.
      */
-    private fun onActivityEvent(event: ActivityEvent) {
-        if (event is ActivityEvent.ConfigurationChanged) {
+    private fun onActivityEvent(event: ActivityEvent) = when (event) {
```
dan-peluso, line 34: Do we care about doing anything in the other states here? Like closing the form on pause or etc

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoFormsOverlayActivity.kt
evan-masseau: This was intentional, dismiss will kill the activity. In the future we won't even use dismiss here, we will issue JS to close the form that way (maybe with fallback timer in place in case of nonresponsive webview)

sdk/core/src/test/java/com/klaviyo/core/utils/TakeIfTest.kt
evan-masseau: AI would never

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoPresentationManager.kt
evan-masseau: Possibly, this could grow with further testing, but this was all that seemed to matter now

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoPresentationManager.kt
evan-masseau: Great catch, `finish()` alone was not automatically nulling out the weak reference, so I will explicitly null it out instead. I think maybe the weak reference woulda allowed that activity to be GC'd, but best to release it explicitly!

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoPresentationManager.kt
```diff
@@ -24,34 +31,45 @@ internal class KlaviyoPresentationManager() : PresentationManager {
...
-            )
-        } ?: run {
-            Registry.log.warning("Unable to show IAF - null activity reference")
-        }
+    /**
+     * Launch the overlay activity
+     */
+    override fun present(formId: String?) = presentationState.takeIf<PresentationState.Hidden>()?.let {
+        presentationState = PresentationState.Presenting(formId)
+        Registry.config.applicationContext.startActivity(KlaviyoFormsOverlayActivity.launchIntent)
```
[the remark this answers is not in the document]
evan-masseau: tried inspecting source code, I think `applicationContext.startActivity` might just be a proxy for `activity.startActivity`?? In any case it doesn't appear to make a difference, and I don't have deal with the fact that `currentActivity` is nullable.

sdk/forms/src/test/java/com/klaviyo/forms/presentation/KlaviyoPresentationManagerTest.kt
```diff
@@ -9,64 +11,200 @@ import com.klaviyo.forms.webview.WebViewClient
...
+        verify { spyLog.debug("No-op dismiss: overlay activity is not presented") }
     }
 
     @Test
-    fun `dismiss should fail gracefully when activity is null`() {
-        every { mockLifecycleMonitor.currentActivity } returns null
-        KlaviyoPresentationManager().dismiss()
-        verify(exactly = 0) { mockWebViewClient.detachWebView(mockActivity) }
-        verify { spyLog.warning("Unable to dismiss IAF - null activity reference") }
+    fun `test full presentation cycle`() {
```
dan-peluso, line 177: great coverage here 👍

dan-peluso (approved):

dan-peluso:
Only follow up on this is - do we need to remove anything from react native or internal activity tracking for this? My suspicion is no, we'll probably need a lot of that anyhow, but I do recall adding a bunch of other internal things when we had to work around the 'track the activity' constraint.

PR #264, opened 2025-05-16
IAF | SDK -> Onsite JS communication layer

evan-masseau:
# Description

Adds interfaces and implementations to support JS injection for SDK -> Klaviyo.js communication per [spec doc.](https://docs.google.com/document/d/1MVhfEzdoNW6yYC1NBo1O8dUXxNngRZGU79KyOtHI-XM/edit?tab=t.0#heading=h.ah2wdjbm4de1) 
In the next PR, I will add profile and lifecycle observers to react to changes using this service. 

## Due Diligence

  - This PR only adds the interfaces and implementations, but this was built off a PoC in which I tested this pattern does work for injecting JS commands.
  - Not time yet

## Changelog / Code Overview

- Added JS functions for setting profile data attribute and dispatching lifecycle events in a separate JS file
- Load that into the webview by file url
- Kotlin interface/class to serve as an API for evaluating those functions 
- Added `JavascriptEvaluator` to decouple JS evaluation from owning a ref to the webview
- Implemented that interface on the webview client. 
 

## Test Plan

Requires some code modification bc this PR is only meant to add the services, without implementing the observers. To verify that the JS injection is working: 
- Modify `KlaviyoWebViewClient.onJsHandshakeCompleted` to add a call to `Registry.get<OnsiteBridge>().setProfile` or `Registry.get<OnsiteBridge>().dispatchLifecycleEvent` e.g.:
```
        Registry.get<OnsiteBridge>().dispatchLifecycleEvent(
            OnsiteBridge.LifecycleEventType.foreground,
            OnsiteBridge.LifecycleSessionBehavior.restore
        )
```
- Run the test app, open a form
- Open up chrome remote debugger and verify your JS was run -- for example use setProfile and then check that the data-attribute was correctly updated. Or add a console.log to the lifecycleEvent JS method and verify that logged.

## Related Issues/Tickets
 
[Spec](https://docs.google.com/document/d/1MVhfEzdoNW6yYC1NBo1O8dUXxNngRZGU79KyOtHI-XM/edit?tab=t.0#heading=h.ah2wdjbm4de1)
[CHNL-19932](https://klaviyo.atlassian.net/browse/CHNL-19932)


amber-klaviyo (approved):
I'm not the assigned reviewer- I'm just trying to get a better understanding of some of the IAF and onsite changes that are being worked on

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
```diff
@@ -184,6 +190,26 @@ internal class KlaviyoWebViewClient(
...
     }
 
+    /**
+     * Evaluate JavaScript in the webview, if possible
+     */
+    override fun evaluateJavascript(
+        javascript: String,
+        callback: (Boolean) -> Unit
+    ) = webView?.let { webView ->
+        Registry.lifecycleMonitor.currentActivity?.runOnUiThread {
```
dan-peluso, line 200: Small thing but is this guaranteed to execute after something like a 'background' event is sent? Maybe I'm confusing the order of operations here, but my concern is that we might not be able to run something on the UI thread if we're in a paused or killed state

sdk/forms/src/main/java/com/klaviyo/forms/bridge/OnsiteBridge.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.klaviyo.analytics.model.ImmutableProfile
+
+/**
+ * API for communicating data and events from native to the onsite-in-app JS module
+ */
+@Suppress("EnumEntryName", "ktlint:enum-entry-name-case")
+internal interface OnsiteBridge {
+
+    enum class LifecycleEventType {
+        background,
```
dan-peluso, line 12: nit - any reason we're not following all caps enum standards?

evan-masseau: simplified the code because the js function names and keywords are lowercase too.

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
evan-masseau: In my testing so far, runOnUiThread worked while the app was backgrounded (android really lets you do whatever huh...)

dan-peluso (approved):

PR #265, opened 2025-05-16
IAF | Implements App Session with Profile, Lifecycle, Company ID observers

evan-masseau:
# Description
This builds on the previous PR #264 to start observing events (e.g. lifecycle events and profile state changes) to be injected  into the WebView. This adds the interfaces and organizational code for those observers. To try to break things up a bit, I will have subsequent PRs to add observers for lifecycle events, profile changes, and company ID change individually. 

## Due Diligence

## Changelog / Code Overview

- Had to add a new bridge message type for `jsReady` to deal with a timing issue. The JS isn't immediately ready when we load in the HTML string, or even in the `onPageStarted` callback method. But waiting for the handshake is too late to inject the initial profile data. 
- `Observer` interface for individual observer types (e.g profile data, analytics events, lifecycle)
- `ObserverCollection` interface to collect and manage all observers
- Updated webview client to start observers on handShook, and stop them before webview is destroyed.
- Extracted a `HandshakeSpec` data class rather than coding `JSONObjects` directly. This way I can assign handshake type/version in the `Observer` instances, where the logic actually lives.

Plus, these branches were built on top of this, and merged in:
- Profile observers #266
- Lifecycle observers #267 
- In-app forms session implementation #270, #271 

## Test Plan

The base work in this PR was organizational, and sufficiently covered by unit tests and a basic regression cycle. 
Each PR built on top of it had its own testing steps that I'm going to copy over to be tested as an integrated branch:

### Segmentation:
To your klaviyo test account, add a web form that is targeted to a segment/list. Note an email or other identifier of a profile in that segment that should see the form. You can use the iaf-tester.html to confirm this web form does appear to that profile.
Build android test app with composite build pointing at this branch of the SDK.
In the test app, input the profile identifier from above. On the forms page, change form environment to web.
Trigger a form -- you should see your targeted form!

### App Session
- In the test app code, call `registerForInAppForms` with the session timeout set to something low like 10 seconds.
- Create a form that appears on every app open (display frequency of 0 days). 
- Launch the app and open a form. 
- Background the app
- Wait 11s
- Foreground the app -- the form should appear again because your app session ended

### Unregister method
To test the new unregister method, add a button to the test app UI that calls unregister. 
- Create a form that appears on every app open (display frequency of 0 days). 
- Launch app -- form should open
- Tap Unregister and then register again
- Form should open again

### Misc
State change observer implementation changed a bit under the hood. To verify it is still working: 
Make an API call with an invalid identifier -- StateSideEffects is responsible for catching the 400 error and removing that value from state so you should see this value removed from the test app UI
Re-initialize with a different company ID -- you should see an API call to unregister push token
Repeat step 1 -- should get the same result as before

## Related Issues/Tickets
 

sdk/forms/src/main/assets/onsite-bridge.js
```diff
@@ -65,3 +65,12 @@ window.dispatchAnalyticsEvent = function (metric, strProperties) {
...
 }
+
+// Notify the SDK over the native bridge that these local JS scripts are initialized
+var bridgeName = document.head.getAttribute("data-native-bridge-name") || ""
+
+if (window[bridgeName] && window[bridgeName].postMessage) {
+    window[bridgeName].postMessage(JSON.stringify({type: "jsReady"}))
+} else {
+    console.error("Unknown bridge name: " + bridgeName)
+}
```
evan-masseau, line 76: Had to roll my own sorta lifecycle event to know when this JS is loaded.

sdk/forms/src/main/java/com/klaviyo/forms/bridge/Observer.kt
```diff
@@ -0,0 +1,9 @@
+package com.klaviyo.forms.bridge
+
+/**
+ * General purpose observer interface for data/events that we inject into the webview
+ */
+interface Observer {
+    fun startObserver(): Unit
+    fun stopObserver()
+}
```
evan-masseau, line 9: Naming is hard, might revisit some of these names before merging.

evan-masseau: TODO: change visibility to internal on the interfaces and implementations!

sdk/forms/src/main/java/com/klaviyo/forms/bridge/BridgeMessage.kt
```diff
@@ -153,7 +168,7 @@ internal sealed class BridgeMessage {
          * Parse out the android platform deep link
          */
         private fun JSONObject.getDeepLink(): String {
-            val routeString = getString("android")
+            val routeString = optString("android")
```
evan-masseau, line 171: Noticed while updating this, that getString throws, so the intended exception type below is never thrown

sdk/forms/src/main/java/com/klaviyo/forms/bridge/BridgeMessage.kt
```diff
@@ -134,16 +151,14 @@ internal sealed class BridgeMessage {
...
         private fun JSONObject.getEventProperties(): Map<EventKey, Serializable> {
             val map = mutableMapOf<EventKey, Serializable>()
             val propertyMap = getJSONObject("properties")
             propertyMap.keys().forEach {
-                try {
-                    map[EventKey.CUSTOM(it)] = propertyMap.get(it) as Serializable
-                } catch (e: Exception) {
-                    Registry.log.error("Failed to write property $it to profile event payload", e)
+                (propertyMap.opt(it) as? Serializable)?.let { value ->
+                    map[EventKey.CUSTOM(it)] = value
```
evan-masseau, line 161: The try/catch was used to skip non-serializable or null values. Using `opt` achieves the same thing without try/catch

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
```diff
@@ -54,6 +55,9 @@ internal class KlaviyoWebViewClient(
         Registry.log.debug("Klaviyo webview is already initialized")
     } ?: KlaviyoWebView().let { webView ->
         val nativeBridge: BridgeMessageHandler = Registry.get()
+        val observerCollection = Registry.get<ObserverCollection>()
+        val handshake = nativeBridge.handshake + observerCollection.handshake()
```
evan-masseau, line 59: Handshake is now a compiled list of 
- Native bridge messages, from JS -> SDK
- Injected data from SDK -> JS.

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
```diff
@@ -71,7 +75,7 @@ internal class KlaviyoWebViewClient(
             .replace("SDK_NAME", Registry.config.sdkName)
             .replace("SDK_VERSION", Registry.config.sdkVersion)
             .replace("BRIDGE_NAME", nativeBridge.name)
-            .replace("BRIDGE_HANDSHAKE", handShakeData)
+            .replace("BRIDGE_HANDSHAKE", handshake.compileJson())
```
evan-masseau, line 78: TODO should have a test covering this whole chained block of `replace`s

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
```diff
@@ -199,7 +208,7 @@ internal class KlaviyoWebViewClient(
     ) = webView?.let { webView ->
         Registry.lifecycleMonitor.currentActivity?.runOnUiThread {
             webView.evaluateJavascript(javascript) { result ->
-                callback(result === "true")
```
evan-masseau, line 202: mistake from earlier PR, all JS callbacks were evaluating false here despite succeeding -_-

sdk/forms/src/test/java/com/klaviyo/forms/bridge/KlaviyoBridgeMessageHandlerTest.kt
```diff
@@ -75,6 +73,24 @@ internal class KlaviyoBridgeMessageHandlerTest : BaseTest() {
...
     }
 
+    @Test
+    fun `allowed origin returns the base URL`() {
+        /**
+         * @see com.klaviyo.forms.bridge.KlaviyoBridgeMessageHandler.allowedOrigin
+         */
+        val expected = setOf(Registry.config.baseUrl)
+        assertEquals(expected, bridgeMessageHandler.allowedOrigin)
+    }
```
evan-masseau, line 83: unrelated, just adding a line of coverage :D

sdk/forms/src/main/java/com/klaviyo/forms/InAppForms.kt
```diff
@@ -25,8 +28,13 @@ fun Klaviyo.registerForInAppForms(
     Registry.apply {
         registerOnce<PresentationManager> { KlaviyoPresentationManager() }
         registerOnce<BridgeMessageHandler> { KlaviyoBridgeMessageHandler() }
-        registerOnce<WebViewClient> { KlaviyoWebViewClient(config) }
+        registerOnce<WebViewClient> {
+            KlaviyoWebViewClient(config).apply {
+                Registry.register<JavaScriptEvaluator>(this)
```
dan-peluso, line 33: nit - you could add a name to the scope of the registry on line 28 and avoid having to call `Registry` again here

sdk/forms/src/main/java/com/klaviyo/forms/bridge/BridgeMessage.kt
```diff
@@ -3,91 +3,106 @@ package com.klaviyo.forms.bridge
...
 import com.klaviyo.analytics.networking.requests.AggregateEventPayload
-import com.klaviyo.core.Registry
 import java.io.Serializable
-import org.json.JSONArray
 import org.json.JSONObject
 
 /**
+ * Encapsulates messages sent from JS to SDK via the NativeBridge, i.e. [BridgeMessageHandler].
  * This should be updated with any new message types we add coming from the onsite-in-app-forms
+ * By convention, class names should be same as the message type, but in lower camel case
```
dan-peluso, line 12: smidge confusing - you mean the message type is in lower camel case, not the message type right?

sdk/forms/src/main/java/com/klaviyo/forms/bridge/BridgeMessage.kt
```diff
@@ -3,91 +3,106 @@ package com.klaviyo.forms.bridge
...
+     */
     data object HandShook : BridgeMessage()
 
-    data class Show(
+    /**
+     * Sent from the onsite-in-app-forms when a form is about to appear as a signal to present the webview
+     *
+     * @param formId The form ID of the form that is appearing
+     */
+    data class FormWillAppear(
```
dan-peluso, line 32: lol tb to this debate

sdk/forms/src/main/java/com/klaviyo/forms/bridge/BridgeMessage.kt
```diff
@@ -3,91 +3,106 @@ package com.klaviyo.forms.bridge
...
         val event: Event
     ) : BridgeMessage()
 
-    data class DeepLink(
+    /**
+     * Sent from the onsite-in-app-forms when a deep link is opened
+     *
+     * @param route The deep link route to be opened (usually a URL)
+     */
+    data class OpenDeepLink(
```
dan-peluso, line 61: Just wanna be clear I am down for the verb + message format of these 👍

sdk/forms/src/main/java/com/klaviyo/forms/bridge/BridgeMessage.kt
```diff
@@ -3,91 +3,106 @@ package com.klaviyo.forms.bridge
...
+     * @param formId The form ID of the form that is disappearing
+     */
+    data class FormDisappeared(
         val formId: String
     ) : BridgeMessage()
 
+    /**
+     * Sent from the onsite-in-app-forms when an irrecoverable error occurs and the webview should be closed
+     *
+     * @param reason
```
dan-peluso, line 77: nit do we need this if we don't actually have an explanation or is it redundant? (or maybe this was left in by mistake?)

sdk/forms/src/main/java/com/klaviyo/forms/bridge/BridgeMessage.kt
```diff
@@ -3,91 +3,106 @@ package com.klaviyo.forms.bridge
...
-                        IAF_TYPE_VERSION_KEY to 1
-                    ),
-                    mapOf(
-                        IAF_MESSAGE_TYPE_KEY to IAF_MESSAGE_TYPE_ABORT,
-                        IAF_TYPE_VERSION_KEY to 1
-                    )
-                )
-            ).toString()
+            listOf(
+                HandshakeSpec(keyName<HandShook>(), 1),
```
dan-peluso, line 98: great improvement here

sdk/forms/src/main/java/com/klaviyo/forms/bridge/BridgeMessage.kt
dan-peluso: Except that we lose the exception log right? Do we care about that?

sdk/forms/src/main/java/com/klaviyo/forms/bridge/Observer.kt
```diff
@@ -0,0 +1,21 @@
+package com.klaviyo.forms.bridge
+
+/**
+ * General purpose interface for observing events and passing data into the webview
+ */
+internal interface Observer {
```
dan-peluso, line 6: Are you dead set on this name? I feel like observer may be _too_ general, especially since it's an incredibly popular OOP pattern. Maybe something like WebEventObserver? Or maybe something between that and just plain observer?

sdk/forms/src/main/java/com/klaviyo/forms/bridge/ObserverCollection.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    /**
+     * Compiles the handshake data from all observers into a single list
+     */
+    val handshake: List<HandshakeSpec> get() = observers.map { it.handshake }
+
+    /**
+     * Start all observers in the collection
+     */
+    fun startObservers() = observers.forEach { it.startObserver() }
```
dan-peluso, line 20: Do we care if this has a return type of the list of observers? The equals sign is clean but if we can avoid returning the list (unless we'll use it later) that might help confusion down the line

sdk/forms/src/main/java/com/klaviyo/forms/bridge/ObserverCollection.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    /**
+     * Stop all observers in the collection
+     */
+    fun stopObservers() = observers.forEach { it.stopObserver() }
```
dan-peluso, line 25: same comment here

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
```diff
@@ -39,21 +40,17 @@ internal class KlaviyoWebViewClient(
...
     /**
      * Initialize a webview instance, with protection against duplication
      * and initialize klaviyo.js for in-app forms with handshake data injected in the document head
      */
     override fun initializeWebView(): Unit = webView?.let {
         Registry.log.debug("Klaviyo webview is already initialized")
     } ?: KlaviyoWebView().let { webView ->
-        val nativeBridge: BridgeMessageHandler = Registry.get()
+        val nativeBridge = Registry.get<BridgeMessageHandler>()
+        val observerCollection = Registry.get<ObserverCollection>()
```
dan-peluso, line 51: So do these throw if we don't have them, or would we expect this to continue and just not function? Wondering if we want to try/catch and log or do some sort of check before moving on

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
```diff
@@ -39,21 +40,17 @@ internal class KlaviyoWebViewClient(
...
+        val handshake = nativeBridge.handshake + observerCollection.handshake
```
dan-peluso, line 52: could we add a type to this to make it a smidge easier to read (sorry doing this from github web I know it probably looks nicer on AS but the plus is throwing me off)

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
```diff
@@ -136,6 +137,7 @@ internal class KlaviyoWebViewClient(
      */
     override fun destroyWebView() = apply {
         handshakeTimer?.cancel()
+        Registry.get<ObserverCollection>().stopObservers()
```
dan-peluso, line 140: same question about this potentially being null - generally just trying to be safe around us properly disposing of these objects in memory before we remove them from the registry or they stop becoming relevant to us (like  maybe after we call `stopObservers`

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
dan-peluso: L

sdk/forms/src/test/java/com/klaviyo/forms/InAppFormsTest.kt
```diff
@@ -44,12 +53,18 @@ internal class InAppFormsTest : BaseTest() {
...
+        assert(!Registry.isRegistered<JavaScriptEvaluator>())
+        assert(!Registry.isRegistered<OnsiteBridge>())
+        assert(!Registry.isRegistered<ObserverCollection>())
 
         Klaviyo.registerForInAppForms()
 
-        assert(Registry.isRegistered<PresentationManager>())
-        assert(Registry.isRegistered<BridgeMessageHandler>())
-        assert(Registry.isRegistered<WebViewClient>())
+        assertNotNull(Registry.get<PresentationManager>())
```
dan-peluso, line 62: nit (and definitely don't have to do this in this PR or maybe ever) it'd be cool to make these less verbose with some sort of scoping method so that we can make these look cleaner
```
assertRegistered {
    PresentationManager::class.java
    BridgeMessageHandler::class.java
    ...
```

just thinking we do this all the time in unit tests

sdk/forms/src/test/java/com/klaviyo/forms/bridge/KlaviyoBridgeMessageHandlerTest.kt
dan-peluso: This battle is won in inches 👍

sdk/forms/src/main/java/com/klaviyo/forms/bridge/Observer.kt
evan-masseau: No, I meant to change this, lost a comment somewhere along the way. Agreed it's too generic

sdk/forms/src/main/java/com/klaviyo/forms/bridge/ObserverCollection.kt
evan-masseau: Good call, I often forget about return type when I do a short hand like this

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
evan-masseau: They throw, I will double check but I think it is caught before it can crash the host app though.

sdk/forms/src/test/java/com/klaviyo/forms/InAppFormsTest.kt
evan-masseau: I'm a big fan of readability

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
evan-masseau: Yeah I can check this one too. The pattern I took here though is that the observer classes live on if stopped, so even after a stop, it'll remain in the registry, it just isn't listening until started again

sdk/forms/src/test/java/com/klaviyo/forms/bridge/KlaviyoBridgeMessageHandlerTest.kt
evan-masseau: I learned how to inspect the branch coverage, not good for my mental health but great for our coverage report

sdk/forms/src/main/java/com/klaviyo/forms/bridge/BridgeMessage.kt
evan-masseau: I think no, since these events only come from the webview, logging an error here would be completely out of the control of the developer, so it'd really only be for us. Besides which, this property map came from a JSON string in the first place, so it _must_ be serializable data. Its really just a formality to make the compiler happy.

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
evan-masseau: OK Yeah because the entrypoint to in-app forms, `Klaviyo.registerForInAppForms` is wrapped in `safeApply`, the host app is guarded against Registry exceptions. The general idea is that it is incumbent upon an entrypoint method like `initialize` or `registerForInAppForms` to setup the package, e.g. by handling registration/initialization of dependencies. That should mean that the individual services don't have to have a whole lot of repetitive checks of whether a service is yet registered. I've gone back and forth on converting to more of a constructor injection DI pattern, where instead of using `Registry.get` inline like I am here, dependent services would be a constructor argument defaulted to `Registry.get<ThatInterface>()`. The advantage being the whole instance can't even get into a weird partial initialization state that way. TBH the only reason I haven't done it yet is lack of time. In some cases it can create dependency cycles too, but usually that is solvable by splitting interfaces up further.

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
evan-masseau: See above answer for more musings, but short answer is this is safe 1 - bc we have a single point of entry that handles registrations, and 2 - bc exceptions are handled internally even if there were a registry error.

sdk/forms/src/test/java/com/klaviyo/forms/InAppFormsTest.kt
evan-masseau: Like the idea, and I took a crack, but can't figure a way to do it because all the Registry methods are generic, and I'm not sure how to call `Registry.get<someClass>()` where `someClass` is an actually a variable of type `Class` or event `KType`.

sdk/forms/src/test/java/com/klaviyo/forms/bridge/KlaviyoBridgeMessageHandlerTest.kt
evan-masseau: Just to share this wonderful finding... click on the color block in the gutter here and it'll pop up in-depth coverage for that line, can help figure out what logic branches you missed.
Screenshot 2025-05-20 at 21 38 40

sdk/forms/src/main/java/com/klaviyo/forms/bridge/Observer.kt
evan-masseau: I think I'll go with something like `BridgeObserver`, but I also want to rename a couple of classes within the `forms.bridge` subpackages, so I'm gonna pinky-swear to come back to this before it gets merged fully, since I have a couple PRs based off this branch now. I'll do the renaming once those get flattened into one.

dan-peluso: pinky swear protocol, I dig it

sdk/forms/src/test/java/com/klaviyo/forms/InAppFormsTest.kt
dan-peluso: Ah fair, yeah generics in DI I feel like nobody has really gotten right. Thanks for trying!

evan-masseau: yeah the only option i saw was to do add a different registry method using an argument instead of generics, like `get(type: KType)` which should work, but doesn't seem worth modifying the real code for tests. Could be something to keep in mind for future if a need arises though.

sdk/forms/src/main/java/com/klaviyo/forms/bridge/Observer.kt
evan-masseau: Renames: https://github.com/klaviyo/klaviyo-android-sdk/pull/268

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -37,14 +37,18 @@ object Klaviyo {
...
+     * This registration is a lambda invoked when the service is required, not instantiated now
+     */
+    private fun initializeServices() = Registry.apply {
+        registerOnce<ApiClient> { KlaviyoApiClient }
+        registerOnce<State> {
+            KlaviyoState().also { state ->
+                register<StateSideEffects>(StateSideEffects(state))
+            }
+        }
     }
```
evan-masseau, line 52: Previously I was using the one-time nature of an `init` (called the first time you access Klaviyo, usually to initialize) to have APIClient register only once. Now that we have `RegisterOnce`, I don't have to use an Object initializer, which I tend to avoid anyway.

sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
```diff
@@ -57,26 +62,61 @@ internal class KlaviyoState : State {
...
-        mutableListOf<StateObserver>()
+    private val stateChangeObservers = Collections.synchronizedList(
+        CopyOnWriteArrayList<StateChangeObserver>()
     )
 
+    /**
+     * Maps deprecated [StateObserver]s to the new [StateChangeObserver]
+     * to maintain backwards compatibility while moving on to the new data type
+     */
+    private val deprecatedObserverMap = mutableMapOf<StateObserver, StateChangeObserver>()
```
evan-masseau, line 73: I'm pretty proud of this one

dan-peluso: Do we have any plans on testing backward compatibility? Might not be a bad idea to add something to the testing doc (knowing that yes it will be very annoying to test)

evan-masseau: 100% yes, expecting to test with the old test app's code but pointed to the new branch, and verifying test app behaviors that rely on this observer.

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -83,10 +89,6 @@ object Klaviyo {
 
         Registry.get<ApiClient>().startService()
 
-        Registry.register<State>(KlaviyoState())
-        Registry.getOrNull<StateSideEffects>()?.detach()
```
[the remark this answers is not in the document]
evan-masseau: Detaching here was an artifact of creating a new State instance. Under the hood, state is actually unchanged, since by creating a new instance it was really just re-loading state from disk, but all properties in state are automatically backed up to disk on set anyways. So now that I'm not re-starting state, I don't need to restart these observers either.

sdk/core/src/main/java/com/klaviyo/core/Registry.kt
```diff
@@ -178,18 +172,18 @@ object Registry {
...
+                services[type] = lazyService
+                return lazyService
             }
 
             is Any -> throw InvalidRegistration(type)
             else -> {
                 if (type == typeOf<Config>()) {
                     throw MissingConfig()
                 } else {
-                    throw throw MissingRegistration(type)
```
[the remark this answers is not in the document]
evan-masseau: I know right? I think it just threw and the extra throw was superfluous. The IDE called it "unreachable code" probs because it throws before it can... throw again? lol

sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
evan-masseau: And BTW I did do this test in the original PR that added this change. Just want to re-do similar tests on the fully merged forms 1.5 changeset.

PR #266, opened 2025-05-19
IAF | Profile observers

evan-masseau:
# Description
Building on #265, adds profile observer implementation

## Due Diligence

## Changelog / Code Overview

- Adds `ProfileObserver` to inject identifiers into the webview
  - When webview first initializes
  - Whenever identifiers change in state

## Test Plan

To your klaviyo test account, add a web form that is targeted to a segment/list. Note an email or other identifier of a profile in that segment that should see the form. You can use the `iaf-tester.html` to confirm this web form does appear to that profile. 
Build android test app with composite build pointing at this branch of the SDK.
In the test app, input the profile identifier from above. On the forms page, change form environment to web.
Trigger a form -- you _should_ see your targeted form!

## Related Issues/Tickets
 
### DIFF
--- a/sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoObserverCollection.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoObserverCollection.kt
@@ -4,8 +4,10 @@ package com.klaviyo.forms.bridge
  * Manages the collection of observers that inject data into the webview
  */
 internal class KlaviyoObserverCollection : ObserverCollection {
-    override val observers: List<Observer> = listOf(
-        LifecycleObserver(),
-        ProfileObserver(),
-    )
+    override val observers: List<Observer> by lazy {
+        listOf(
+            LifecycleObserver(),
+            ProfileObserver(),
+        )
+    }
 }
--- a/sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoObserverCollection.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoObserverCollection.kt
@@ -5,6 +5,7 @@ package com.klaviyo.forms.bridge
  */
 internal class KlaviyoObserverCollection : ObserverCollection {
     override val observers: List<Observer> = listOf(
-        LifecycleObserver()
+        LifecycleObserver(),
+        ProfileObserver(),
     )
 }
--- a/sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoObserverCollection.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoObserverCollection.kt
@@ -7,7 +7,7 @@ internal class KlaviyoObserverCollection : ObserverCollection {
     override val observers: List<Observer> by lazy {
         listOf(
             LifecycleObserver(),
-            ProfileObserver(),
+            ProfileObserver()
         )
     }
 }
--- a/sdk/forms/src/main/java/com/klaviyo/forms/bridge/ProfileObserver.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/bridge/ProfileObserver.kt
@@ -1,6 +1,7 @@
 package com.klaviyo.forms.bridge
 
 import com.klaviyo.analytics.model.Keyword
+import com.klaviyo.analytics.model.ProfileKey
 import com.klaviyo.analytics.state.State
 import com.klaviyo.core.Registry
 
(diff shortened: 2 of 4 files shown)


ajaysubra (approved):

evan-masseau:
@ab1470 I tagged you too, since you're assigned the ios equivalent of this work

sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoObserverCollection.kt
```diff
@@ -1,12 +1,10 @@
...
- *  - Profile identifiers
- *  - Lifecycle events
- *  - Analytics events
- *  - Host App navigation events
+ * Manages the collection of observers that inject data into the webview
  */
 internal class KlaviyoObserverCollection : ObserverCollection {
-    override val observers: List<Observer> = emptyList()
+    override val observers: List<Observer> = listOf(
+        ProfileObserver()
```
dan-peluso, line 8: Should we do this lazily or as a computed value?

evan-masseau: Lazily would be ok, computed would not because we don't want multiple instances. But lazily didn't feel necessary since instantiating these doesn't really do anything, not until they've been `start`ed.

sdk/forms/src/main/java/com/klaviyo/forms/bridge/ProfileObserver.kt
```diff
@@ -0,0 +1,38 @@
...
+    )
+
+    override fun startObserver() = Registry.get<State>().onStateChange(::onStateChange).also {
+        // Set initial profile identifiers on startup
+        onStateChange(null, null)
+    }
+
+    override fun stopObserver() = Registry.get<State>().offStateChange(::onStateChange)
+
+    private fun onStateChange(key: Keyword?, v: Any?) = when (key?.name) {
```
dan-peluso, line 23: doesn't seem like we're using `v` here? Also if we do keep it could we name it `value`

sdk/forms/src/main/java/com/klaviyo/forms/bridge/ProfileObserver.kt
```diff
@@ -0,0 +1,38 @@
...
+        in IDENTIFIERS, null -> Registry.get<OnsiteBridge>().setProfile(
+            Registry.get<State>().getAsProfile()
+        )
+        else -> Unit
+    }
+
+    companion object {
+        private val IDENTIFIERS = listOf(
+            "external_id",
```
dan-peluso, line 32: Are we able to pull these from the analytics package instead of inlining them? Not sure if they're declared as constants there either, but I believe we're already depending on that module

sdk/forms/src/main/java/com/klaviyo/forms/bridge/ProfileObserver.kt
```diff
@@ -0,0 +1,38 @@
...
+ */
+internal class ProfileObserver : Observer {
+    override val handshake: HandshakeSpec = HandshakeSpec(
+        type = "profileMutation",
+        version = 1
+    )
+
+    override fun startObserver() = Registry.get<State>().onStateChange(::onStateChange).also {
+        // Set initial profile identifiers on startup
+        onStateChange(null, null)
```
dan-peluso, line 18: This has a bit of a smell to it - I'm trying to trace the code and see what happens here
we call the state change with nulls -> it then sets the profile to the one we have in the registry, and then every time we get a state change in the profile we also update that observer? Maybe instead of calling the state change helper we could just directly call the set profile stuff from the registry, then we could also remove the 'IDENTIFIERS,null` from line 24. I do appreciate the code reuse, but it does feel a smidge like we're making the contract potentially unsafe for the sake of this init (also tricky to follow).

sdk/forms/src/main/java/com/klaviyo/forms/bridge/ProfileObserver.kt
dan-peluso: Gah I wish we had 'private to the repo' type variables in Android :rip:

sdk/forms/src/main/java/com/klaviyo/forms/bridge/ProfileObserver.kt
dan-peluso: But only if we want to use it with the `::functonCall` notation right? I think we could call it as a lamda and avoid having this arg

sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoObserverCollection.kt
dan-peluso: Yeah, it's pretty lightweight so definitely a nit. I just didn't want to bloat the collection file until we needed it

evan-masseau: Lazy is fine 👍

sdk/forms/src/main/java/com/klaviyo/forms/bridge/ProfileObserver.kt
evan-masseau: I don't love the null thing either, it represents when profile is reset, hence no individual key was changed. I didn't want to get into a whole refactor of the State observer right now though. I do see your point about calling it with null here though, I think pulling out an additional helper method instead of calling the callback method would help readability.

sdk/forms/src/main/java/com/klaviyo/forms/bridge/ProfileObserver.kt
evan-masseau: Actually, I could export them as a list `analytics`, I didn't think of that before.

dan-peluso (approved):

PR #267, opened 2025-05-20
IAF | Lifecycle observers to support foreground/backgrounding tracking

evan-masseau:
# Description

Builds on #265 to add specific implementation of lifecycle observers for tracking foregrounding/backgrounding of the host app. Partially implements app session tracking, the rest of that implementation will be in another PR to keep things more reviewable. 

## Due Diligence

## Changelog / Code Overview

- Updated the lifecycle monitor in core to also track when an app is foregrounded, based on activity tracking
- Added forms observer for foreground/background events to pass those into the webview
- Partial session tracking implementation, to refresh the webview on session timeout

## Test Plan

Not much to test yet because app session isn't fully implemented and fender isn't doing anything with lifecycle event yet. 
But we can test against this [fender branch](https://github.com/klaviyo/fender/pull/42121) just to confirm that the lifecycle event is detected and doesn't console.log a validation error!

## Related Issues/Tickets
 
[CHNL-19932](https://klaviyo.atlassian.net/browse/CHNL-19932) 
 

[CHNL-19932]: https://klaviyo.atlassian.net/browse/CHNL-19932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
@@ -22,9 +22,15 @@ internal object KlaviyoLifecycleMonitor : LifecycleMonitor, Application.Activity
     )
 
     override var currentActivity: Activity? by WeakReferenceDelegate()
+        private set
 
     @AdvancedAPI
     override fun assignCurrentActivity(activity: Activity) {
+        if (activity != currentActivity) {
+            // If we missed this activity's creation, then we need to increment the count now
+            activeActivities++
+        }
+
         currentActivity = activity
     }
 
@@ -50,6 +56,10 @@ internal object KlaviyoLifecycleMonitor : LifecycleMonitor, Application.Activity
     }
 
     override fun onActivityStarted(activity: Activity) {
+        if (activeActivities == 0) {
+            broadcastEvent(ActivityEvent.FirstStarted(activity))
+        }
+
         activeActivities++
         broadcastEvent(ActivityEvent.Started(activity))
     }
--- a/sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
@@ -29,9 +29,8 @@ internal object KlaviyoLifecycleMonitor : LifecycleMonitor, Application.Activity
         if (activity != currentActivity) {
             // If we missed this activity's creation, then we need to increment the count now
             activeActivities++
+            currentActivity = activity
         }
-
-        currentActivity = activity
     }
 
     override fun onActivityEvent(observer: ActivityObserver) {
--- a/sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
@@ -12,7 +12,15 @@ sealed class ActivityEvent(val activity: Activity? = null, val bundle: Bundle? =
 
     val type: String get() = this.javaClass.simpleName
 
+    /**
+     * An activity was created
+     */
     class Created(activity: Activity, bundle: Bundle?) : ActivityEvent(activity, bundle)
+
+    /**
+     * Application moved to the foreground, i.e. went from 0 to 1 active activities
+     */
+    class FirstStarted(activity: Activity) : ActivityEvent(activity)
 
     class Started(activity: Activity) : ActivityEvent(activity)
 
@@ -24,6 +32,9 @@ sealed class ActivityEvent(val activity: Activity? = null, val bundle: Bundle? =
 
     class Stopped(activity: Activity) : ActivityEvent(activity)
 
+    /**
+     * Application moved to the background, called after the last application activity is stopped
+     */
     class AllStopped : ActivityEvent()
 
     class ConfigurationChanged(val newConfig: Configuration) : ActivityEvent()
--- a/sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
@@ -8,35 +8,61 @@ import com.klaviyo.core.utils.AdvancedAPI
 
 typealias ActivityObserver = (activity: ActivityEvent) -> Unit
 
+/**
+ * Represent different events emitted in response to lifecycle triggers from the host application
+ */
 sealed class ActivityEvent(val activity: Activity? = null, val bundle: Bundle? = null) {
 
+    /**
+     * Get the type of the event as a string (e.g. for logging)
+     */
     val type: String get() = this.javaClass.simpleName
 
     /**
-     * An activity was created
+     * Emitted when [Activity.onCreate] is called from an activity within the host app
      */
     class Created(activity: Activity, bundle: Bundle?) : ActivityEvent(activity, bundle)
 
     /**
-     * Application moved to the foreground, i.e. went from 0 to 1 active activities
+     * Emitted when [Activity.onStart] is called from an activity within the host app
+     */
+    class Started(activity: Activity) : ActivityEvent(activity)
+
+    /**
+     * Emitted when the host application moves to the foreground
+     * i.e. an activity [Started], and the application transitions from 0 to 1 started activity
      */
     class FirstStarted(activity: Activity) : ActivityEvent(activity)
 
-    class Started(activity: Activity) : ActivityEvent(activity)
-
+    /**
+     * Emitted when [Activity.onResume] is called from an activity within the host app
+     */
     class Resumed(activity: Activity) : ActivityEvent(activity)
 
+    /**
+     * Emitted when [Activity.onSaveInstanceState] is called from an activity within the host app
+     */
     class SaveInstanceState(activity: Activity, bundle: Bundle) : ActivityEvent(activity, bundle)
 
+    /**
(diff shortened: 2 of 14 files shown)


sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
```diff
@@ -22,9 +22,15 @@ internal object KlaviyoLifecycleMonitor : LifecycleMonitor, Application.Activity
     )
 
     override var currentActivity: Activity? by WeakReferenceDelegate()
+        private set
 
     @AdvancedAPI
```
dan-peluso, line 27: Lol which part of this is `AdvancedAPI`

sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
```diff
@@ -22,9 +22,15 @@ internal object KlaviyoLifecycleMonitor : LifecycleMonitor, Application.Activity
     )
 
     override var currentActivity: Activity? by WeakReferenceDelegate()
+        private set
```
dan-peluso, line 25: 👍

sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
```diff
@@ -22,9 +22,15 @@ internal object KlaviyoLifecycleMonitor : LifecycleMonitor, Application.Activity
...
 
     @AdvancedAPI
     override fun assignCurrentActivity(activity: Activity) {
+        if (activity != currentActivity) {
+            // If we missed this activity's creation, then we need to increment the count now
+            activeActivities++
```
dan-peluso, line 31: This has me a little worried that we're being too ambitious trying to track the number of active activities. Is the idea that this will only get fired once per session? Would overlayed activities or accidental activity creations (like if someone's bad code created multiple of the same activity) break this?

sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
```diff
@@ -22,9 +22,15 @@ internal object KlaviyoLifecycleMonitor : LifecycleMonitor, Application.Activity
     )
 
     override var currentActivity: Activity? by WeakReferenceDelegate()
+        private set
 
     @AdvancedAPI
     override fun assignCurrentActivity(activity: Activity) {
+        if (activity != currentActivity) {
```
dan-peluso, line 29: Should we be using the default .equals for comparing here or maybe the activity class name instead? Again worried about repeats

sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
```diff
@@ -12,8 +12,16 @@ sealed class ActivityEvent(val activity: Activity? = null, val bundle: Bundle? =
 
     val type: String get() = this.javaClass.simpleName
 
+    /**
+     * An activity was created
+     */
     class Created(activity: Activity, bundle: Bundle?) : ActivityEvent(activity, bundle)
```
dan-peluso, line 18: Why do we have a comment above this one but not the others below?

sdk/forms/src/main/assets/onsite-bridge.js
```diff
@@ -26,16 +26,14 @@ window.setProfile = function (external_id, email, phone_number, anonymous_id) {
  * Dispatches a lifecycle event to be detected by klaviyo.js
  *
  * @param type - The type of lifecycle event to dispatch: "foreground" | "background"
- * @param session - Session behavior: "persist" | "restore" | "purge"
```
dan-peluso, line 29: I probably missed what this did originally, but just curious why it's now removed

sdk/forms/src/main/java/com/klaviyo/forms/InAppForms.kt
```diff
@@ -26,10 +26,11 @@ fun Klaviyo.registerForInAppForms(
 ): Klaviyo = safeApply {
     // Register IAF services
     Registry.apply {
+        register<InAppFormsConfig>(config)
         registerOnce<PresentationManager> { KlaviyoPresentationManager() }
         registerOnce<BridgeMessageHandler> { KlaviyoBridgeMessageHandler() }
         registerOnce<WebViewClient> {
-            KlaviyoWebViewClient(config).apply {
+            KlaviyoWebViewClient().apply {
```
dan-peluso, line 33: rip to my one contribution here haha - just want to know why we're preferring to add the config to the registry instead of passing it directly.

sdk/forms/src/main/java/com/klaviyo/forms/bridge/LifecycleObserver.kt
```diff
@@ -0,0 +1,58 @@
...
+    override val handshake: HandshakeSpec = HandshakeSpec(
+        type = "lifecycleEvent",
+        version = 1
+    )
+
+    override fun startObserver() = Registry.lifecycleMonitor.onActivityEvent(::onLifecycleEvent)
+
+    override fun stopObserver() = Registry.lifecycleMonitor.offActivityEvent(::onLifecycleEvent)
+
+    private fun onLifecycleEvent(activity: ActivityEvent): Unit = when (activity) {
```
dan-peluso, line 24:
```suggestion
    private fun onLifecycleEvent(activity: ActivityEvent) {
     when (activity) {
```

sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
evan-masseau: oh because I started adding to all of them, and realized most were self explanatory

sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
evan-masseau: `assignCurrentActivity` is only meant to be used to set initial activity when `Klaviyo.initialize` isn't called from application.onCreate, it is the fix that I put into place for React Native.  It has an `AdvancedAPI` tag that raises a compiler error unless you opt in, as a warning to only use it if you really know what you're doing, besides also living in the `core` package. 

Here, I just noticed that unlike when we actually detect an activity from its create callback, we weren't incrementing the counter here. I could make this even more purpose-build (i.e. only for when we missed the first activity) by adding `&& activeActivities == 0`

From what I've researched, this approach of using a counter is a pretty common way to track activities and foregroung/background transition for the app.

sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
evan-masseau: see below answer

sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
evan-masseau: don't think so, because if it is a new activity that we missed tracking, we'd care about the distinct instance, not just its type.

sdk/forms/src/main/assets/onsite-bridge.js
evan-masseau: Earlier, we were going to have fender code move `sessionStorage` data in and out of `localStorage` as a way of having session data outlive the app being terminated. But then we decided to treat app termination similar to closing a tab in a web browser, so we didn't need this feature anymore.

sdk/forms/src/main/java/com/klaviyo/forms/bridge/LifecycleObserver.kt
evan-masseau: but-why-gif-1

sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
evan-masseau: I'll just go all in, why not.

sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
evan-masseau: I added some more to the method documentation in the interface, to really highlight when to use this and that a better alternative is to use the real lifecycle monitoring.

sdk/forms/src/main/java/com/klaviyo/forms/bridge/LifecycleObserver.kt
evan-masseau: Since all this method consists of is a `when` with 2 conditions, an expression body is more expedient. I prefer to avoid the added indentation of a block body where I can. I often tend toward an expression body because I like to break my code down into smaller functional units.

sdk/core/src/main/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitor.kt
dan-peluso: I only bring this up because I've had some nightmare Activity situations where navigation implementations lead to activities basically being made from scratch every time and not reused. Definitely more of an issue on the app devs writing bad activity code, but I do wonder if we can be smart around that sort of thing. Especially when activities can be launched basically at any point.

dan-peluso (approved):

PR #270, opened 2025-05-23
IAF | Persisting the webview after close

evan-masseau:
# Description

Finally updates the WebView to stay in memory after closing, and adds proper hooks to reinitialize on company ID change and lifecycle timeout.

## Due Diligence


sdk/core/src/main/java/com/klaviyo/core/utils/KlaviyoThreadHelper.kt
```diff
@@ -0,0 +1,26 @@
...
+        if (mainLooper == Looper.myLooper()) {
+            // Already on main thread, run immediately
+            block()
+        } else {
+            // Post to main thread
+            getHandler(mainLooper)
+                .post(block)
+        }
+    }
+}
```
evan-masseau, line 26: Simple threading helper, particularly convenient for mocking in unit tests.

This version of `runOnUiThread` is a little cleaner than earlier approaches I had. It also avoids re-dispatching if we are already on the main thread.
- Relying on`view.post` means the posted runner will only execute if the view is presented
- Reying on `currentActivity` requires mildly annoying optional chaining. 
- Both of these would re-dispatch even if called from the main thread.

sdk/forms/src/main/java/com/klaviyo/forms/InAppForms.kt
```diff
@@ -41,3 +44,28 @@ fun Klaviyo.registerForInAppForms(
...
+ * Resets the in-app forms listeners with the current configuration.
+ */
+@UiThread
+internal fun Klaviyo.reInitializeInAppForms() = safeApply {
+    // If config is missing, initial registration has not occurred so we can't re-initialize
+    Registry.getOrNull<InAppFormsConfig>()?.let { config ->
+        unregisterInAppForms()
+        registerForInAppForms(config)
+    }
+}
```
evan-masseau, line 71: I tried putting this re-initializing code in a couple different places before settling on an internal extension function.

First in the `PresentationManager`, because dismissing made sense there, but destroy and reinit webview didn't seem like "presenter's" responsibility. And `WebViewClient` also seems inappropriate for the inverse reasons.

Then tried doing this in the observers that trigger re-init -- but felt like it should be a single call from those observers, the observer shouldn't each need to "know" to chain all those things together. 

Then I remembered the `unregisterInAppForms` ticket was coming up, the work is kinda overlapping, so I came up with this approach of reinitializing being the equivalent of un registering and re-registering.

sdk/forms/src/main/java/com/klaviyo/forms/InAppForms.kt
```diff
@@ -41,3 +44,28 @@ fun Klaviyo.registerForInAppForms(
...
+ * hiding any currently displayed forms and preventing any further forms from being presented.
+ */
+@UiThread
+fun Klaviyo.unregisterInAppForms() = safeApply {
+    Registry.apply {
+        unregister<InAppFormsConfig>()
+        getOrNull<PresentationManager>()?.dismiss()
+        getOrNull<WebViewClient>()?.destroyWebView()
+    }
+}
```
evan-masseau, line 59: Unregister refers to the in-app observers and triggering logic, doesn't mean we need to unregister the underlying services. Just stop them using up resources -- finish activity if running, destroy the webview, stop the observers, and it also makes sense to discard the registered config.

sdk/core/src/main/java/com/klaviyo/core/utils/KlaviyoThreadHelper.kt
```diff
@@ -0,0 +1,26 @@
...
+import android.os.HandlerThread
+import android.os.Looper
+
+/**
+ * Abstraction of our interactions with handlers/threads for isolation purposes
+ * @deprecated Use [com.klaviyo.core.Registry.threadHelper] instead
+ */
+object KlaviyoThreadHelper : ThreadHelper {
+    override fun getHandler(looper: Looper) = Handler(looper)
+    override fun getHandlerThread(name: String?) = HandlerThread(name)
```
dan-peluso, line 13: There's no reason we'd want to store a specific Handler or Handler thread in some sort of registry / memory right? Just want to be sure that making a new one every time we call this is the expected behavior

sdk/core/src/main/java/com/klaviyo/core/utils/KlaviyoThreadHelper.kt
```diff
@@ -0,0 +1,26 @@
...
+    override fun runOnUiThread(block: () -> Unit) {
+        val mainLooper = Looper.getMainLooper()
+
+        if (mainLooper == Looper.myLooper()) {
+            // Already on main thread, run immediately
+            block()
```
dan-peluso, line 19: on first glance I thought this was trying to act like a thread block, but I guess it's supposed to be interpreted as 'code block' right? Any shot we could rename this? Certainly a nit, but block as a name was just confusing me.

sdk/core/src/main/java/com/klaviyo/core/utils/KlaviyoThreadHelper.kt
evan-masseau: Hm yeah good question, I didn't change this code, just moved it, but going from the comments this appears to have been the intended behavior 

```
     * This method is synchronized to avoid potentially starting the same thread twice.
     * Since we only ever have one thread running for our network requests, this is fine but if we ever extrapolate on this, we may want to revisit this logic
     * e.g: Synchronizing on the object instance (this) because I don't think we need to synchronize on anything else in this object. We may want to use a proper lock if we need more synchronized blocks or utilize more threading
```

sdk/forms/src/main/java/com/klaviyo/forms/InAppForms.kt
```diff
@@ -41,3 +44,28 @@ fun Klaviyo.registerForInAppForms(
...
+
+/**
+ * Resets the in-app forms listeners with the current configuration.
+ */
+@UiThread
+internal fun Klaviyo.reInitializeInAppForms() = safeApply {
+    // If config is missing, initial registration has not occurred so we can't re-initialize
+    Registry.getOrNull<InAppFormsConfig>()?.let { config ->
```
dan-peluso, line 67: Same thing here, maybe a debug log if there was no config?

sdk/forms/src/main/java/com/klaviyo/forms/bridge/JsBridgeObserverCollection.kt
```diff
@@ -12,7 +12,7 @@ internal interface JsBridgeObserverCollection {
     /**
      * Compiles the handshake data from all observers into a single list
      */
-    val handshake: List<HandshakeSpec> get() = observers.map { it.handshake }
+    val handshake: List<HandshakeSpec> get() = observers.mapNotNull { it.handshake }
```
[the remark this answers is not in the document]
evan-masseau: had to check, but ya the "not null" applies to the result of the transformer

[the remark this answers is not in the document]
dan-peluso: Oh nice

dan-peluso (approved):

PR #273, opened 2025-06-02
IAF | Close form in the webview when back button is pressed

evan-masseau:
# Description

Rather than just dismissing the activity, if a form is currently presented, this tells `klaviyo.js` to close the form within the webview. That will result in a `formDisappeared` message which in turn dismisses the activity. Also added a fallback to dismiss the activity if `formDisappeared` is not received within the expected timeframe.

## Due Diligence

## Changelog / Code Overview

- Added `closeForm` JS method and associated bridge function. 
- Moved all the handshake items for SDK -> Native events into `JsBridge`, in hindsight makes much more sense there!
- Added `formVersionId` alongside `formId` -- that'll be the preferred form identifier for closing a form.

## Test Plan

`closeForm` is not yet supported on Fender, but this is still partially testable because the activity should still get dismissed as a fallback. 

## Related Issues/Tickets
 
 https://klaviyo.atlassian.net/browse/CHNL-21562


sdk/forms/src/main/assets/klaviyo-js-bridge.js
```diff
@@ -9,7 +9,7 @@
  * @param anonymous_id
  * @returns {boolean}
  */
-window.setProfile = function (external_id, email, phone_number, anonymous_id) {
+window.profileMutation = function (external_id, email, phone_number, anonymous_id) {
```
evan-masseau, line 12: Consolidated these method names with their associated handshake name.

sdk/forms/src/main/java/com/klaviyo/forms/bridge/CompanyObserver.kt
```diff
@@ -11,11 +11,6 @@ import com.klaviyo.forms.reInitializeInAppForms
  * to reload klaviyo.js for the new company
  */
 internal class CompanyObserver : JsBridgeObserver {
-    /**
-     * At this time, company ID doesn't have a handshake spec because
-     * it only resets the webview, doesn't communicate with the onsite module
-     */
-    override val handshake: HandshakeSpec? = null
```
evan-masseau, line 18: Realized that these handshake items make more sense to live in the `JsBridge` interface, similar to how native bridge has a handshake collection, than on the individual observers. Not all observers even have a handshake required.

sdk/forms/src/main/java/com/klaviyo/forms/bridge/JsBridge.kt
```diff
@@ -2,24 +2,46 @@ package com.klaviyo.forms.bridge
...
     /**
      * Dispatch lifecycle events for onsite JS package to consume
      */
-    fun dispatchLifecycleEvent(type: LifecycleEventType)
+    fun lifecycleEvent(type: LifecycleEventType)
+
+    /**
+     * Open a form in the webview by its ID
+     */
+    fun openForm(formId: FormId)
```
evan-masseau, line 40: This is for later, I have an idea that would allow for manually triggering a form by ID from the test app.

sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoJsBridge.kt
```diff
@@ -11,20 +11,48 @@ import com.klaviyo.forms.webview.JavaScriptEvaluator
...
+        ),
+        HandshakeSpec(
+            type = HelperFunction.lifecycleEvent.name,
+            version = 1
+        ),
+        HandshakeSpec(
+            type = HelperFunction.closeForm.name,
+            version = 1
+        )
+    )
```
evan-masseau, line 33: Moved handshake collection here for Native -> JS messages, makes more sense, similar to NativeBridge: https://github.com/klaviyo/klaviyo-android-sdk/blob/5873d71310c00a5a9b429d7cda58fedcc3c8127b/sdk/forms/src/main/java/com/klaviyo/forms/bridge/NativeBridge.kt#L25

sdk/forms/src/main/java/com/klaviyo/forms/bridge/NativeBridgeMessage.kt
```diff
@@ -137,7 +140,8 @@ internal sealed class NativeBridgeMessage {
                 )
 
                 keyName<FormDisappeared>() -> FormDisappeared(
-                    formId = jsonData.optString("formId").ifEmpty { "" }
+                    formId = jsonData.optString("formId"),
+                    formVersionId = jsonData.optInt("formVersionId")
```
evan-masseau, line 144: todo should be same as above

sdk/forms/src/main/assets/klaviyo-js-bridge.js
```diff
@@ -66,6 +66,40 @@ window.dispatchAnalyticsEvent = function (metric, strProperties) {
...
+/**
+ * Force open a Klaviyo form with the given formId.
+ *
+ * Note: This is just standard klaviyo.js functionality, not an onsite-in-app-forms listener.
+ *
+ * @param formId
+ * @returns {boolean}
+ */
+window.openForm = function (formId) {
+    window._klOnsite = window._klOnsite || [];
```
dan-peluso, line 78: Could you explain what this line does? I'm worried that this could be null and throw on the line below

sdk/forms/src/main/java/com/klaviyo/forms/bridge/JsBridge.kt
dan-peluso: Maybe worth elaborating that in the comment above? I feel like if I was a new dev to this I might think this is a relevant function for standard fucntionality

sdk/forms/src/main/assets/klaviyo-js-bridge.js
evan-masseau: Yea, This line actually ensures that `_klOnsite` is set to _something_ so that the next line won't throw.

Its a neat little trick onsite uses for queueing up requests prior to klaviyo.js loading. If klaviyo.js has loaded already, this line is a no-op. If it hasn't, this line initializes `window.klOnsite` to an plain old array. When klaviyo.js does load, it'll grab that array and run any operations that had been enqueued, then replace `_klOnsite` with a proxy, so if you call `push(something)` from then on, those command get run immediately.

sdk/forms/src/main/assets/klaviyo-js-bridge.js
```diff
@@ -66,6 +66,40 @@ window.dispatchAnalyticsEvent = function (metric, strProperties) {
...
+/**
+ * Close a Klaviyo form by formId, or any currently open forms
+ *
+ * @param {string} formId
+ * @returns {boolean}
+ */
+window.closeForm = function (formId) {
+    document.head.dispatchEvent(
+        new CustomEvent(
+            'closeForm',
```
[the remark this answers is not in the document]
evan-masseau: sadly no

dan-peluso (approved):
Nothing major, glad we got timeouts playing nice for the unit tests

PR #281, opened 2025-06-17
IAF | Two timing related bugs

evan-masseau:
# Description

Squashed two bugs with one stone. Added this `runWithCurrentOrNextActivity` utility for operations that are sensitive to the current activity. 

## Due Diligence

## Changelog / Code Overview

Fixes two bugs both related to being sensitive to activity transitions: 
- Deep links: fixed an unreliable timing issue where current activity can be null when deep link is received during the brief transition from our overlay back to the host app's activity. 
- Time delayed forms: if a time delayed form is evaluated to display after the app has been backgrounded, the webview could force the app to reopen when it presented our form overlay activity. Fixed by postponing presentation till the app is next resumed

## Test Plan

- Deep links: 
  - Open a form with a deep link and tap the deep linked button.
  - On the play store build `3.4.0-alpha.1` we saw intermittent issue where the form would close without navigating to the deep link. (I was able to isolate the issue because of a debug log, when the deep link didn't work, we were logging "Current activity is null)")
  - With this fix, deep link should work 100% of the time. 
- Form while app is backgrounded: 
  - Configure a form to display on a 10 second delay
  - Open the app to trigger the form, but background the app before it appears
  - On prior to this branch: the app would re-open to display the form!
  - After this fix, the form will only appear once you open the app (before your the inactivity timeout)
  - Additional check: If you wait for your form session to timeout, the form should not appear on next launch (it should wait the 15s timer as expected)

## Related Issues/Tickets
 
 https://klaviyo.atlassian.net/browse/CHNL-22081
https://klaviyo.atlassian.net/browse/CHNL-22082

### DIFF
--- a/sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
@@ -11,7 +11,7 @@ typealias ActivityObserver = (activity: ActivityEvent) -> Unit
 /**
  * Represent different events emitted in response to lifecycle triggers from the host application
  */
-sealed class ActivityEvent(val activity: Activity? = null, val bundle: Bundle? = null) {
+sealed class ActivityEvent(open val activity: Activity? = null, val bundle: Bundle? = null) {
 
     /**
      * Get the type of the event as a string (e.g. for logging)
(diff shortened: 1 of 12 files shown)


sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
```diff
@@ -11,7 +11,7 @@ typealias ActivityObserver = (activity: ActivityEvent) -> Unit
 /**
  * Represent different events emitted in response to lifecycle triggers from the host application
  */
-sealed class ActivityEvent(val activity: Activity? = null, val bundle: Bundle? = null) {
+sealed class ActivityEvent(open val activity: Activity? = null, val bundle: Bundle? = null) {
```
evan-masseau, line 14: Allows me to specify the non-nullable activity in some child classes

sdk/core/src/test/java/com/klaviyo/core/config/KlaviyoConfigTest.kt
```diff
@@ -40,8 +40,6 @@ internal class KlaviyoConfigTest : BaseTest() {
         super.setup()
         mockkStatic(PackageManager.PackageInfoFlags::class)
         every { PackageManager.PackageInfoFlags.of(any()) } returns mockPackageManagerFlags
-        every { mockContext.packageManager } returns mockPackageManager
-        every { mockContext.packageName } returns BuildConfig.LIBRARY_PACKAGE_NAME
```
evan-masseau, line 44: Duplicated by `BaseTest`

ajaysubra (approved):

sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
dan-peluso: Love this

evan-masseau: Amusingly, copilot told it couldn't be done and I was like... hmm i'm gonna try that sounds wrong

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoPresentationManager.kt
```diff
@@ -38,19 +52,17 @@ internal class KlaviyoPresentationManager() : PresentationManager {
...
-                presentationState.takeIf<Presenting>()?.let {
-                    overlayActivity = activity
-                    Registry.get<WebViewClient>().attachWebView(activity)
-                    presentationState = Presented(it.formId)
-                    Registry.log.debug("Presentation State: $presentationState")
-                }
+        is ActivityEvent.Created -> event.activity.takeIf<KlaviyoFormsOverlayActivity>()?.let { activity ->
+            presentationState.takeIf<Presenting>()?.let {
+                overlayActivity = activity
+                Registry.get<WebViewClient>().attachWebView(activity)
```
dan-peluso, line 58: how do we handle this registry.get if the webview client is not added yet? Ik unlikely, but wondering if we need to ?.let this or maybe try catch if android decides to android on us

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoPresentationManager.kt
```diff
@@ -104,3 +138,37 @@ internal class KlaviyoPresentationManager() : PresentationManager {
...
+        event.takeIf<ActivityEvent.Resumed>()?.let { event ->
+            Registry.log.verbose("Invoking postponed observer on resume")
+            job(event.activity)
+            observer?.let { offActivityEvent(it) }
+            token?.cancel()
+        }
+    }
+    onActivityEvent(observer)
+
+    return token
```
dan-peluso, line 173: why are we calling this token?

PR #283, opened 2025-06-17
CI | Run tests in parallel

evan-masseau:
Thought I'd try running our tests in parallel to speed up CI a little.

Looks like a minimum of 1.5-2 minutes is just compiling code, which all the parallel tests will have to do anyways. And then 1-2 minutes on the longer test suites. So maybe shaves a minute off. 
Serially:
Screenshot 2025-06-17 at 12 11 37
Screenshot 2025-06-17 at 12 11 54

Parallel:
Screenshot 2025-06-17 at 12 10 51

I tried front-loading the build/assemble task but it slowed things down because the build jobs still have to run twice, and that is net longer even though the second run does skip a few steps.

### DIFF
--- a/.github/workflows/android-feature.yml
+++ b/.github/workflows/android-feature.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        module: [ :sdk:core, :sdk:analytics, :sdk:forms, :sdk:push-fcm ]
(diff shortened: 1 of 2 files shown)


dan-peluso:
o hell yeah

amber-klaviyo (approved):

PR #284, opened 2025-06-17
IAF | Switch the session config from long to duration

evan-masseau:
# Description

Use the kotlin Duration type instead of Long, partly for the readability, partly for parity w/iOS, and partly for the INFINITE const.

## Due Diligence

## Changelog / Code Overview

Switched the timeout config type from Long to Duration. 

## Test Plan

Update the test app to use the same type, and test that session timeout still works as expected.

## Related Issues/Tickets
 
 Just an idea from reviewing the equivalent iOS code https://github.com/klaviyo/klaviyo-swift-sdk/pull/366/files#r2152909506

### DIFF
--- a/sdk/fixtures/src/main/java/com/klaviyo/fixtures/StaticClock.kt
+++ b/sdk/fixtures/src/main/java/com/klaviyo/fixtures/StaticClock.kt
@@ -1,6 +1,7 @@
 package com.klaviyo.fixtures
 
 import com.klaviyo.core.config.Clock
+import kotlin.time.Duration
 
 /**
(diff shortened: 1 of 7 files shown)


MIGRATION_GUIDE.md
```diff
@@ -2,6 +2,15 @@
 This document provides guidance on how to migrate from the old version of the SDK to a newer version. 
 It will be updated as new versions are released including deprecations or breaking changes.
 
+# 3.4.0
```
dan-peluso, line 5: Should we make the base of this the RC branch?

PR #285, opened 2025-06-23
[WIP] README updates and other pre-release fixes.

evan-masseau:
# Description

## Due Diligence

### DIFF
--- a/sdk/forms/consumer-rules.pro
+++ b/sdk/forms/consumer-rules.pro
@@ -1,0 +1,1 @@
+-keepnames class com.klaviyo.forms.bridge.** { *; }
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -5,8 +5,19 @@ It will be updated as new versions are released including deprecations or breaki
 # 3.4.0
 
 ### Improvements
-- A deep link from an in-app form will now be issued *after* the form has closed, instead of during the close animation.
 
+#### In-App Forms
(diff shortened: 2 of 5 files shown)


sdk/forms/consumer-rules.pro
```diff
@@ -0,0 +1 @@
+-keepnames class com.klaviyo.forms.bridge.** { *; }
```
dan-peluso, line 1: Is there any chance we'd want to add these (similar to push-fcm module)
```suggestion
-keep class com.klaviyo.core.** { *; }
-keep class com.klaviyo.analytics.** { *; }
-keepnames class com.klaviyo.forms.bridge.** { *; }
```

evan-masseau: I don't think so... in the analytics package we added `analytics.**` to make sure that the minifier never strips out the `Keyword` classes like profile keys and event keys, because they may appear unused but can be referenced dynamically (like we do with reflection in the react SDK). I'm not 100% we actually needed `core.**` but probably just chucked it in?

Then again, it probably doesn't _hurt_ either. Might be interesting to try a test app that only adds the `forms` package dependency and make sure that no critical core/analytics components get minified/obfuscated out...

evan-masseau: Nice, thanks! So we'll just stick to `-keepnames class com.klaviyo.forms.bridge.** { *; }` and not add the other two?

PR #294, opened 2025-07-30
Add base url as a field on the API request

evan-masseau:
# Description
This adds base url as a field on the API request that is persisted / restored from disk for queued requests. 
It will be necessary for universal click tracking links because we need to send those API requests to arbitrary URLs, not just the configured API base url. 
Incidentally, this also fixes an old, trivial bug where switching base URLs in the config can cause queued requests to wind up sending to the wrong server.

## Due Diligence

- [ ] I have adjusted or added new test cases to team test docs, if applicable. -- Not really applicable

## Changelog / Code Overview

Adds base url as a field on the API request so it can be overridden. 
De/serializes that field to disk when persisting (deserializing falls back on configured base url for migration purposes)

## Test Plan

Use the test app to send some basic event/push/profile event requests.

## Related Issues/Tickets
 
 Relates to universal linking but not explicitly ticketed.

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
@@ -136,9 +136,11 @@ internal object KlaviyoApiClient : ApiClient {
                     "${request.type} Request failed with code ${request.responseCode}, and will be retried up to $attemptsRemaining more times."
                 )
             }
+
             Status.Complete -> Registry.log.verbose(
                 "${request.type} Request succeeded with code ${request.responseCode}"
             )
+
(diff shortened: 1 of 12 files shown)


ab1470 (approved):
LGTM

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -376,4 +381,8 @@ internal object KlaviyoApiClient : ApiClient {
             handler?.postDelayed(this, flushInterval)
         }
     }
+
+    private fun KlaviyoApiRequest.sendAndBroadcast(): KlaviyoApiRequest.Status = send {
```
dan-peluso, line 385: nice extension!

PR #298, opened 2025-08-04
IAF | Inject safe area inset values to work around buggy webview implementation

evan-masseau:
# Description
Android webview has a buggy implementation of safe area insets. Support was only even introduced in May, and they have an extant issue where loading the content before attaching to the view hierarchy causes all safe area insets to compute as 0, with no way to re-compute other than reloading the page 🙄 

With this work around we will compute the safe area ourselves and inject them as css variables. In our JS/CSS package we will check for this variable first, and fall back on the env value if these values are not present.

## Due Diligence

## Changelog / Code Overview

Compute safe area on the native side
Inject that as a CSS var

## Test Plan

- [ ] Asset source the corresponding fender PR and test that a full screen form properly pads the form content (namely the X icon)

## Related Issues/Tickets
 
 https://klaviyo.atlassian.net/browse/CHNL-23925

### DIFF
--- a/sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebView.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebView.kt
@@ -6,11 +6,17 @@ import android.graphics.Color
 import android.util.AttributeSet
 import android.webkit.WebView
 import android.webkit.WebViewClient
+import androidx.core.util.TypedValueCompat.pxToDp
+import androidx.core.view.ViewCompat
+import androidx.core.view.WindowInsetsCompat.Type.displayCutout
+import androidx.core.view.WindowInsetsCompat.Type.ime
+import androidx.core.view.WindowInsetsCompat.Type.systemBars
 import androidx.webkit.WebViewCompat
 import androidx.webkit.WebViewFeature.WEB_MESSAGE_LISTENER
 import androidx.webkit.WebViewFeature.isFeatureSupported
 import com.klaviyo.core.DeviceProperties
 import com.klaviyo.core.Registry
+import com.klaviyo.forms.bridge.JsBridge
 import com.klaviyo.forms.bridge.NativeBridge
 
 /**
@@ -28,6 +34,7 @@ internal class KlaviyoWebView : WebView {
     fun loadTemplate(html: String, client: WebViewClient, bridge: NativeBridge) = configure()
         .apply { webViewClient = client }
         .addBridge(bridge)
+        .monitorSafeArea()
         .loadDataWithBaseURL(
             bridge.allowedOrigin.first(),
             html,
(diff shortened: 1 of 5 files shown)


sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebView.kt
```diff
@@ -66,4 +73,28 @@ internal class KlaviyoWebView : WebView {
...
+            )
+            Registry.get<JsBridge>().setSafeArea(
+                // Convert raw pixels to density independent pixels
+                pxToDp(safeDrawingInsets.left.toFloat(), displayMetrics),
+                pxToDp(safeDrawingInsets.top.toFloat(), displayMetrics),
+                pxToDp(safeDrawingInsets.right.toFloat(), displayMetrics),
+                pxToDp(safeDrawingInsets.bottom.toFloat(), displayMetrics)
+            )
+
+            windowInsets
```
dan-peluso, line 97: I looked a bit into the effect of returning the same listener as the lambda provides - my impression is that this gives us a 'one shot' window inset observation but that if there was a special case (like maybe a pixel fold has a different notch setup when unfolded versus not) this might not apply. I was thinking though, and you'd be the person to ask on this, maybe we'd be covered if the template load gets reapplied on these sort of configuration changes? Super edge case-y but curious if you think this is a concern

evan-masseau: yeah good point, this could be in initializer or something instead

evan-masseau: oh and yeah i could return  consumed, i just didn't really care since the webview doesn't have subviews that would be propagating down to. maybe would be better just in case.
```
  // Return CONSUMED if you don't want the window insets to keep passing
  // down to descendant views.
  WindowInsetsCompat.CONSUMED
```

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebView.kt
```diff
@@ -66,4 +73,28 @@ internal class KlaviyoWebView : WebView {
             addJavascriptInterface(bridge, bridge.name)
         }
     }
+
+    private fun monitorSafeArea() = apply {
+        ViewCompat.setOnApplyWindowInsetsListener(this) { myWebView, windowInsets ->
+            // Retrieve insets as raw pixels
+            val displayMetrics = myWebView.context.resources.displayMetrics
+            val safeDrawingInsets = windowInsets.getInsets(
+                systemBars() or displayCutout() or ime()
```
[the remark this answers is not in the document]
evan-masseau: What are you referring to exactly, because this is how you dynamically request the insets from the device.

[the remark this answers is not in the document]
evan-masseau: If there comes some new category of display with weird new bounds that system bars and cutouts don't account for, then we would need new code to account for it. There is no more dynamic way to get safe areas that I can see than this.

dan-peluso (approved):

Issue #295, opened 2025-07-31
Receiving java.lang.SecurityException from KlaviyoNetworkMonitor.onNetworkChange()


dimakpoq:
### Checklist

- [x] I have read the [contributing guidelines](https://github.com/klaviyo/klaviyo-android-sdk/blob/master/.github/CONTRIBUTING.md)
- [ ] I have determined whether this bug is also reproducible in a vanilla Android project
- [ ] If possible, I've reproduced the issue using the `master` branch or latest release of this package.
- [x] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/klaviyo/klaviyo-android-sdk/issues) or [pull request](https://github.com/klaviyo/klaviyo-android-sdk/pulls).

### Description
Hello Klaviyo team,

We are receiving the following crash from Firebase Crashlytics, it happens on Android 13 (99% of cases) and on Android 11 (1% of the cases):
```markdown
          Caused by java.lang.SecurityException: <apppackagename> was not granted either of these permissions:android.permission.CHANGE_NETWORK_STATE,android.permission.WRITE_SETTINGS.
       at android.os.Parcel.createExceptionOrNull(Parcel.java:3011)
       at android.os.Parcel.createException(Parcel.java:2995)
       at android.os.Parcel.readException(Parcel.java:2978)
       at android.os.Parcel.readException(Parcel.java:2920)
       at android.net.IConnectivityManager$Stub$Proxy.requestNetwork(IConnectivityManager.java:2506)
       at android.net.ConnectivityManager.sendRequestForNetwork(ConnectivityManager.java:4767)
       at android.net.ConnectivityManager.sendRequestForNetwork(ConnectivityManager.java:4904)
       at android.net.ConnectivityManager.requestNetwork(ConnectivityManager.java:5042)
       at android.net.ConnectivityManager.requestNetwork(ConnectivityManager.java:5020)
       at com.klaviyo.core.networking.KlaviyoNetworkMonitor.initializeNetworkListener(KlaviyoNetworkMonitor.kt:120)
       at com.klaviyo.core.networking.KlaviyoNetworkMonitor.onNetworkChange(KlaviyoNetworkMonitor.kt:54)
       at com.klaviyo.analytics.networking.KlaviyoApiClient.startService(KlaviyoApiClient.kt:52)
       at com.klaviyo.analytics.Klaviyo$initialize$1.invoke(Klaviyo.kt:82)
[...]

PR #306, opened 2025-08-19
CHNL-23800 Catch network security exception

dan-peluso:
# Description
Based on this [issue](https://github.com/klaviyo/klaviyo-android-sdk/issues/295

Certainly a deep edge case, considering we assert pretty early on that we do have the correct permissions. Still, I tested out throwing an error at the same place specified in the issue threads and noticed that event without the monitor being attached we can continue to retry requests without the monitor being set up. The only difference is that we don't retry the network batch based on the connection becoming available, we do it based on the retry logic in the requests themselves. 

## Due Diligence

## Test Plan
Tested on a Pixel 4XL running Android 33. Was not able to repro the bug but I'm confident in the fix.

## Related Issues/Tickets
https://github.com/klaviyo/klaviyo-android-sdk/issues/295


evan-masseau (approved):
Great, thanks for investigating, fix looks good

PR #312, opened 2025-09-04
Geofence feature scaffold

evan-masseau:
# Description

Committing the early scaffolding and PoC work on geofencing, lots of TODOs left but want to start iterating through pull requests and reviews. This targets `feat/geofencing` to establish that as the target feature branch for all geofencing work.


sdk/location/src/main/java/com/klaviyo/location/KlaviyoLocationManager.kt
```diff
@@ -0,0 +1,154 @@
...
+    override fun getCurrentGeofences(): List<KlaviyoGeofence> {
+        // TODO return geofences currently monitored
+        return emptyList()
+    }
+
+    fun handleGeofenceIntent(context: Context?, intent: Intent?) {
+        val geofencingEvent = intent?.let { GeofencingEvent.fromIntent(it) } ?: return
+
+        if (geofencingEvent.hasError()) {
+            val errorMessage = GeofenceStatusCodes.getStatusCodeString(geofencingEvent.errorCode)
```
belleklaviyo, line 135: oh thats cool, I didn't know android had their own geofence status codes, specifically the "too many geofences" one. I tested on iOS and even though all the documentation/forums say they only allow max 20, it doesn't actually show any error out of the box if you add too many. guess it just silently removes some over others. might be worth adding that level of validation on iOS side if android has this

belleklaviyo (approved):

.github/workflows/android-feature.yml
```diff
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        module: [ ':sdk:core', ':sdk:analytics', ':sdk:forms', ':sdk:push-fcm' ]
+        module: [ ':sdk:core', ':sdk:analytics', ':sdk:forms', ':sdk:location', ':sdk:push-fcm' ]
```
evan-masseau, line 19: Tests from this module now running in CI
Screenshot 2025-09-09 at 13 31 33

PR #313, opened 2025-09-05
Push DeepLinkHandler Integration

evan-masseau:
# Description
Push notifications with a deep link will now invoke the deep link handler, if registered. If not, unlike IAF or universal tracking links, we won't fall back to broadcasting an intent because the host app already received the intent we would send.
Incidentally, this is a precursor to having a middleware activity to handle push opens and universal links automatically.

## Due Diligence

## Changelog / Code Overview

- `handlePush` will now invoke the developer's deep link handler, if registered and push contained a deep link.
- Added some test utilities
- Took out an unused extension property

## Test Plan

- [ ] Thoroughly test deep linking from all origins (push, form, link) in all supported app states (terminated, backgrounded, foregrounded), with handler or without.

## Related Issues/Tickets
 
[Ticket](https://klaviyo.atlassian.net/browse/CHNL-25178)
[Testing](https://klaviyo.atlassian.net/browse/CHNL-25085)


sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
```diff
@@ -195,28 +194,25 @@ class KlaviyoNotification(private val message: RemoteMessage) {
...
         )
+
+    /**
+     * Create the appropriate intent to send with the notification is tapped
+     * When auto-track is enabled, use our middleware activity to handle the open
+     * Otherwise, use the deep link if available, or fall back to launching the app
+     */
+    private fun makeOpenedIntent(context: Context) = message.deepLink.let { deepLink ->
+        when {
+            // Else if deep link is present, use an ACTION_VIEW intent
```
evan-masseau, line 212:
```suggestion
            // If deep link is present, use an ACTION_VIEW intent
```

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -286,27 +286,36 @@ object Klaviyo {
...
      */
-    fun handlePush(intent: Intent?) = safeApply(preInitQueue) {
-        if (intent?.isKlaviyoNotificationIntent != true) {
-            Registry.log.verbose("Non-Klaviyo intent ignored")
-            return@safeApply
-        }
-
-        val event = Event(EventMetric.OPENED_PUSH)
-        val extras = intent.extras
+    fun handlePush(intent: Intent?): Klaviyo = this
```
evan-masseau, line 289: This is the primary purpose of the PR: if the developer has registered a deep link handler, we should invoke it when they call `handlePush`. 
I made it independent of the  event enqueuing (so that a failure there doesn't affecting linking) but still filtered to pushes originating from klaviyo only.

sdk/analytics/src/test/java/com/klaviyo/analytics/linking/DeepLinkingTest.kt
```diff
@@ -0,0 +1,170 @@
...
+import io.mockk.verify
+import org.junit.After
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+
+internal class DeepLinkingTest : BaseTest() {
```
evan-masseau, line 24: similar to above, renamed file and changed enough that it became a delete/add in the diff.

sdk/fixtures/src/main/java/com/klaviyo/fixtures/MockIntent.kt
```diff
@@ -0,0 +1,87 @@
...
+import io.mockk.CapturingSlot
+import io.mockk.every
+import io.mockk.mockk
+import io.mockk.mockkConstructor
+import io.mockk.slot
+
+/**
+ * Data class to hold all Intent capturing slots for easy access and destructuring
+ */
+data class MockIntent(
```
evan-masseau, line 16: I extracted this test utility for mocking Intent across all tests. Makes it easier to get tests involving intents off the ground, particularly if you need to verify how intent properties are set/mutated.

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
```diff
@@ -58,7 +58,6 @@ class KlaviyoNotification(private val message: RemoteMessage) {
         internal const val BODY_KEY = "body"
         internal const val URL_KEY = "url"
         internal const val IMAGE_KEY = "image_url"
-        internal const val CLICK_ACTION_KEY = "click_action"
```
evan-masseau, line 61: I added this key in the very first version in the event we supported other types of intent actions. I don't think its something we'd ever need to use after all.

sdk/analytics/src/main/java/com/klaviyo/analytics/linking/DeepLinking.kt
```diff
@@ -0,0 +1,96 @@
...
+        sendDeepLinkIntent(uri)
+    }
+
+    /**
+     * Sends an intent to launch the host application.
+     *
+     * @param context The context used to access the package manager and start the activity
+     * @param extras Optional bundle of extras to be added to the launch intent
+     */
+    fun sendLaunchIntent(context: Context, extras: Bundle? = null) =
```
dan-peluso, line 44: when do we expect to use this in the final flow?

dan-peluso: is this where the universal link initially gets triggered from?

dan-peluso, line 44: Also, and I know this is a nit / something we've respectfully disagreed on in the past (no shame on your stance of things) but I do generally feel like we should only be using the '=' instead of brackets on function calls when we are explicitly returning something from the function. This call technically has the signature of returning `Unit?` which I think is a bit funky.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/UniversalClickTrackRequest.kt
```diff
@@ -31,13 +31,24 @@ internal class UniversalClickTrackRequest(
...
+    /**
+     * Use a short timeout for the initial request with callback, since we don't want to
+     * keep the user waiting too long. If it fails, we enqueue the request to be retried later.
+     * For retries, use the normal timeout duration.
+     */
+    override val timeoutDuration: Int
+        get() = if (headers.containsKey(KLAVIYO_CLICK_TIMESTAMP_HEADER)) {
+            super.timeoutDuration
+        } else {
+            2_000
```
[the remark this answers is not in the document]
evan-masseau: True, I could move that into config instead. Milliseconds.

sdk/analytics/src/main/java/com/klaviyo/analytics/linking/DeepLinking.kt
evan-masseau: Yeah trying to break myself of this pattern, its deeply ingrained

dan-peluso (approved):
mostly small questions

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -286,27 +286,36 @@ object Klaviyo {
...
+            }
 
-        extras?.keySet()?.forEach { key ->
-            if (key.contains("com.klaviyo")) {
-                val eventKey = EventKey.CUSTOM(key.replace("com.klaviyo.", ""))
-                event[eventKey] = extras.getString(key, "")
+            Registry.get<State>().pushToken?.let { event[EventKey.PUSH_TOKEN] = it }
+
+            // Not using createEvent here to avoid nested safeApply calls
+            Registry.get<ApiClient>().enqueueEvent(event, Registry.get<State>().getAsProfile())
```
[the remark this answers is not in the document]
evan-masseau: defs shouldn't, i'll double check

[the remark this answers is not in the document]
evan-masseau: I didn't change this code, just added the comment because I was briefly tempted to switch this to `Klaviyo.createEvent()` before remembering why we didn't do that before

sdk/analytics/src/main/java/com/klaviyo/analytics/linking/DeepLinking.kt
evan-masseau: I like the conciseness of an expression, but agree it carries an implication of a return value. Returning `Unit?` is also odd.
I looked at adding a ktlint rule to standardize this a bit better but it didn't seem to have a good way to configure that rule.

PR #318, opened 2025-09-12
Support limited pre-initialize operations by registering context only

evan-masseau:
# Description

**Problem**: some companies use the same app with multiple public API keys. They are not able to provide company ID on initial launch because it must be determined via some app logic first. However, our `initialize` method and internal config setup tightly couples the collection of `Context` to the API key. 

**Goal**: decouple Context and API Key such that we can enable more operations that don't require API key, including universal link handling. At the same time, we still need to fail gracefully in the same manner for an pre-initialize operations that do require company ID. We do  buffer some operations in memory (push-related operations that are constrained by system API timing constraints)

## Due Diligence

## Changelog / Code Overview

I'm updating `registerForLifecycleCallbacks`, since this is already intended to be our required method for devs who can't initialize with API key immediately. It now captures the provided context in the config and registers API and State services which can partially function without an API key. 
To decouple API key and Context, I moved the validation for API key from the config builder, to the API key accessor. That way the same exception type is thrown, handled the same way by `safeCall` / `safeApply`, but more targeted. 

## Test Plan

1. All the existing tests should pass with minimal modification (did have to fix some incorrect mocks). We already had substantial unit testing of pre-init state. 
2. I've done some manual tests with the internal test app, where I just called sdk methods operations before `initialize` to verify they don't crash the app, as expected. Did this with all the public setters, getters, creating an event, handling push, registering for forms, etc. 
4. I [updated the internal test app](https://github.com/klaviyo/klaviyo-android-test-app/pull/103/files) to wait to call initialize till the user hits the "start" button. So that we can consistently test the "late initialize" flow used by some of our customers. Using that new "late initialize" mode, I tested the following (none should crash the app, but they'll fail in different ways)
- [x] Call `registerForLifecycleListeners` to provide partial context to the SDK
- [x] Profile setters/getters -- Must be called after `initialize` (as before)
- [x] `registerForInAppForms` -- Must be called after `initialize` (as before)
- [x] `createEvent` -- Must be called after `initialize` (as before)
- [x] `handlePush` -- API call fails, but retries when API key is provided (as before)
- [x] `registerDeepLinkHander` -- Must call `initialize` or `registerForLifecycleListeners`

## Related Issues/Tickets
 
 https://klaviyo.atlassian.net/browse/CHNL-25429

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
@@ -41,20 +41,6 @@ object Klaviyo {
     private val preInitQueue: Queue<Operation<Unit>> = LinkedList()
 
     /**
-     * Since the analytics module owns these services, it must register them.
-     *
-     * This registration is a lambda invoked when the service is required, not instantiated now
-     */
-    private fun registerServices() = Registry.apply {
-        registerOnce<ApiClient> { KlaviyoApiClient }
-        registerOnce<State> {
-            KlaviyoState().also { state ->
-                register<StateSideEffects>(StateSideEffects(state))
-            }
-        }
-    }
-
-    /**
      * This method is provided for apps that are unable to register their API key immediately
      * on app launch in order enable limited SDK functionality including tracking app lifecycle,
      * automated push token collection, and handling universal tracking links.
@@ -64,8 +50,6 @@ object Klaviyo {
      * @param applicationContext
      */
     fun registerForLifecycleCallbacks(applicationContext: Context) = safeApply {
-        registerServices()
-
         if (!Registry.isRegistered<Config>()) {
             // Register a partial config, missing API Key, to allow lifecycle tracking and context access for partial functionality
             Registry.register<Config>(
@@ -75,6 +59,10 @@ object Klaviyo {
             )
         }
 
+        // Some APIs (such as deep linking) work without an API key, so we can register the core service now
+        Registry.registerOnce<ApiClient> { KlaviyoApiClient }
+
+        // Register lifecycle callbacks to monitor app foreground/background state
         Registry.config.applicationContext.applicationContext.takeIf<Application>()?.apply {
             unregisterActivityLifecycleCallbacks(Registry.lifecycleCallbacks)
             unregisterComponentCallbacks(Registry.componentCallbacks)
@@ -91,8 +79,6 @@ object Klaviyo {
      * @param applicationContext
      */
     fun initialize(apiKey: String, applicationContext: Context) = safeApply {
-        registerServices()
-
         Registry.register<Config>(
             Registry.configBuilder
                 .apiKey(apiKey)
@@ -101,6 +87,12 @@ object Klaviyo {
         )
 
         registerForLifecycleCallbacks(applicationContext)
+
+        Registry.registerOnce<State> {
+            KlaviyoState().also { state ->
+                Registry.register<StateSideEffects>(StateSideEffects(state))
+            }
+        }
 
         Registry.get<ApiClient>().startService()
 
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
@@ -14,7 +14,6 @@ import com.klaviyo.analytics.model.Profile
 import com.klaviyo.analytics.model.ProfileKey
 import com.klaviyo.analytics.networking.ApiClient
 import com.klaviyo.analytics.networking.KlaviyoApiClient
-import com.klaviyo.analytics.networking.requests.ResolveDestinationResult
 import com.klaviyo.analytics.state.KlaviyoState
 import com.klaviyo.analytics.state.State
 import com.klaviyo.analytics.state.StateSideEffects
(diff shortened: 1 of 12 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -90,6 +88,12 @@ object Klaviyo {
 
         registerForLifecycleCallbacks(applicationContext)
 
+        Registry.registerOnce<State> {
+            KlaviyoState().also { state ->
+                Registry.register<StateSideEffects>(StateSideEffects(state))
+            }
+        }
+
```
evan-masseau, line 96: Tried to make State work without waiting for company ID. But it was too complicated to have State side effects (like push token/permission observers) when API key is missing. Besides, it is just _weird_ to allow stateful operations like setting profile data or create events if we won't know the company till later.

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
```diff
@@ -216,6 +219,30 @@ internal class KlaviyoApiClientTest : BaseTest() {
...
+    @Test
+    fun `Enqueues a aggregate event API call`() {
+        assertEquals(0, KlaviyoApiClient.getQueueSize())
+
+        KlaviyoApiClient.enqueueAggregateEvent(AggregateEventPayload("{}"))
+
+        assertEquals(1, KlaviyoApiClient.getQueueSize())
+        assertEquals(false, postedJob?.force)
+    }
+
```
evan-masseau, line 245: just noticed missing tests coverage

sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
```diff
@@ -108,7 +108,14 @@ object KlaviyoConfig : Config {
...
     override lateinit var sdkVersion: String private set
-    override lateinit var apiKey: String private set
+    private var _apiKey: String = ""
+    override var apiKey: String
+        get() = _apiKey.ifEmpty {
+            throw MissingAPIKey()
+        }
+        set(value) {
+            _apiKey = value
+        }
```
evan-masseau, line 118: This is kinda the key change for the whole PR: instead of throwing `MissingAPIKey` on `build()`, I'm now allowing a Config to be built without API key (so that Context can be captured). To maintain that API key validation for callers, I'm still throwing `MissingAPIKey` if you try to _access_ or _set_ an empty API key.

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -41,29 +41,29 @@ object Klaviyo {
...
-     * apps that are not able to [initialize] Klaviyo immediately on app launch, but would like to
-     * utilize Klaviyo Forms
+     * Your API key still must be provided as early as possible for full SDK functionality!
      *
      * @param applicationContext
      */
     fun registerForLifecycleCallbacks(applicationContext: Context) = safeApply {
-        val application = applicationContext.applicationContext as? Application
-        application?.apply {
+        if (!Registry.isRegistered<Config>()) {
```
evan-masseau, line 53: Extending the use of `registerForLifecycleCallbacks` a little, to capture the provided context so we can use it in subsequent calls. In particular, needed it to be able to resolve a universal tracking link's destination URL and return it to the host app (or fallback to enqueueing the click track API request)

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -41,29 +41,29 @@ object Klaviyo {
...
+            // Register a partial config, missing API Key, to allow lifecycle tracking and context access for partial functionality
+            Registry.register<Config>(
+                Registry.configBuilder
+                    .applicationContext(applicationContext)
+                    .build()
+            )
+        }
+
+        // Some APIs (such as deep linking) work without an API key, so we can register the core service now
+        Registry.registerOnce<ApiClient> { KlaviyoApiClient }
```
evan-masseau, line 63: and API requests that do require an API key are all still behind the `Klaviyo` facade, and will have the same behavior as before if you call them before properly initializing.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -103,25 +103,25 @@ internal object KlaviyoApiClient : ApiClient {
         profile: Profile,
         callback: ResolveDestinationCallback
     ) {
+        // Create request outside of coroutine, so initialization errors are raised to the caller
+        val request = UniversalClickTrackRequest(trackingUrl, profile)
```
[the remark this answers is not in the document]
dan-peluso: Gosh the way coroutines hide some of these exceptions is so sneaky

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -41,29 +41,29 @@ object Klaviyo {
...
-                register<StateSideEffects>(StateSideEffects(state))
-            }
-        }
-    }
-
-    /**
-     * Use this method to register Klaviyo for lifecycle functions. This is necessary for
-     * apps that are not able to [initialize] Klaviyo immediately on app launch, but would like to
-     * utilize Klaviyo Forms
+     * Your API key still must be provided as early as possible for full SDK functionality!
```
dan-peluso, line 48: tellem!

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -41,29 +41,29 @@ object Klaviyo {
...
+                    .applicationContext(applicationContext)
+                    .build()
+            )
+        }
+
+        // Some APIs (such as deep linking) work without an API key, so we can register the core service now
+        Registry.registerOnce<ApiClient> { KlaviyoApiClient }
+
+        // Register lifecycle callbacks to monitor app foreground/background state
+        Registry.config.applicationContext.applicationContext.takeIf<Application>()?.apply {
```
dan-peluso, line 66: nit - should we just use the passed applicationContext instead of grabbing the registry one? Not sure it makes a huge difference but I imagine less interactions with the register is probably slightly better from a performance perspective

dan-peluso (approved):

sdk/core/src/main/java/com/klaviyo/core/KlaviyoException.kt
```diff
@@ -22,8 +22,12 @@ fun <ReturnType> safeCall(
 ): ReturnType? = try {
     block()
 } catch (e: KlaviyoException) {
-    Registry.log.error(e.message, e)
-    errorQueue?.add(block)
+    errorQueue?.apply {
+        Registry.log.warning("The operation will be retried.", e)
+        add(block)
```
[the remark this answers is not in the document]
evan-masseau: Yeah this isn't new, I just added the log line. It only retries subclasses of `KlaviyoException`, and only if the caller provided a queue to add the block to. We only actually use this for `handlePush`, to retry opened push events if the SDK isn't initialized yet.

PR #319, opened 2025-09-18
Testing the release verification action

evan-masseau:
When this one is _merged_, it should send an alert in slack bc this branch hasn't actually been updated to 4.1.0 in code yet.

### DIFF
--- a/.github/workflows/android-release.yml
+++ b/.github/workflows/android-release.yml
@@ -7,7 +7,6 @@ on:
   push:
     branches:
       - 'rel/*'
-      - 'ecm/*'
     tags:
       - '*'
 


.github/workflows/android-release.yml
```diff
@@ -4,9 +4,11 @@ name: Release Branch Version Check
 
 # Triggers the workflow when we open/update a PR from a rel/* branch
 on:
-  pull_request:
+  push:
```
evan-masseau, line 7: We're used to using pull request as a trigger, but in this case push is better: 
- It'd catch you out as soon as you make a rel/* branch without setting the rel branch
- It'll warn us every time we merge/push another update to that branch 
- Also checks pushed tags, technically this is post-release, but we have like 7 days after a tag is created before a release becomes immutable, plenty of time to correct it.

PR #322, opened 2025-09-18
Update trigger by event quote handling

dan-peluso:
# Description
We were getting some js failures around the profile event function due to the quotes that get added here. This change adds a new helper to the SDK which allows for sending the raw JS function and updates the unit test for this.

## Due Diligence

## Test Plan
Tested with assetSource and opening a push and those worked. Also sanity checked an existing profile event and that also worked. Here is a video of that:
[Screen_recording_20250918_142250.webm](https://github.com/user-attachments/assets/b573066b-d449-4939-96b9-b0e8e0b9f1fe)

### DIFF
--- a/sdk/forms/src/test/java/com/klaviyo/forms/bridge/KlaviyoJsBridgeTest.kt
+++ b/sdk/forms/src/test/java/com/klaviyo/forms/bridge/KlaviyoJsBridgeTest.kt
@@ -55,7 +55,7 @@ class KlaviyoJsBridgeTest : BaseTest() {
         verify {
             jsEvaluator.evaluateJavascript(
                 eq(
-                    """window.profileEvent("Fate Sealed","{"name":"Anna Karenina","location":"Saint Petersburg"}")"""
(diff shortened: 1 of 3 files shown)


sdk/forms/src/test/java/com/klaviyo/forms/bridge/KlaviyoJsBridgeTest.kt
```diff
@@ -55,7 +55,7 @@ class KlaviyoJsBridgeTest : BaseTest() {
         verify {
             jsEvaluator.evaluateJavascript(
                 eq(
-                    """window.profileEvent("Fate Sealed","{"name":"Anna Karenina","location":"Saint Petersburg"}")"""
+                    """window.profileEvent("Fate Sealed",{"name":"Anna Karenina","location":"Saint Petersburg"})"""
```
evan-masseau, line 58: 🤦 jeez it was right there in my face

evan-masseau (approved):

dan-peluso:
> Did you check this on iOS too btw?

Also cannot get it to work on ios but still working through the fix

PR #327, opened 2025-10-01
Fix crash in release builds

evan-masseau:
# Description

Discovered an error message and a crash in logcat while testing latest 4.1.0 release. The crash seems to be caused by using a typealias as a generic type with Registry functions. I'm not totally sure the exact mechanism, this is the error I saw: 
`java.lang.IllegalArgumentException: Class declares 0 type parameters, but 2 were provided`

I figured converting it to an `interface` would allow it to survive shrinkers, and that was borne out in testing. 

## Due Diligence

Can't really cover this with unit tests to my knowledge

## Test Plan

Tested by creating local signed release builds, with and without this change. Without, the app crashes as soon as code that interacts with the `DeepLinkHandler` type (registering one, or attempting to invoke it) is executed. With this change, the handler works fine in a release build.

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/linking/DeepLinking.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/linking/DeepLinking.kt
@@ -16,7 +16,9 @@ import com.klaviyo.core.lifecycle.LifecycleMonitor.Companion.ACTIVITY_TRANSITION
  * Callback type for handling a deep link. When registered, this callback is invoked with any
  * deep links originating from Klaviyo services, instead of broadcasting an [Intent].
  */
-typealias DeepLinkHandler = (uri: Uri) -> Unit
+fun interface DeepLinkHandler {
+    operator fun invoke(uri: Uri)
+}
 
 /**
  * Utility for handling any deep links into the host application originating from Klaviyo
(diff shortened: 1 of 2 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/model/Event.kt
```diff
@@ -23,7 +23,7 @@ class Event(val metric: EventMetric, properties: Map<EventKey, Serializable>?) :
         get() = when (val value = this[EventKey.VALUE]) {
             is Double -> value
             else -> try {
-                value.toString().toDouble()
+                value?.toString()?.toDouble()
```
evan-masseau, line 26: this was causing a lower priority issue, which is just that any event that had no `value` set would hit the exception below, because instead of returning `null` here, we'd cast to `"null"` then hit a number format exception 🤦

dan-peluso (approved):

sdk/analytics/src/main/java/com/klaviyo/analytics/linking/DeepLinking.kt
```diff
@@ -16,7 +16,9 @@ import com.klaviyo.core.lifecycle.LifecycleMonitor.Companion.ACTIVITY_TRANSITION
  * Callback type for handling a deep link. When registered, this callback is invoked with any
  * deep links originating from Klaviyo services, instead of broadcasting an [Intent].
  */
-typealias DeepLinkHandler = (uri: Uri) -> Unit
+fun interface DeepLinkHandler {
+    operator fun invoke(uri: Uri)
+}
```
evan-masseau, line 21: I'm not 🤓 enough about shrinkers to fully explain it, but using an interface instead of typealias seems to preserve the type or something, so that even in a release build you can safely use `Registry.register<DeepLinkHandler>()` and `Registry.get<DeepLinkHandler>()`

PR #329, opened 2025-10-02
Avoid crashing on unsupported links

evan-masseau:
# Description

Fixes an issue discovered in testing universal links where a universal tracking link that resolves to a destination url that the app's intent filters don't support, the app would crash. 
Relatedly, a push notification containing a link that doesn't belong to the host app would simply not open anything. This fixes that by verifying the intent first, and if the link is unsupported, log an error and fallback to using the app's launch intent (like we didn't have a deep link at all).

## Due Diligence

## Changelog / Code Overview

Added helper method for starting an activity from an intent only if that intent resolves to an activity
Added required queries permission for that intent filter resolution
Updated uses of `startActivity` where there's chance of ambiguity -- universal tracking links and external links from IAF webviews.

## Test Plan

Send yourself a universal tracking link whose destination url is not supported by the app (easiest way is to use the universal="true" attribute). When you open that link, the app should just open to the home page and not redirect (nor crash)

Same goes for including an unsupported link as a deep link on an in-app form. 

Send yourself a notification with a url the app doesn't support. When you tap on it without this fix, it'll just do nothing. After this fix, it'll open the app. 

## Related Issues/Tickets
 
 https://klaviyo.atlassian.net/browse/CHNL-26487

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/linking/DeepLinking.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/linking/DeepLinking.kt
@@ -11,6 +11,7 @@ import com.klaviyo.analytics.networking.requests.ResolveDestinationResult
 import com.klaviyo.analytics.state.State
 import com.klaviyo.core.Registry
 import com.klaviyo.core.lifecycle.LifecycleMonitor.Companion.ACTIVITY_TRANSITION_GRACE_PERIOD
+import com.klaviyo.core.utils.startActivityIfResolved
 
 /**
  * Callback type for handling a deep link. When registered, this callback is invoked with any
@@ -91,7 +92,7 @@ object DeepLinking {
      * @param extras Optional bundle of extras to be added to the launch intent
      */
     fun sendLaunchIntent(context: Context, extras: Bundle? = null) {
-        makeLaunchIntent(context, extras)?.let { context.startActivity(it) }
+        makeLaunchIntent(context, extras)?.startActivityIfResolved(context)
     }
 
     /**
@@ -102,8 +103,8 @@ object DeepLinking {
     private fun sendDeepLinkIntent(uri: Uri) {
         Registry.lifecycleMonitor.runWithCurrentOrNextActivity(
             ACTIVITY_TRANSITION_GRACE_PERIOD
-        ) { activity ->
-            activity.startActivity(makeDeepLinkIntent(uri, activity))
+        ) { context ->
+            makeDeepLinkIntent(uri, context).startActivityIfResolved(context)
         }
     }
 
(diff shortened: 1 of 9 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/linking/DeepLinking.kt
```diff
@@ -102,8 +103,8 @@ object DeepLinking {
     private fun sendDeepLinkIntent(uri: Uri) {
         Registry.lifecycleMonitor.runWithCurrentOrNextActivity(
             ACTIVITY_TRANSITION_GRACE_PERIOD
-        ) { activity ->
-            activity.startActivity(makeDeepLinkIntent(uri, activity))
```
evan-masseau, line 106: Previously threw an `ActivityNotFound` exception

dan-peluso (approved):

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
```diff
@@ -87,13 +103,9 @@ class KlaviyoNotification(private val message: RemoteMessage) {
      * @return Whether a message was displayed
      */
     @WorkerThread
+    @Suppress("MissingPermission")
```
evan-masseau, line 106: Couldn't figure out another way to make this method testable with mocks, but if the permission check isn't in the body of the function, android studio won't compile the code, so I have to explicitly suppress the check :-/

dan-peluso: Mocking context associated things truly the bane of being an android dev

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
```diff
@@ -133,7 +145,7 @@ class KlaviyoNotification(private val message: RemoteMessage) {
      * @param context
      * @return [Notification.Builder] to display
      */
-    private fun buildNotification(context: Context): NotificationCompat.Builder =
+    internal fun buildNotification(context: Context): NotificationCompat.Builder =
```
dan-peluso, line 148: 👍

dan-peluso (approved):

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
evan-masseau: Yeah... they should have an annotation like `@ChecksPermission("NAME")` that indicates a custom method checks for that permission and satisfies the other check.

PR #331, opened 2025-10-07
Formerly Add formsDataLoaded event -- Now just test fix and minor comment edit

evan-masseau:
# Description

~an attempt at adding `formsDataLoaded` and using it as the basis for starting the profile event observer.~
Never mind -- realized we can buffer from `handShook` to `formsDataLoaded` on the fender side for slightly simpler pattern. 

#330 

## Due Diligence

### DIFF
--- a/sdk/analytics/src/test/java/com/klaviyo/analytics/state/KlaviyoStateTest.kt
+++ b/sdk/analytics/src/test/java/com/klaviyo/analytics/state/KlaviyoStateTest.kt
@@ -63,7 +63,7 @@ internal class KlaviyoStateTest : BaseTest() {
         state.onProfileEvent(observer)
 
         val job = launch(Dispatchers.IO) {
-            state.createEvent(mockk(), mockk())
+            state.createEvent(mockk(relaxed = true), mockk())
         }
 
         val job2 = launch(Dispatchers.Default) {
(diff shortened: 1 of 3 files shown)


sdk/forms/src/main/java/com/klaviyo/forms/bridge/JsBridgeObserver.kt
```diff
@@ -5,6 +5,13 @@ package com.klaviyo.forms.bridge
  */
 internal interface JsBridgeObserver {
 
+    /**
+     * The message that triggers this observer to start observing
+     * By default, this is [NativeBridgeMessage.JsReady], meaning the observer
+     * will start observing as soon as the local JS asset is ready.
+     */
+    val startOn: NativeBridgeMessage get() = NativeBridgeMessage.JsReady
```
dan-peluso, line 13: I guess I am curious what the mapping is of our observer types and which message they become 'ready' on is. Does this make sense for the lifecycle observer for example?

evan-masseau: Yeah had the same thought about lifecycle events. But we don't actually use them for anything right now so I don't have a good answer either way

dan-peluso (approved):

PR #350, opened 2025-10-27
Use suspend for ApiClient methods that make a call immediately

evan-masseau:
# Description

Converts the two API methods that run immediately and return a result to use async/suspend functions. Maintained the callback versions for Java support and so as not to make a breaking change to SDK's API surface.

## Due Diligence

## Changelog / Code Overview

Suspend functions allow the caller to handle threading concerns rather than the API client implementation, makes more sense this way. However, the callback version is valuable for Java users since evidently it is difficult to use kotlin's coroutines from Java.

## Test Plan

Unit tests updated
Test deep links still work as expected with test app. 
Propagate this up to the sample app + geofencing PR and test with sample app (needs to be done for all the geofencing stuff anyway)

Tested in the internal test app. Deep linking still working, and API status broadcasts work as expected
Screenshot 2025-10-28 at 14 08 50

## Related Issues/Tickets
 
 See base geofencing branch PR
 

---

> [!NOTE]
> Converts immediate API calls to suspend functions (with legacy callbacks retained), updates deep linking to resolve URLs via coroutines and UI-thread handler, and adds safe coroutine utilities with comprehensive tests.
> 
> - **Networking (ApiClient/KlaviyoApiClient)**:
>   - Add suspend variants: `suspend resolveDestinationUrl(...)` and `suspend fetchGeofences()`; retain callback overloads for Java interop.
>   - Implement suspend flows with `withContext(Registry.dispatcher)` and refactor click-tracking resolution via `resolveOrEnqueue()`; use `safeLaunch` for callbacks.
>   - Streamline request broadcasting by notifying on status transitions via `sendAndBroadcast()`.
> - **Deep Linking**:
>   - Invoke `DeepLinkHandler` on UI thread; resolve universal links asynchronously using suspend API and `safeLaunch`.
> - **Core**:
>   - Enhance `safeCall` to catch non-Klaviyo exceptions (rethrow in debug); add `CoroutineScope.safeLaunch` helper.
> - **Tests**:
>   - Update and expand unit tests for coroutine-based APIs, UI-thread handling, and error handling; add `KlaviyoExceptionTest`; adjust dispatchers in tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7b198e31d8c4ce22c9aaff7603eb5bbb85826ac7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/sdk/core/src/main/java/com/klaviyo/core/KlaviyoException.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/KlaviyoException.kt
@@ -1,15 +1,16 @@
 package com.klaviyo.core
 
+import com.klaviyo.core.config.KlaviyoConfig
 import java.util.Queue
 import kotlin.coroutines.CoroutineContext
 import kotlin.coroutines.EmptyCoroutineContext
+import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.CoroutineStart
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 
 typealias Operation<T> = () -> T
-typealias SuspendOperation = suspend CoroutineScope.() -> Unit
 
 /**
  * Base class for exceptions thrown within the Klaviyo SDK
@@ -19,9 +20,15 @@ typealias SuspendOperation = suspend CoroutineScope.() -> Unit
 abstract class KlaviyoException(final override val message: String) : Exception(message)
 
 /**
- * Safely invoke a function and log KlaviyoExceptions rather than crash
- * Take care not to nest [safeCall] invocations, because the inner exception
+ * Safely perform an operation, catch and log Exceptions rather than crash.
+ * Primarily meant to catch [KlaviyoException] as a guard against invalid operations such as invoking Klaviyo prior to initializing.
+ * In production, this also safeguards the host app against crashing due to uncaught exceptions in Klaviyo code.
+ *
+ * Note: Take care not to nest [safeCall] / [safeApply] invocations, because the inner exception
  * will not halt execution of the outer method.
+ *
+ * Warning: This is not a substitute for proper exception handling at the call site,
+ * reserve this for safeguarding against wholly unexpected failures.
  */
 fun <ReturnType> safeCall(
     errorQueue: Queue<Operation<ReturnType>>? = null,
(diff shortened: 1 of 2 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -104,53 +104,68 @@ internal object KlaviyoApiClient : ApiClient {
...
-            // Get the result of the request
-            val result = request.getResult()
-
-            // Invoke the callback with the result (and notify observers)
-            withContext(Dispatchers.Main) {
-                broadcastApiRequest(request)
-                callback(result)
-            }
+    ): ApiRequest = UniversalClickTrackRequest(trackingUrl, profile).apply {
+        CoroutineScope(Registry.dispatcher).safeLaunch {
```
dan-peluso: Is there any concern about this `safeLaunch` only catching klaviyo exceptions and not all exceptions?

dan-peluso: Also have you looked at all into `CoroutineExceptionHandler` ?

sdk/core/src/main/java/com/klaviyo/core/KlaviyoException.kt
```diff
@@ -38,3 +45,25 @@ fun <Caller> Caller.safeApply(
...
+/**
+ * Launch a coroutine wrapped in a generic exception handler
+ */
+fun CoroutineScope.safeLaunch(
+    context: CoroutineContext = EmptyCoroutineContext,
+    start: CoroutineStart = CoroutineStart.DEFAULT,
+    errorQueue: Queue<SuspendOperation>? = null,
+    block: SuspendOperation
+): Job = launch(context, start) {
+    try {
```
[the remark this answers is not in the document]
evan-masseau: Mirroring safeCall, which is done that way bc it's only meant as a backstop for klaviyo config errors. Other types of exceptions should really be handled closer to the call site where we actually have the context to know what to do with that exception

I did look at that coroutine exception handler too, but it seemed over broad for what I was trying to do here. 

All that said, open to looking in to additional coroutine error handling too if necessary

ab1470 (approved):
nice modernization. Looks good to me, although my qualifications to give it a thorough review are limited

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
evan-masseau: Yeah i added this to address the fact that with the callback-based approach, a config/initialization error would be caught before launching the co-routine, but with the suspend function it has to be handled within the coroutine's scope. Otherwise the exception handling for API logic is already robust. 

specfically, this:
``` 
        // Create request outside of coroutine, so initialization errors are raised to the caller
```

PR #353, opened 2025-11-06
Changes to analytics events module to support some geofence event work

evan-masseau:
# Description

I broke out the changes to `analytics` needed for adding geofence transition events support. The main change here is to generalize (and improve) the "special snowflake" logic that we use for `$opened_push` events to prioritize those events being created ASAP. Also exposes the "enriched" event data post API call. 

## Due Diligence

## Changelog / Code Overview

- Make EventMetric.CUSTOM open so we can extend it to implement geofence-specific event types in location module
- Add a flag for special klaviyo events, using the established klaviyo convention of $ prefixes
- Treat klaviyo events as high priority (like we do for push) by putting them at the head of the queue and flushing immediately on track
- Return the enriched Event object so that caller can access the UUID from the API request

## Test Plan

Tested this with the sample app considerably

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/state/PersistentObservableProfile.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/state/PersistentObservableProfile.kt
@@ -29,7 +29,7 @@ internal class PersistentObservableProfile(
                         profile[key] = deserializeValue(json.get(key))
                     }
                 }
-            } catch (e: Throwable) {
+            } catch (e: Exception) {
                 Registry.log.warning("Invalid stored JSON for $key", e)
                 null
             }
--- a/sdk/core/src/main/java/com/klaviyo/core/KlaviyoException.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/KlaviyoException.kt
@@ -43,7 +43,7 @@ fun <ReturnType> safeCall(
         Registry.log.error(e.message, e)
     }
     null
-} catch (e: Throwable) {
+} catch (e: Exception) {
     if (KlaviyoConfig.isDebugBuild) {
         // To avoid development blindness, re-throw uncaught exceptions in DEBUG builds
         throw e
--- a/sdk/location/src/main/AndroidManifest.xml
+++ b/sdk/location/src/main/AndroidManifest.xml
@@ -5,6 +5,8 @@
 
     <application>
         <!-- Register Klaviyo service for receiving geofence broadcasts -->
-        <receiver android:name="com.klaviyo.location.KlaviyoGeofenceReceiver"/>
+        <receiver
+            android:name="com.klaviyo.location.KlaviyoGeofenceReceiver"
+            android:exported="false" />
     </application>
 </manifest>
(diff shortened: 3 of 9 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -170,22 +169,29 @@ internal object KlaviyoApiClient : ApiClient {
...
     /**
-     * Enqueues an [KlaviyoApiRequest] to run in the background
+     * Enqueues one or more [KlaviyoApiRequest]s to send on a background thread
      * These requests are sent to the Klaviyo asynchronous APIs
      *
      * This method will initialize the API queue and the batching thread
      * if this is the first request made since launch.
      */
-    fun enqueueRequest(vararg requests: KlaviyoApiRequest) {
+    fun enqueueRequest(vararg requests: KlaviyoApiRequest, headOfLine: Boolean = false) {
```
evan-masseau: @ajaysubra I thought you'd be curious to see this update, prompted by your suggestion on an earlier PR (this is the same code, I just broke it up into 2 PRs for ease of review). 

This argument allows the priority events like opened push and geofence events to jump the queue, so in addition to flushing immediately, we'll move it to the front of the line (if there is a line anyway)

ajaysubra: Nice. @belleklaviyo FYI.

sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
```diff
@@ -175,22 +175,24 @@ internal class KlaviyoState : State {
...
-        val shadowedEvent = event.copy().apply {
-            uniqueId = uniqueId ?: apiRequest.uuid
-            setProperty(EventKey.TIME, apiRequest.queuedTime)
-            DeviceProperties.buildEventMetaData().forEach { entry ->
-                setProperty(entry.key, entry.value)
+    override fun createEvent(event: Event, profile: Profile): Event = Registry.get<ApiClient>()
+        .enqueueEvent(event, profile)
+        .let { apiRequest ->
+            // Copy and enrich event with metadata and time (equivalent formating to the API request)
+            event.copy().apply {
```
dan-peluso: or does this prevent us from the list copy method I sent above? I guess I'm worried about the state mutation in that case - is there a reason we need to enqueue the event before we set these properties?

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -170,22 +169,29 @@ internal object KlaviyoApiClient : ApiClient {
...
         if (apiQueue.isEmpty()) {
             initBatch()
         }
 
+        // Reverse the arg order if headOfLine is true, so that first arg winds up first in line
+        requests.takeIf { headOfLine }?.reverse()
```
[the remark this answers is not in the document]
evan-masseau: Yeah mutation in place is always a little unclear. TBH I didn't even realize `reversed` was another option. 

FWIW cursor bugged me about mutating an argument affecting the caller too, but that part doesn't concern me because this is only a list because its using `vararg requests`, which means the caller never owned a reference to the array I'm reversing here, they'd just call it like `enqueueRequest(req1, req2, req3, headOfLine = true)`

PR #358, opened 2025-11-20
Proguard tweaks

evan-masseau:
# Description

While reviewing logs from a production build I noticed some class names are dropped, while others are retained. So I took a pass over the consumer rules and found that we're being a bit inconsistent, and overly broad. I made the following updates, which we'll need to test in a prod build of both RN and Anrdoid (more below)

- Use `keepnames` across the board -- Keeping class names has minimal impact to shrinking, because proguard/R8 can still obfuscate the rest of the code. But keeping names addresses one of two categories of shrinker issues we hit in the past: that we sometimes use reflection to access class names as constants/keywords, which renaming breaks.
- Targeted `keep` rules -- Keeping a whole package prevents the shrinker from doing its job _at all_. We do still have some classes that we know get referenced dynamically by reflection, so we can be targeted to keep just those. 
- Modules own their own rules -- We were relying, for example, on the `push` module to tell the consumer to keep the `analytics` symbols. If someone happened to only use analytics package, the rule wouldn't have actually applied.

## Due Diligence

Targeting this at `feat/geofencing` in part because it includes location module changes, and because that is our only in progress feature for Q4 so it'd be a good to merge there so we include it in upcoming tests.

## Changelog / Code Overview

Potentially breaking changes that we need to test carefully in a release build of both RN and Android test apps:
- Using `keepnames` on all class names now, we were missing the push and location modules
- Keep _names_ from `core`, but not all symbols. I'm not sure why we ever kept all of `core`, I suspect it was just a precaution.
- Only keep names from `analytics`, not _all_ symbols
  - BUT keep all  `analytics.model` classes. Particularly _keywords_ are a case that we definitely reference dynamically and a shrinker might try to treeshake out.

## Test Plan

In a release build of both RN and Android, re-test the issues we've had with shrinkers in the past: 
- RN: creating an event using a named event keyword constant, like Opened Push or Viewed Product
- Both platforms:  deep linking with a handler callback function used to fail because the shrinker would drop an important typealias
- Both platforms: in-app forms used to fail because we use some class name constants as keywords. If they get renamed, it would break JS/Native bridging logic.

Here is a breakdown of size differences. The middle row is from using `keepnames class com.klaviyo.package { * }` which keeps names of all members, functions names, etc. I think we only need to keep class names. That row was actually larger, because before this change we were allowing location and push modules to be fully obfuscated. The third row is allowing obfuscator to rename members while retaining class names.

  | Version                                       | APK Size        | vs Baseline            | DEX Size        | vs Baseline           |
  |-----------------------------------------------|-----------------|------------------------|-----------------|-----------------------|
  | Baseline (feat/geofencing)                    | 2,490,642 bytes | -                      | 2,916,896 bytes | -                     |
  | Keepnames all members | 2,490,642 bytes | 0 bytes                | 2,922,816 bytes | +5,920 bytes          |
  | This branch (ecm/location-proguard)             | 2,474,258 bytes | -16,384 bytes (-0.66%) | 2,852,312 bytes | -64,584 bytes (-2.2%) |

## Related Issues/Tickets
 
Might sound silly, but this is related to enterprise readiness and SLOs for SDK size. This will allow shrinkers to be a little more effective on our SDK.

### DIFF
--- a/sdk/location/consumer-rules.pro
+++ b/sdk/location/consumer-rules.pro
@@ -1,0 +1,2 @@
+# Keep only class names for stack traces, allow member obfuscation
+-keepnames class com.klaviyo.location.**


sdk/analytics/consumer-rules.pro
```diff
@@ -1 +1,4 @@
--keep class com.klaviyo.core.** { *; }
\ No newline at end of file
+# Keep only class names for stack traces, allow member obfuscation
+-keepnames class com.klaviyo.analytics.**
```
dan-peluso: damn so seems like we were doing this completely incorrectly before

evan-masseau:
> Changes look fine but what are your thoughts on not including this with geofencing changes? I'm a little worried of the extent we'll have to test these production build changes in addition to geofencing stuff (also because I feel like we only have been burned by proguard at seemingly random production times)

@dan-peluso True, I could split it up. I did it this way because I originally was just adding the proguard rules for location module. I could also just pause this until we have a `rel/4.2.0` branch so it isn't simply targeting `feat/geofencing`. Getting close to that anyway.

dan-peluso:
> > Changes look fine but what are your thoughts on not including this with geofencing changes? I'm a little worried of the extent we'll have to test these production build changes in addition to geofencing stuff (also because I feel like we only have been burned by proguard at seemingly random production times)
> 
> @dan-peluso True, I could split it up. I did it this way because I originally was just adding the proguard rules for location module. I could also just pause this until we have a `rel/4.2.0` branch so it isn't simply targeting `feat/geofencing`. Getting close to that anyway.

But we definitely want this in the release? Idk I'm conflicted cause it seems like we maybe got this very wrong in our original fix, but I also am nervous that if we do a mass geofencing release and this change is coupled with that then it could be trickier to debug

PR #361, opened 2025-11-21
Fix API request error handling to prevent premature request dropping

evan-masseau:
# Description
Fixes error handling in the API request layer that can cause network-related failures or server errors to be treated as permanent failures, resulting in requests being dropped from the queue instead of retried.

This is something we've been aware of for a bit but haven't had the time to address. My investigation into geofence events brought it back to the fore, because in some cases geofence events were dropped because the device was offline (usually, it was just not even _attempting_ the request and just putting it in the queue. But I did observe dropped requests at least once).

## Due Diligence

  - Part of geofencing feature work

### Changes Made

**1. Fixed IOException Handling** (`KlaviyoApiRequest.kt`)
- Added detection for network-related IOExceptions (SocketTimeoutException, SocketException, UnknownHostException)
- Network errors now return `Status.PendingRetry` instead of `Status.Failed` when under max retry attempts
- Only returns `Status.Failed` after exhausting all retry attempts or for non-network IOExceptions
- Added appropriate logging at WARNING level for retries, ERROR for permanent failures

**2. Fixed 5xx Server Error Handling** (`KlaviyoApiRequest.kt`)
- Added explicit handling for all 5xx server errors (500-599 range)
- Server errors treated as temporary issues that should be retried
- Returns `Status.PendingRetry` when under max attempts, `Status.Failed` only after exhausting retries
- 4xx client errors still fail immediately (correct behavior for bad requests, auth failures, etc.)

**3. Added Comprehensive Tests** (`KlaviyoApiRequestTest.kt`)
- 17 new test cases covering network IOExceptions, server errors, and retry behavior
- Tests verify PendingRetry vs Failed status under various conditions
- Updated existing test that incorrectly expected 500 errors to fail immediately

### Why This Matters

**Before**: Requests encountering temporary network issues (Doze mode, brief outages, socket timeouts) or server-side temporary failures could be permanently dropped from the queue. This was rare, because we do check the network connection right before attempting a send, and Klaviyo servers are very resilient. But we have had a few short incidents where webservers are down and 502-ing and that would have led to data loss.

**After**: Requests properly retry until network becomes available or server recovers, ensuring delivery of critical events like geofence transitions, profile updates, and push token registrations.

### Unit Tests
All 338 tests pass, including 17 new tests:

**IOException Scenarios:**
- SocketTimeoutException → returns PendingRetry ✅
- SocketException → returns PendingRetry ✅
- UnknownHostException → returns PendingRetry ✅
- IOException with "network" in message → returns PendingRetry ✅
- Network error after max attempts → returns Failed ✅
- Non-network IOException → returns Failed immediately ✅

**Response Code Scenarios:**
- 500, 502, 503, 504 → returns PendingRetry ✅
- 5xx after max attempts → returns Failed ✅
- 400, 401, 403, 404 → returns Failed immediately ✅

**Run tests:**
\`\`\`bash
export JAVA_HOME=~/Library/Java/JavaVirtualMachines/temurin-17.0.15/Contents/Home
./gradlew :sdk:analytics:testDebugUnitTest
\`\`\`

### Integration Testing Recommendations

1. **Doze Mode Testing**: Enable Doze via adb and verify geofence events eventually send after maintenance window
2. **Network Flapping**: Simulate network interruptions and verify requests retry
3. **Server Error Simulation**: Mock 503 responses and verify retry with exponential backoff

### Lint
\`\`\`bash
./gradlew ktlintFormat
\`\`\`
All style checks pass ✅

## Related Issues/Tickets

Part of geofencing feature work. This fix addresses the underlying issue where API requests were being permanently dropped due to transient network conditions, which particularly affects background operations like geofence event reporting.

Complements the WorkManager implementation (separate PR) that ensures requests are attempted during Doze maintenance windows.

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
@@ -68,7 +68,6 @@ internal open class KlaviyoApiRequest(
         const val HTTP_ACCEPTED = HttpURLConnection.HTTP_ACCEPTED
         const val HTTP_MULT_CHOICE = HttpURLConnection.HTTP_MULT_CHOICE
         const val HTTP_RETRY = 429 // oddly not a const in HttpURLConnection
-        const val HTTP_UNAVAILABLE = HttpURLConnection.HTTP_UNAVAILABLE
         const val HTTP_BAD_REQUEST = HttpURLConnection.HTTP_BAD_REQUEST
 
         // JSON keys for persistence
@@ -433,15 +432,8 @@ internal open class KlaviyoApiRequest(
 
         status = when (responseCode) {
             in successCodes -> Status.Complete
-            HTTP_RETRY, HTTP_UNAVAILABLE -> {
-                if (attempts < maxAttempts) {
-                    Status.PendingRetry
-                } else {
-                    Status.Failed
-                }
-            }
-            in 500..599 -> {
-                // All 5xx server errors are retryable (temporary server issues)
+            // 429 rate limit, and all 5xx server errors are treated as retryable
+            HTTP_RETRY, in 500..599 -> {
                 if (attempts < maxAttempts) {
                     Status.PendingRetry
                 } else {


evan-masseau:
bugbot run

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -383,8 +437,15 @@ internal open class KlaviyoApiRequest(
                     Status.Failed
                 }
             }
-            // TODO - Special handling of unauthorized i.e. 401 and 403?
```
amber-klaviyo: it looks like this todo was removed without dedicated logic being added (as far as I can tell). What was the initial reason for wanting special handling for 4XX and why did you ultimately decide it wasn't needed?

evan-masseau: Ah yeah that was an old old thought, where if company ID were configured incorrectly, we could just like stop sending API requests at all or something until a new API key was set. But having thought about that more in intervening time, that sounds too dangerous.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -326,8 +368,20 @@ internal open class KlaviyoApiRequest(
...
+            val isRetryable = isRetryableIOException(ex)
+
+            status = if (isRetryable && attempts < Registry.config.networkMaxAttempts) {
+                Registry.log.warning(
+                    "Retryable I/O error on attempt $attempts: ${ex.javaClass.simpleName}",
+                    ex
+                )
+                Status.PendingRetry
+            } else {
+                val reason = if (!isRetryable) "non-retryable" else "max attempts reached"
```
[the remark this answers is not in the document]
evan-masseau: Yea, this will log both the reason for deciding to fail (non-retryable exception, vs a retryable exception that just ran out of tries) as well as the exception class name (ProtocolException etc)

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -222,7 +222,7 @@ internal object KlaviyoApiClient : ApiClient {
             Status.Unsent -> Registry.log.verbose("${request.type} Request enqueued")
             Status.Inflight -> Registry.log.verbose("${request.type} Request inflight")
             Status.PendingRetry -> {
-                val attemptsRemaining = Registry.config.networkMaxAttempts - request.attempts
+                val attemptsRemaining = request.maxAttempts - request.attempts
```
evan-masseau: This was just for a logging, caught it while adjusting the other uses though.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/requests/KlaviyoApiRequest.kt
```diff
@@ -298,6 +304,42 @@ internal open class KlaviyoApiRequest(
...
+    /**
+     * Determines if an IOException (or any exception in its cause chain) is retryable.
+     * Based on patterns from Google's java-core library, we check both the immediate
+     * exception type and traverse the cause chain for wrapped retryable exceptions.
+     *
+     * @param exception The exception to check
+     * @param depth Current recursion depth to prevent infinite loops (max 10 levels)
+     * @return true if the exception or any of its causes are retryable network errors
+     */
+    private fun isRetryableIOException(exception: Throwable?, depth: Int = 0): Boolean {
```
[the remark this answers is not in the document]
evan-masseau: I was able to test a few of these exceptions do indeed get retried now, via proxyman. Namely timeout and ssl exceptions.

PR #374, opened 2025-12-02
Clean up geofence intent broadcast receiver

evan-masseau:
# Description

Removes the async aspects of geofence broadcast receiver. Queueing of the requests is synchronous, and the work manager added in #367 should handle the asynchronous aspect of actually sending the API requests now.

## Due Diligence

## Test Plan

Substantial testing description in this slack thread
https://klaviyo.slack.com/archives/C092G8ASQH1/p1764791056903339

## Related Issues/Tickets
 
Follow up from #367 which moved all the async logic of sending high priority events to a workmanager

### DIFF
--- a/sdk/location/src/main/java/com/klaviyo/location/KlaviyoLocationManager.kt
+++ b/sdk/location/src/main/java/com/klaviyo/location/KlaviyoLocationManager.kt
@@ -2,7 +2,6 @@ package com.klaviyo.location
 
 import android.annotation.SuppressLint
 import android.app.PendingIntent
-import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
 import com.google.android.gms.location.Geofence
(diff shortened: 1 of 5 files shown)


evan-masseau:
bugbot run

sdk/location/src/main/java/com/klaviyo/location/KlaviyoLocationManager.kt
```diff
@@ -333,144 +323,71 @@ internal class KlaviyoLocationManager : LocationManager {
      */
     override fun handleGeofenceIntent(
         context: Context,
-        intent: Intent,
-        pendingResult: BroadcastReceiver.PendingResult
+        intent: Intent
     ) {
         val geofencingEvent = GeofencingEvent.fromIntent(intent) ?: run {
```
ab1470: So now the only thing this `val` does is log a warning? Would it be more appropriate to call it something like `logGeofencingEvent`?

ab1470: Oh ok so looks like this instantiates a `GeofencingEvent`. Is everything after the `?:` like a `catch` clause that fires if the `GeofencingEvent` fails to instantiate?

dan-peluso (approved):

sdk/location/src/main/java/com/klaviyo/location/KlaviyoLocationManager.kt
```diff
@@ -333,144 +323,71 @@ internal class KlaviyoLocationManager : LocationManager {
...
-                // Clean up observer/timer
-                observer?.let { Registry.get<ApiClient>().offApiRequest(it) }
-                timeout?.cancel()
-
-                // And tell broadcast receiver that we've finished
-                pendingResult.finish()
+        // Create a geofencing event via analytics package for each geofence transition reported
+        geofencingEvent.triggeringGeofences?.map { it.toKlaviyoGeofence() }?.forEach { kGeofence ->
+            if (Registry.getOrNull<State>()?.apiKey == null) {
+                Registry.log.info("Automatically initialized Klaviyo from geofence event")
```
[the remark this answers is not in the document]
evan-masseau: At the moment, the plan is to go forward with both platforms able to auto-initialize like this, if the dev hasn't already

PR #384, opened 2025-12-15
Minor Key/Value pairs json refactor

evan-masseau:
# Description

Following up on my comment, just some code org suggestions. I had to look at it locally to think about where things belong, and figured quicker to just send you a PR with the suggestions. 

## Due Diligence

## Changelog / Code Overview

- Move JSONUtil to core
- Put all three of these constants that are used cross-package into core.Constants
- Fixed a few compile warnings.

## Test Plan

Unit test coverage should already be solid, but also here is a screenshot of an opened push event with proper key/values in it including the json decoded string for key_value_pairs
Screenshot 2025-12-15 at 12 09 50 PM

### DIFF
--- a/sdk/core/src/test/java/com/klaviyo/core/utils/JSONUtilTest.kt
+++ b/sdk/core/src/test/java/com/klaviyo/core/utils/JSONUtilTest.kt
@@ -5,7 +5,9 @@ import com.klaviyo.core.utils.JSONUtil.toHashMap
 import kotlin.collections.get
 import org.json.JSONArray
 import org.json.JSONObject
-import org.junit.Assert
+import org.junit.Assert.assertEquals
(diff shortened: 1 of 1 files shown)


sdk/core/src/main/java/com/klaviyo/core/Constants.kt
```diff
@@ -7,5 +7,15 @@ object Constants {
     /**
      * Package prefix used for Klaviyo intent extras and data keys
      */
-    const val PACKAGE_PREFIX = "com.klaviyo"
+    const val PACKAGE_PREFIX = "com.klaviyo."
```
evan-masseau: simpler to include the `.` here based on everywhere it was used.

PR #386, opened 2025-12-16
Add Java API interop tests

evan-masseau:
# Description
Adds Java unit tests that demonstrate and document Java interoperability patterns for the Klaviyo Android SDK. These tests verify that Java developers can call SDK methods correctly.

## Due Diligence
- [x] I have tested this on an emulator and/or a physical device.
- [x] I have added sufficient unit/integration tests of my changes.
- [ ] I have adjusted or added new test cases to team test docs, if applicable.
- [x] I am confident these changes are compatible with all Android versions the SDK currently supports.

## Release/Versioning Considerations
- [x] `Patch` Contains internal changes or backwards-compatible bug fixes.
- [ ] `Minor` Contains changes to the public API.
- [ ] `Major` Contains **breaking** changes.
- [ ] Contains readme or migration guide changes.
- [ ] This is planned work for an upcoming release.

Note: This PR only adds tests - no production code changes.

### New Test Files
| File | Purpose |
|------|---------|
| `KlaviyoJavaApiTest.java` | Tests all `Klaviyo` singleton methods from Java |
| `InAppFormsJavaApiTest.java` | Tests In-App Forms APIs from Java |
| `GeofencingJavaApiTest.java` | Tests Geofencing APIs from Java |

### New Mock Utilities (Kotlin)
Easier to write mocks for these new java tests in Kotlin.

| File | Purpose |
|------|---------|
| `KlaviyoMock.kt` | MockK bridge for mocking Klaviyo singleton from Java tests |
| `InAppFormsMock.kt` | Mocks for In-App Forms extensions + Duration helpers |
| `GeofencingMock.kt` | Mocks for Geofencing extensions |

## Test Plan
- [x] Tests of current Java implementation pass
- [ ] Next PR: fix Java bridging so that the commented out tests can pass as well, and more ergonomic syntax works
- [x] All existing Kotlin tests continue to pass
- [x] ktlint passes

## Related Issues/Tickets
- GitHub Issue: #385 (Java developers can't call `Klaviyo.initialize()`)
- Follow-up PR: #387 (adds annotations to enable clean Java syntax)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #387, opened 2025-12-16
Improve Java interoperability

### LINKED ISSUE #385: non-static method initialize(String,Context) cannot be referenced from a static context Klaviyo.initialize
### Checklist

- [x] I have read the [contributing guidelines](https://github.com/klaviyo/klaviyo-android-sdk/blob/master/.github/CONTRIBUTING.md)
- [ ] I have determined whether this bug is also reproducible in a vanilla Android project
- [ ] If possible, I've reproduced the issue using the `master` branch or latest release of this package.
- [x] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/klaviyo/klaviyo-android-sdk/issues) or [pull request](https://github.com/klaviyo/klaviyo-android-sdk/pulls).

### Description

```markdown
Hi, I'm trying to intialize Klaviyo, but cannot build my app as I'm receiving the error: 

error: non-static method initialize(String,Context) cannot be referenced from a static context
                Klaviyo.initialize(BuildConfig.KLAVIYO_API_KEY, application);
```

### Expected behavior

```markdown
I was expecting Klaviyou to be initialized with my API key
```

### Actual behavior

```markdown
App crashes. shows error.
```

### Steps to reproduce

```markdown
Install the latest build gradle 4.1.0

Intialize on the onCreate method:

`public class MainActivity extends AppCompatActivity {

    Toolbar toolbar;
    BottomNavigationView bottomNavigationView;

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
Klaviyo.initialize("KLAVIYO_PUBLIC_API_KEY",getApplicationContext());



`
```

### The Klaviyo Android SDK version information

4.1.0

### Device Information

_No response_

### Android Studio Version

_No response_

### Android API Level

_No response_

evan-masseau:
# Description
Improves Java interoperability by adding Kotlin annotations (`@JvmStatic`, `@JvmOverloads`, `@JvmName`) that generate clean static methods for Java callers. This addresses the compilation error reported in #385 and makes overall java integration more ergonomic.

Also expands README with java example code.

**Before:**
```java
Klaviyo.INSTANCE.initialize("api_key", context);
Klaviyo.INSTANCE.setEmail("test@example.com");
```

**After:**
```java
Klaviyo.initialize("api_key", context);
Klaviyo.setEmail("test@example.com");
```

## Due Diligence

Note: Tested via unit tests. Java tests from PR #386 verify the interop patterns work correctly.

**Note on API changes:** While this adds new static entry points, it's fully backward compatible:
- Existing Kotlin code continues to work unchanged
- Existing Java code using `INSTANCE` syntax continues to work (`@JvmStatic` generates both static and instance methods)

### Core SDK Changes
| File | Change |
|------|--------|
| `Klaviyo.kt` | Added `@JvmStatic` to all public methods; added static wrapper methods for Intent extension properties |
| `Profile.kt` | Added `@JvmOverloads` to constructor; resolve fluent setter ambiguity |
| `Event.kt` | Resolved fluent setter ambiguity in java |

### Module-Specific Changes
| File | Change |
|------|--------|
| `InAppFormsConfig.kt` | Added secondary constructors: no-arg (default timeout) and `Int` seconds for Java compatibility |
| `InAppForms.kt` | Added `KlaviyoForms` object with `@JvmStatic` methods for clean Java access (returns `Unit`) |
| `Geofencing.kt` | Added `KlaviyoLocation` object with `@JvmStatic` methods for clean Java access (returns `Unit`) |

### Documentation
- `README.md` - Added Java code examples throughout using collapsible `<details>` tags

### Technical Notes
- **MockK compatibility:** `@JvmStatic` requires `mockkStatic()` in addition to `mockkObject()` for proper stub recording. Without this, MockK executes the real static method during stub recording, causing `AbstractMethodError`.

## Test Plan
- [x] All existing unit tests pass
- [x] Java interop tests from PR #386 pass with new static syntax
- [x] ktlint passes

## Related Issues/Tickets
- Closes #385 (Java developers can't call `Klaviyo.initialize()`)
- Builds on PR #386 (Java interop tests)

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,10 @@ Profile identifiers and other attributes can be set all at once using the `Profi
    <summary>Kotlin</summary>
 
    ```kotlin
+   import com.klaviyo.analytics.Klaviyo
+   import com.klaviyo.analytics.model.Profile
+   import com.klaviyo.analytics.model.ProfileKey
+
    val profile = Profile(
        externalId = "USER_IDENTIFIER",
        email = "kermit@klaviyo-demo.com",
@@ -228,10 +232,11 @@ Profile identifiers and other attributes can be set all at once using the `Profi
    import com.klaviyo.analytics.Klaviyo;
    import com.klaviyo.analytics.model.Profile;
    import com.klaviyo.analytics.model.ProfileKey;
+   import java.io.Serializable;
    import java.util.HashMap;
    import java.util.Map;
 
-   Map<ProfileKey, String> properties = new HashMap<>();
+   Map<ProfileKey, Serializable> properties = new HashMap<>();
    properties.put(ProfileKey.FIRST_NAME.INSTANCE, "Kermit");
    properties.put(new ProfileKey.CUSTOM("instrument"), "banjo");
 
@@ -252,6 +257,9 @@ Or individually with additive fluent setters:
    <summary>Kotlin</summary>
 
    ```kotlin
+   import com.klaviyo.analytics.Klaviyo
+   import com.klaviyo.analytics.model.ProfileKey
+
    Klaviyo.setExternalId("USER_IDENTIFIER")
        .setEmail("kermit@klaviyo-demo.com")
        .setPhoneNumber("+12223334444")
@@ -286,6 +294,9 @@ to overwrite it with a new profile object.
    <summary>Kotlin</summary>
 
    ```kotlin
+   import com.klaviyo.analytics.Klaviyo
+   import com.klaviyo.analytics.model.ProfileKey
+
    // Start a profile for Kermit
    Klaviyo.setEmail("kermit@klaviyo-demo.com")
        .setPhoneNumber("+12223334444")
@@ -343,6 +354,11 @@ Additional event properties can be specified as part of the `Event` object:
    <summary>Kotlin</summary>
 
    ```kotlin
+   import com.klaviyo.analytics.Klaviyo
+   import com.klaviyo.analytics.model.Event
+   import com.klaviyo.analytics.model.EventKey
+   import com.klaviyo.analytics.model.EventMetric
+
    val event = Event(EventMetric.VIEWED_PRODUCT)
        .setProperty(EventKey.CUSTOM("Product"), "Coffee Mug")
        .setValue(10.0)
@@ -408,6 +424,9 @@ Add the following to your application or main activity's `.onCreate()` method:
    <summary>Kotlin</summary>
 
    ```kotlin
+   import com.google.firebase.messaging.FirebaseMessaging
+   import com.klaviyo.analytics.Klaviyo
+
    // Fetches the current push token and registers with Push SDK
    FirebaseMessaging.getInstance().token.addOnSuccessListener { pushToken ->
        Klaviyo.setPushToken(pushToken)
@@ -469,6 +488,10 @@ This method will check if the app was opened from a notification originating fro
    <summary>Kotlin</summary>
 
    ```kotlin
+   import android.content.Intent
+   import android.os.Bundle
+   import com.klaviyo.analytics.Klaviyo
+
    // Main Activity
 
    override fun onCreate(savedInstanceState: Bundle?) {
@@ -686,6 +709,7 @@ splash screen or loading animations have completed. Depending on your app's arch
    ```kotlin
    import com.klaviyo.analytics.Klaviyo
    import com.klaviyo.forms.registerForInAppForms
+   import com.klaviyo.forms.unregisterFromInAppForms
 
    // You can register as soon as you've initialized
    Klaviyo
@@ -694,6 +718,9 @@ splash screen or loading animations have completed. Depending on your app's arch
 
    // ... or any time thereafter
    Klaviyo.registerForInAppForms()
+
+   // ... Unregister later, if appropriate (e.g. on logout)
+   Klaviyo.unregisterFromInAppForms()
    ```
 </details>
 
@@ -710,6 +737,9 @@ splash screen or loading animations have completed. Depending on your app's arch
 
    // ... or any time thereafter
    KlaviyoForms.registerForInAppForms();
+
+   // ... Unregister later, if appropriate (e.g. on logout)
+   KlaviyoForms.unregisterFromInAppForms();
    ```
 </details>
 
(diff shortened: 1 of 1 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -361,28 +382,67 @@ object Klaviyo {
...
+
+    /**
+     * Checks whether a notification intent originated from Klaviyo
+     */
+    @Deprecated(
+        "Use isKlaviyoNotificationIntent instead, will be removed in the next major version",
+        ReplaceWith("isKlaviyoNotificationIntent")
+    )
+    val Intent.isKlaviyoIntent: Boolean
+        @JvmName("_isKlaviyoIntent")
```
evan-masseau: ick, don't like these hackarounds, just stick to the method java intends which is to create methods in place of the extension property

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -49,6 +49,7 @@ object Klaviyo {
      *
      * @param applicationContext
      */
+    @JvmStatic
```
evan-masseau: `@JvmStatic` allows the Java dev to call 
```
Klaviyo.initialize()
```

instead of 
```
Klaviyo.INSTANCE.initialize()
```

it is an additive change, so this doesn't break for folks currently using `INSTANCE`

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -285,6 +301,8 @@ object Klaviyo {
      * @param value [Double?] value to assign the event
      * @return Returns [Klaviyo] for call chaining
      */
+    @JvmStatic
+    @JvmOverloads
```
evan-masseau: `@JvmOverloads` creates overload methods when there are optional args (shocked this isn't default behavior)

sdk/analytics/src/main/java/com/klaviyo/analytics/model/Event.kt
```diff
@@ -29,13 +29,17 @@ class Event(val metric: EventMetric, properties: Map<EventKey, Serializable>?) :
                 null
             }
         }
+
+        @JvmSynthetic
```
evan-masseau: Previously, the fluent setter and property setter conflicted. This hides the auto-generated setter, and keeps the fluent setter, since they do the same thing anyway.

sdk/forms/src/main/java/com/klaviyo/forms/InAppFormsConfig.kt
```diff
@@ -13,7 +13,7 @@ import kotlin.time.Duration.Companion.seconds
  *  To disable session timeout altogether, use [Duration.INFINITE]
  */
 data class InAppFormsConfig(
-    private val sessionTimeoutDuration: Duration = DEFAULT_SESSION_TIMEOUT
+    private val sessionTimeoutDuration: Duration
```
evan-masseau: Kotlin duration doesn't translate nicely to Java unfortunately. As a result, you couldn't even call this constructor from Java, which cascaded up to not even being able to call `registerForInAppForms`. These secondary constructors make it possible to use `InAppFormsConfig` from Java without introducing any breaking changes to Kotlin code.

README.md
```diff
@@ -568,18 +779,38 @@ and drive location-based marketing.
 To begin, call `Klaviyo.registerGeofencing()` after initializing the SDK with your public API key.
 We recommend calling this as early as possible in your application lifecycle, ideally at app launch.
 
-```kotlin
-import com.klaviyo.analytics.Klaviyo
-import com.klaviyo.location.registerGeofencing
+<details open>
```
dan-peluso: nit what are your thoughts on consolidating the register / unregister logic so that we only have one code block example with all the functions? Just so we don't need distinctive sections. Total style thing though definitely feel free to ignore

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
dan-peluso: Yeah wow that is shocking

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -352,37 +372,78 @@ object Klaviyo {
...
         intent?.data?.let { uri ->
             DeepLinking.handleUniversalTrackingLink(uri)
         }
     } ?: intent.isKlaviyoUniversalTrackingIntent
 
+    /**
+     * Checks whether a notification intent originated from Klaviyo.
+     * Java-friendly static method.
+     */
+    @JvmStatic
```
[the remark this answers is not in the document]
dan-peluso: This fix makes sense, but I wonder if we should have some sort of way to test this with a java app as a 'just in case'. Typing that out I think it would probably be way more effort than its' worth' but maybe something to consider for future UI testing cases (if we are ever allowed to work on that)

sdk/forms/src/main/java/com/klaviyo/forms/InAppForms.kt
```diff
@@ -93,3 +93,35 @@ private fun Registry.inAppIsRegistered(): Boolean = listOf(
...
+    /**
+     * Start an In-App Forms session.
+     * Java-friendly static method.
+     *
+     * @param config see [InAppFormsConfig] for configuration options.
+     * @see Klaviyo.registerForInAppForms
+     */
+    @JvmStatic
+    @JvmOverloads
+    @UiThread
```
dan-peluso: wow totally forgot about this one

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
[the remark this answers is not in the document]
evan-masseau: yeah i kinda made the same assessment, a whole app wasn't worth it, but the new .java test files should show that all the prior syntax still work.

sdk/analytics/src/main/java/com/klaviyo/analytics/model/Event.kt
evan-masseau: "fluent setter" refers to method chaining, e.g. `.setValue(123).setUniqueId("ID")`

The property's setter conflicts with the version above that returns `this`, so i just keep the one that returns `this` and hid the one that returns void.

README.md
```diff
@@ -236,12 +333,32 @@ A list of common Klaviyo-defined event metrics is provided in `EventMetric`, or
...
-val event = Event(EventMetric.VIEWED_PRODUCT)
-    .setProperty(EventKey.CUSTOM("Product"), "Coffee Mug")
-    .setValue(10.0)
-Klaviyo.createEvent(event)
-```
+<details open>
+   <summary>Kotlin</summary>
+
+   ```kotlin
+   val event = Event(EventMetric.VIEWED_PRODUCT)
```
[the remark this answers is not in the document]
evan-masseau: We don't typically look at all of them at once like this, its a fair point!

PR #396, opened 2026-01-14
Add getForegroundInfo() to QueueFlushWorker for expedited work support

evan-masseau:
# Description
Add `getForegroundInfo()` implementation to QueueFlushWorker to support expedited WorkManager jobs on Android 11 and below.

On older Android versions, expedited work manager jobs are actually just implemented with a foreground service. So you have to implement this method, which puts a low priority notification in the system tray to indicate that the app is running in the background. So I added a very simple little "syncing" message (it flashes so fast that I can't even see it anyways, I can attest as a long time Android user this used to be a fairly common thing, you'd have to be watching pretty closely to even see it). 

[Additional context](https://klaviyo.slack.com/archives/C092G8ASQH1/p1768250371126219)

## Due Diligence

## Changelog / Code Overview
- Added `getForegroundInfo()` method to QueueFlushWorker that provides a low-priority notification for foreground service support
- Creates notification channel for SDK background operations (low importance, no badge)
- Uses system upload icon 
- Uses string resources for the text of the notification and channel name so the dev could override/localize it if desired.
- On Android 12+, expedited jobs don't run as foreground services, so notification won't be shown
- On Android 11 and below, provides required foreground service notification when expedited work runs.
- Using a low priority notification, and on this version of android, no user permission is required for notifications.

## Test Plan
- [x] Build and run on Android 11 and below to verify foreground service notification appears when queue flush is triggered
- [x] Verify notification uses upload icon
- [x] Test on Android 12+ to ensure expedited work still functions without notification

Here's what it looks like (it flashes so fast that I had to add an artificial delay to capture a screenshot)
Screenshot_20260114_112231

[Screen_recording_20260114_111319.webm](https://github.com/user-attachments/assets/1794f4a4-0aca-4d96-b854-bc45d31d560e)

## Impact
I was able to repro the bug, the user impact would have been pretty minimal, since it specifically only crashed while the app was backgrounded, so it wouldn't have interrupted user interactions. In some cases they might have gotten a "XYZ Application Stopped" alert. But it seems like if the app were running, it would just sync without using a foreground service.

## Related Issues/Tickets
 
CHNL-30224

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/QueueFlushWorker.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/QueueFlushWorker.kt
@@ -1,6 +1,5 @@
 package com.klaviyo.analytics.networking
 
-import android.R
 import android.app.Notification
 import android.content.Context
 import androidx.core.app.NotificationChannelCompat
@@ -11,6 +10,7 @@ import androidx.work.CoroutineWorker
 import androidx.work.ForegroundInfo
 import androidx.work.WorkerParameters
 import com.klaviyo.analytics.Klaviyo
+import com.klaviyo.analytics.R
 import com.klaviyo.core.Registry
 import com.klaviyo.core.config.Config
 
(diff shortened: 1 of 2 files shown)


dan-peluso (approved):
LGTM except the string thing

PR #398, opened 2026-01-21
Fix form activity presentation context

evan-masseau:
# Description

Flutter integration work exposed an issue with how we display the in-app form overlay activity. By using application context, we are forced to use the "new task" flag, which in turn can lead to the form winding up in its own task, separate from the host app's main activity.

## Due Diligence

Not yet planned, but I think at least a patch version 4.2.1 might be in order here.  Maybe we do a backward patch to 4.1 as well, not sure if worth it since this hasn't been externally reported, so it may be unique to flutter, or at least an activity setup that is uncommon outside of flutter

## Changelog / Code Overview

- Use the activity context instead of app context
- Take out the new task flag from the intent

## Test Plan

We never encountered this on RN or vanilla Android, but I suspect it might be an unfortunate timing thing on flutter. We should verify on Android, RN and Flutter that this still works to display the form above the app's primary task / main activity. 

## Related Issues/Tickets
 
 https://klaviyo.atlassian.net/browse/CHNL-30554


dan-peluso:
🐎

PR #401, opened 2026-01-27
Android push action buttons

amber-klaviyo:
# Description
1. Include relevant code from Dan's POC for push action buttons. 
2. Update expected payload structure and parsing for action button data. Create a dedicated action_buttons array that includes id, label, action, and (optional) url. This new format is closely aligned with the iOS structure.  
3. Ensure working behavior for deep_links which opens the app to a deep link'd page
4. Ensure working behavior for open_app which opens the app without requiring a deep link or url

## Due Diligence

## Test Plan
Android 7+
https://github.com/user-attachments/assets/5b4a23ab-86ed-4220-acfe-bcbc78f09d8f

Android <6
https://github.com/user-attachments/assets/a64ec8e2-05d6-4508-a628-5fb150f72667

## Related Issues/Tickets
https://klaviyo.atlassian.net/browse/CHNL-30181?atlOrigin=eyJpIjoiZDczMDA5MmZmZjU0NDk4NWIxMThmNTE3Mzk2NmQ3YjIiLCJwIjoiaiJ9 
https://klaviyo.atlassian.net/browse/CHNL-30188?atlOrigin=eyJpIjoiM2M1ZjYxMjgxMmZlNDVlYWIxMWRiZjJiNDAzMzQzZjYiLCJwIjoiaiJ9

### DIFF
--- a/sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoRemoteMessage.kt
+++ b/sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoRemoteMessage.kt
@@ -144,6 +144,14 @@ object KlaviyoRemoteMessage {
         Registry.log.warning("Error converting string to URL", it)
     }.getOrNull()
 
+    private fun JSONObject.optNonBlankString(key: String): String? {
+        // JSONObject.optString returns the literal "null" for JSON null values.
+        if (isNull(key)) {
+            return null
+        }
+        return optString(key).takeIf { it.isNotBlank() }
+    }
+
     /**
      * Parse [Uri] to sound resource
      */
@@ -220,7 +228,7 @@ object KlaviyoRemoteMessage {
                         )
                         continue
                     }
-                    val label = jsonObject.optString("label").takeIf { it.isNotBlank() }
+                    val label = jsonObject.optNonBlankString("label")
 
                     // Validate required label field
                     if (label == null) {
@@ -235,7 +243,7 @@ object KlaviyoRemoteMessage {
                     // Create appropriate sealed class instance based on action type
                     when (actionType) {
                         ActionButton.TYPE_DEEP_LINK -> {
-                            jsonObject.optString("url").takeIf { it.isNotBlank() }?.let { url ->
+                            jsonObject.optNonBlankString("url")?.let { url ->
                                 ActionButton.DeepLink(
                                     label = label,
                                     url = url
--- a/sdk/push-fcm/src/test/java/com/klaviyo/pushFcm/KlaviyoRemoteMessageTest.kt
+++ b/sdk/push-fcm/src/test/java/com/klaviyo/pushFcm/KlaviyoRemoteMessageTest.kt
@@ -295,6 +295,25 @@ class KlaviyoRemoteMessageTest : BaseTest() {
     }
 
     @Test
+    fun `Test Action Button with null label is skipped`() {
+        val actionButtonsJson = JSONArray().put(
+            JSONObject()
+                .put("label", JSONObject.NULL)
+                .put("action", "open_app")
+        ).toString()
+
+        val messageWithActions = stubMessage.toMutableMap().apply {
+            put(ACTION_BUTTONS_KEY, actionButtonsJson)
+        }
+
+        val msg = mockk<RemoteMessage>()
+        every { msg.data } returns messageWithActions
+
+        val buttons = msg.actionButtons
+        assert(buttons == null)
+    }
+
+    @Test
     fun `Test DEEP_LINK Action Button without URL is skipped`() {
         val actionButtonsData = listOf(
             mapOf(
@@ -304,6 +323,26 @@ class KlaviyoRemoteMessageTest : BaseTest() {
             )
         )
         val actionButtonsJson = JSONArray(actionButtonsData).toString()
+
+        val messageWithActions = stubMessage.toMutableMap().apply {
+            put(ACTION_BUTTONS_KEY, actionButtonsJson)
+        }
+
+        val msg = mockk<RemoteMessage>()
+        every { msg.data } returns messageWithActions
+
+        val buttons = msg.actionButtons
+        assert(buttons == null)
+    }
+
+    @Test
+    fun `Test DEEP_LINK Action Button with null URL is skipped`() {
+        val actionButtonsJson = JSONArray().put(
+            JSONObject()
+                .put("label", "Deep Link")
+                .put("action", "deep_link")
+                .put("url", JSONObject.NULL)
+        ).toString()
 
         val messageWithActions = stubMessage.toMutableMap().apply {
             put(ACTION_BUTTONS_KEY, actionButtonsJson)


sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoRemoteMessage.kt
```diff
@@ -156,6 +157,77 @@ object KlaviyoRemoteMessage {
             }
         }
 
+    /**
+     * Parse action buttons from the iOS-aligned format
+     *
+     * Expected structure:
+     * [{"id":"...", "label":"...", "action":"deep_link|open_app", "url":"..."}]
+     */
+    val RemoteMessage.actionButtons: List<ActionButton>?
```
dan-peluso: What are your thoughts on making a new type for the return here / function we pass this to? If 3 buttons is the hard limit we could add some stronger types for the fields you're looking for using sealed class. Psuedocode but something like this:
```
  /**
   * Sealed class representing different types of notification action buttons
   */
  sealed class ActionButton {
      abstract val id: String
      abstract val label: String

      /**
       * Button that opens the app without navigating to a specific destination
       */
      data class OpenApp(
          override val id: String,
          override val label: String
      ) : ActionButton()

      /**
       * Button that opens the app and navigates to a deep link destination
       */
      data class DeepLink(
          override val id: String,
          override val label: String,
          val url: String
      ) : ActionButton()
  }
```

This is definitely going into some kotlin specifics, so I'm down to pair on this cause I think it's a large-ish ask but I think it would help around how we handle types further down the line.

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoRemoteMessage.kt
```diff
@@ -156,6 +157,77 @@ object KlaviyoRemoteMessage {
...
+        get() = this.data[KlaviyoNotification.ACTION_BUTTONS_KEY]?.let { jsonString ->
+            Registry.log.verbose("Parsing action_buttons from: $jsonString")
+            try {
+                val jsonArray = JSONArray(jsonString)
+                val buttons = mutableListOf<ActionButton>()
+                Registry.log.verbose("JSON array has ${jsonArray.length()} buttons")
+                for (i in 0 until jsonArray.length()) {
+                    val jsonObject = jsonArray.getJSONObject(i)
+                    val button = ActionButton(
+                        id = jsonObject.optString("id"),
```
dan-peluso: optString returns an empty string if that field isn't in the json - which I think could break logic further. We should do some sort of check that these strings aren't empty.

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoRemoteMessage.kt
```diff
@@ -156,6 +157,77 @@ object KlaviyoRemoteMessage {
...
+     * [{"id":"...", "label":"...", "action":"deep_link|open_app", "url":"..."}]
+     */
+    val RemoteMessage.actionButtons: List<ActionButton>?
+        get() = this.data[KlaviyoNotification.ACTION_BUTTONS_KEY]?.let { jsonString ->
+            Registry.log.verbose("Parsing action_buttons from: $jsonString")
+            try {
+                val jsonArray = JSONArray(jsonString)
+                val buttons = mutableListOf<ActionButton>()
+                Registry.log.verbose("JSON array has ${jsonArray.length()} buttons")
+                for (i in 0 until jsonArray.length()) {
```
dan-peluso: similarly I think this should probably have a hard cut at 3

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
```diff
@@ -146,9 +153,11 @@ class KlaviyoNotification(private val message: RemoteMessage) {
      * @param context
      * @return [Notification.Builder] to display
      */
-    internal fun buildNotification(context: Context): NotificationCompat.Builder =
-        NotificationCompat.Builder(context, message.channel_id)
-            .setContentIntent(makePendingIntent(context))
+    internal fun buildNotification(context: Context): NotificationCompat.Builder {
+        val requestCodeBase = generateId()
+        val actionRequestCodeBase = requestCodeBase + ACTION_REQUEST_CODE_OFFSET
```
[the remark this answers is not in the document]
amber-klaviyo: it is necessary and I cleaned up the code / added comments to help explain

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoRemoteMessage.kt
```diff
@@ -156,6 +162,140 @@ object KlaviyoRemoteMessage {
...
+                        )
+                        continue
+                    }
+
+                    val actionType = ButtonActionType.fromString(
+                        jsonObject.optString("action", "open_app")
+                    )
+
+                    // Create appropriate sealed class instance based on action type
+                    val button = when (actionType) {
```
dan-peluso: we could be very _kotlin_ about this and `?.let{}` after the `when{}` to avoid making the button variable at all, but super duper nit don't worry about it if that's confusing

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoRemoteMessage.kt
```diff
@@ -156,6 +162,140 @@ object KlaviyoRemoteMessage {
...
+                    button?.let {
+                        Registry.log.verbose("Parsed button $i: $it")
+                        buttons.add(it)
+                    }
+                }
+
+                Registry.log.verbose("Successfully parsed ${buttons.size} valid action buttons")
+                buttons.takeIf { it.isNotEmpty() }
+            } catch (e: Exception) {
+                Registry.log.warning(
```
dan-peluso:
```suggestion
                Registry.log.error(
```

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoRemoteMessage.kt
```diff
@@ -156,6 +162,140 @@ object KlaviyoRemoteMessage {
...
+
+        companion object {
+            /**
+             * Parse a string value to a ButtonAction, with a fallback to OPEN_APP
+             *
+             * @param value The string representation of the action
+             * @return The corresponding ButtonAction, or OPEN_APP if not recognized
+             */
+            fun fromString(value: String): ButtonActionType =
+                entries.find { it.value.equals(value, ignoreCase = true) } ?: OPEN_APP
```
dan-peluso: again more of a product thing but I do think it'd be a little odd to ignore the case for such a strong contract. But if we expect that to change from the backend (whether or not to send us upper or lowercase), you could argue the client should maybe be flexible enough to accomodate for that?

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoRemoteMessage.kt
```diff
@@ -156,6 +162,140 @@ object KlaviyoRemoteMessage {
...
+            override val action: ButtonActionType = ButtonActionType.OPEN_APP,
+            override val id: String,
+            override val label: String
+        ) : ActionButton()
+
+        /**
+         * Button that opens the app and navigates to a deep link destination
+         */
+        data class DeepLink(
+            override val action: ButtonActionType = ButtonActionType.DEEP_LINK,
```
dan-peluso: This field feels redundant in the context of the data class being of type 'DeepLink' already

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
```diff
@@ -232,4 +241,97 @@ class KlaviyoNotification(private val message: RemoteMessage) {
...
+     */
+    private fun NotificationCompat.Builder.addActionButtons(
+        context: Context,
+        requestCodeBase: Int
+    ): NotificationCompat.Builder {
+        val actionButtons = message.actionButtons ?: return this
+
+        // Parser has already validated and limited buttons to MAX_ACTION_BUTTONS
+        actionButtons.forEachIndexed { index, button ->
+            val requestCode = requestCodeBase + index + ACTION_REQUEST_CODE_OFFSET
```
dan-peluso: what does 'requestCode' actually represent here? Do we still need the offset if we're already incrementing by the index?

amber-klaviyo: I had added a comment clarifying this but it looks like it was removed at some point. I'll add it back in.

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
```diff
@@ -232,4 +241,97 @@ class KlaviyoNotification(private val message: RemoteMessage) {
...
+        )
+    }
+
+    private fun makeResolvedDeepLinkIntent(
+        context: Context,
+        index: Int,
+        url: String
+    ): Intent? {
+        val uri = url.toUri()
+        return DeepLinking.makeDeepLinkIntent(uri, context)
```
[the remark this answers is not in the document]
amber-klaviyo: I followed up with Belle on this, and we've decided not to support malformed action types in any capacity. A bad action type will generate a warning and the button will be skipped, same as if it were missing key data (like the label). The code has been updated to reflect this change.

Issue #407, opened 2026-02-06
4.2.0 introduced android.permission.ACCESS_BACKGROUND_LOCATION that messes up app review if geofencing feature not used


kovbe21-mindtech:
### Checklist

- [x] I have read the [contributing guidelines](https://github.com/klaviyo/klaviyo-android-sdk/blob/master/.github/CONTRIBUTING.md)
- [ ] I have determined whether this bug is also reproducible in a vanilla Android project
- [ ] If possible, I've reproduced the issue using the `master` branch or latest release of this package.
- [x] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/klaviyo/klaviyo-android-sdk/issues) or [pull request](https://github.com/klaviyo/klaviyo-android-sdk/pulls).

### Description

```markdown
The SDK defines the android.permission.ACCESS_BACKGROUND_LOCATION permission. Possibly added for geofencing support. Our app does not need this feature, but our merged manifest file now contains this permission as well.

Android Play store now wants us to declare its usage, though we do not use it in our app.

The app is a react-native expo app, this sdk is used via klaviyo-react-native-sdk.
```

### Expected behavior

```markdown
Do not include the permission at sdk level, instead have the developers declare it explicitly if needed.
```

### Actual behavior

```markdown
The android.permission.ACCESS_BACKGROUND_LOCATION is declared explicitly.
```

### Steps to reproduce

```markdown
Install version 4.2.0, check for merged manifest.
```

### The Klaviyo Android SDK version information

4.2.0

adammatis:
Unitl this will be fixed, it can be added to blockedPermissions list in expo app config

```
{
  "android": {
    "blockedPermissions": ["android.permission.ACCESS_BACKGROUND_LOCATION"]
  }
}
```

evan-masseau:
Thanks for reporting, and for sharing the temporary work around for expo. Looks like we'll have to strip that permission out of the SDK and have folks add it themselves if they're opting to use location features.

evan-masseau:
For vanilla React Native developers, here is an equivalent temporary workaround to remove the permission from your merged manifest, put this in your android app manifest:

```xml
  <manifest xmlns:android="http://schemas.android.com/apk/res/android"
      xmlns:tools="http://schemas.android.com/tools">

      <!-- Remove location permissions inherited from Klaviyo SDK -->
      <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" tools:node="remove" />
[...]

PR #408, opened 2026-02-06
Add noPermissions flavor to location module

evan-masseau:
Publishes a classifier-based variant of the location module that excludes location permission declarations from its manifest. Wrapper SDKs (React Native, Flutter) can consume this version to avoid adding location permissions by default, without using manifest merger hacks that affect other libraries.

Consumption:
- Default: com.klaviyo:location:VERSION (includes location permissions)
- New: com.klaviyo:location:VERSION:noPermissions@aar (no location permissions)

# Description

The new noPermissions module will be used for the hybrid platforms (RN/Flutter) so that we don't automatically add the dangerous `ACCESS_BACKGROUND_LOCATION` automatically for all developers on those platforms. On android, we'll continue to just point folks to use the default flavor, since on android you only have to opt in to the modules you are using. 

## Due Diligence

As far as android is concerned, this is not a breaking change at all, in fact we could technically publish this under the same version since its a new artifact. For React Native i'd argue it is a patch because it was a bug to include the permission in the first place, but we should communicate clearly to existing customers (migration guide, release notes) about the change. 

## Changelog / Code Overview

- Added new `location` module build flavor to exclude the dangerous background location permission that triggers additional play store review. 

## Test Plan

I have tested:
- That the default syntax for the location module has not changed
- That jitpack builds both artifact flavors
- Permissions are only included in the default artifact flavor
- Building the sample app against both artifacts compiles. Geofencing works as before with the default flavor, and with the `noPermissions` flavor it works if you include the permissions in the app's manifest. 
- Sample app still compiles if you use the `noPermission` artifact, and does not crash (because we are doing proper permission checks).  

## Related Issues/Tickets
 
 #407

### DIFF
--- a/sample/build.gradle.kts
+++ b/sample/build.gradle.kts
@@ -53,6 +53,9 @@ android {
         vectorDrawables {
             useSupportLibrary = true
         }
+
+        // Location module has permission flavors; use 'default' (includes location permissions)
+        missingDimensionStrategy("permissions", "default")
     }
 
     buildTypes {
--- a/.github/workflows/jitpack.yml
+++ b/.github/workflows/jitpack.yml
@@ -63,12 +63,21 @@ jobs:
             if [[ $output == *"BUILD FAILED"* ]]; then
               echo "Build Failed!"
               echo
-              echo "$output" 
+              echo "$output"
               exit 1
             elif [[ $output == *"BUILD SUCCESSFUL"* ]]; then
               echo "Build Complete!"
               echo
-              echo "$output" 
+              echo "$output"
+
+              # Verify location module noPermissions flavor was built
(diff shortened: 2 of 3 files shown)


sample/build.gradle.kts
```diff
@@ -53,6 +53,9 @@ android {
         vectorDrawables {
             useSupportLibrary = true
         }
+
+        // Location module has permission flavors; use 'default' (includes location permissions)
+        missingDimensionStrategy("permissions", "default")
```
evan-masseau: Just a sample app thing to specify the flavor of the location module, since the sample app doesn't have a default/noPermission flavor of its own, it has to specify what dimension to use.

PR #411, opened 2026-02-17
Switch noPermissions variant from classifier to separate artifactId

evan-masseau:
## Summary

- Replaces the classifier-based publishing (`com.klaviyo:location:VERSION:noPermissions@aar`) with a separate artifactId (`com.klaviyo:location-no-permissions:VERSION`)
- Adds a `singleVariant("noPermissionsRelease")` block so AGP generates a proper software component for the noPermissions flavor
- Creates a dedicated `MavenPublication` backed by `components["noPermissionsRelease"]`

## Motivation

The classifier + `@aar` approach suppresses transitive dependency resolution — consumers of the `noPermissions` artifact wouldn't automatically get `play-services-location` and other transitive deps at runtime. Publishing as a separate artifactId produces its own `.module` metadata file with proper dependency declarations.

## Changes

**`sdk/location/build.gradle.kts`**
- Added `singleVariant("noPermissionsRelease")` with sources and javadoc jars
- Replaced the `artifact(tasks.named("bundleNoPermissionsReleaseAar")) { classifier = "noPermissions" }` block with a new `create<MavenPublication>("noPermissionsRelease")` that publishes from the `noPermissionsRelease` component
- Split the single publication into two independent publications, one per flavor

## Consumer impact

| Before | After |
|--------|-------|
| `com.klaviyo:location:VERSION:noPermissions@aar` | `com.klaviyo:location-no-permissions:VERSION` |

The default artifact (`com.klaviyo:location:VERSION`) is unchanged.

## Test plan

- [x] JitPack build succeeds (label `jitpack:build` on this PR to trigger)
- [x] Verify `location-noPermissions` artifact has transitive dependencies in its `.module` file
- [x] Verify `location` artifact is unaffected
- [ ] Consume both artifacts from a test app to confirm dependency resolution

### DIFF
--- a/.github/workflows/jitpack.yml
+++ b/.github/workflows/jitpack.yml
@@ -70,14 +70,22 @@ jobs:
               echo
               echo "$output"
 
-              # Verify location-no-permissions artifact was published
-              if [[ $output != *"location-no-permissions"* ]]; then
+              # Verify all expected artifacts were published
+              expected_artifacts=("analytics" "core" "forms" "location" "location-no-permissions" "push-fcm")
+              missing=()
+              for artifact in "${expected_artifacts[@]}"; do
+                if ! echo "$output" | grep -q "Found artifact: com.klaviyo:${artifact}:"; then
+                  missing+=("$artifact")
+                fi
+              done
+
+              if [ ${#missing[@]} -ne 0 ]; then
                 echo
-                echo "ERROR: location-no-permissions artifact not found in build output!"
+                echo "ERROR: Missing artifacts: ${missing[*]}"
                 exit 1
               fi
               echo
-              echo "Verified: location-no-permissions artifact was published"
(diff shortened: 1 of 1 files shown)


evan-masseau:
bugbot run

dan-peluso (approved):

.github/workflows/jitpack.yml
```diff
@@ -70,14 +70,22 @@ jobs:
               echo
               echo "$output"
 
-              # Verify location module noPermissions flavor was built
-              if [[ $output != *"noPermissions"* ]]; then
+              # Verify all expected artifacts were published
```
evan-masseau: Noticed that this workflow wasn't verifying artifacts individually, so added this to check them all, not just the no-permissions one.

PR #413, opened 2026-02-19
Extract forms-core module with ContentProvider auto-registration

evan-masseau:
## Summary
- Extract lightweight `forms-core` module with `FormsProvider` interface and extension functions
- Full `forms` module auto-registers its provider via `FormsInitProvider` ContentProvider
- Extension functions throw `MissingModule` when provider absent (enables hybrid SDK feature detection)
- Add `MissingModule` exception class to core `Registry.kt`

## Changes
- **New module**: `sdk/forms-core/` — contains `FormsProvider` interface, `InAppFormsConfig`, extension functions (`registerForInAppForms`, `unregisterFromInAppForms`, `reInitializeInAppForms`), `KlaviyoForms` Java-friendly static wrapper
- **New files in `sdk/forms/`**: `KlaviyoFormsProvider` (implementation), `FormsInitProvider` (ContentProvider for auto-registration)
- **Moved from `sdk/forms/` to `sdk/forms-core/`**: `InAppFormsConfig.kt`, `InAppForms.kt` (refactored)
- **Updated**: `settings.gradle.kts`, `jitpack.yml`, ProGuard rules, `AndroidManifest.xml`, existing tests

## Architecture
`forms-core` is a lightweight module that:
1. Defines the `FormsProvider` interface and public API surface (extension functions + `KlaviyoForms`)
2. Can be depended on by wrapper SDKs (Flutter, React Native) without pulling in the full forms implementation
3. When the full `forms` module is present, it auto-registers `KlaviyoFormsProvider` via ContentProvider
4. When absent, extension functions throw `MissingModule` with a helpful error message

## Test plan
- [x] `./gradlew :sdk:forms-core:build` passes
- [x] `./gradlew :sdk:forms:build` passes
- [x] All SDK module tests pass (`core`, `analytics`, `forms-core`, `forms`, `push-fcm`, `location`)
- [x] `./gradlew ktlintCheck` clean
- [ ] Verify ContentProvider in merged manifest

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sdk/forms-core/src/main/java/com/klaviyo/forms/InAppForms.kt
+++ b/sdk/forms-core/src/main/java/com/klaviyo/forms/InAppForms.kt
@@ -22,11 +22,9 @@ import com.klaviyo.core.safeApply
 @UiThread
 fun Klaviyo.registerForInAppForms(
     config: InAppFormsConfig = InAppFormsConfig()
-): Klaviyo {
-    val provider = Registry.getOrNull<FormsProvider>()
-        ?: throw MissingKlaviyoModule("forms")
-    return safeApply { provider.register(config) }
-}
+): Klaviyo = Registry.getOrNull<FormsProvider>()?.let { provider ->
+    safeApply { provider.register(config) }
+} ?: throw MissingKlaviyoModule("forms")
 
 /**
  * Halt the In-App Forms services and observers,
@@ -35,11 +33,10 @@ fun Klaviyo.registerForInAppForms(
  * @throws MissingKlaviyoModule if the `com.klaviyo:forms` module is not on the classpath.
  */
 @UiThread
-fun Klaviyo.unregisterFromInAppForms(): Klaviyo {
-    val provider = Registry.getOrNull<FormsProvider>()
-        ?: throw MissingKlaviyoModule("forms")
-    return safeApply { provider.unregister() }
-}
+fun Klaviyo.unregisterFromInAppForms(): Klaviyo =
+    Registry.getOrNull<FormsProvider>()?.let { provider ->
+        safeApply { provider.unregister() }
+    } ?: throw MissingKlaviyoModule("forms")
 
 /**
  * Java-friendly static methods for In-App Forms.
--- a/sdk/location-core/src/main/java/com/klaviyo/location/Geofencing.kt
+++ b/sdk/location-core/src/main/java/com/klaviyo/location/Geofencing.kt
@@ -12,11 +12,10 @@ import com.klaviyo.core.safeApply
  *
  * @throws MissingKlaviyoModule if the `com.klaviyo:location` module is not on the classpath.
  */
-fun Klaviyo.registerGeofencing(): Klaviyo {
-    val provider = Registry.getOrNull<GeofencingProvider>()
-        ?: throw MissingKlaviyoModule("location")
-    return safeApply { provider.register() }
-}
+fun Klaviyo.registerGeofencing(): Klaviyo =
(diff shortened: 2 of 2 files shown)


sdk/forms/src/main/AndroidManifest.xml
```diff
@@ -4,5 +4,9 @@
         <activity
                 android:name=".presentation.KlaviyoFormsOverlayActivity"
                 android:theme="@style/KlaviyoFormsOverlayTheme"/>
+        <provider
+            android:name="com.klaviyo.forms.FormsInitProvider"
+            android:authorities="${applicationId}.klaviyo.forms.init"
+            android:exported="false" />
```
dan-peluso: I know I could ask AI but what would making this 'true' change?

sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoFormsProvider.kt
```diff
@@ -0,0 +1,68 @@
...
+import com.klaviyo.forms.presentation.PresentationManager
+import com.klaviyo.forms.webview.JavaScriptEvaluator
+import com.klaviyo.forms.webview.KlaviyoWebViewClient
+import com.klaviyo.forms.webview.WebViewClient
+
+/**
+ * Resets the In-App Forms listeners with the current configuration.
+ */
+@UiThread
+internal fun Klaviyo.reInitializeInAppForms(): Klaviyo {
```
dan-peluso: does this need to be updated more in the no-op state as well? To throw

sdk/forms/src/main/AndroidManifest.xml
evan-masseau: Who says you aren't asking an AI right now? 

`ContentProvider` has a lot more uses than what we're doing with it here, exporting it as actually a valid option when its used for data sharing between apps. In our case, I don't think it'd really make any difference since our isn't being used for any actual content querying/exposure. It'd just be some random extra noise on the host app's public surface.

sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoFormsProvider.kt
evan-masseau: nah, this one's only used internally anyways (note `internal`)

evan-masseau: (and it always was)

sdk/forms-core/consumer-rules.pro
```diff
@@ -0,0 +1,2 @@
+# Keep only class names for stack traces, allow member obfuscation
+-keepnames class com.klaviyo.forms.**
```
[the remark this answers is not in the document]
dan-peluso: I guess we'll see when the prod test build completes! lol

dan-peluso (approved):
🚢 :shipit:

sdk/forms-core/consumer-rules.pro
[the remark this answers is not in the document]
evan-masseau: No — `FormsInitProvider` (the ContentProvider) lives in the `forms` module, not `forms-core`. The `forms/consumer-rules.pro` already has `-keep class com.klaviyo.forms.FormsInitProvider`. The `forms-core` module only has the interface/config/extensions, no ContentProvider, so the wildcard `-keepnames` is sufficient here (same pattern as `location-core`).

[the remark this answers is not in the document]
evan-masseau: Ah, thanks "me" for answering that

PR #414, opened 2026-02-23
feat: Add form lifecycle hooks for in-app forms

ajaysubra:
## Summary

Adds lifecycle event callbacks for in-app forms, enabling tracking of form interactions in third-party analytics platforms (Amplitude, Segment, etc.).

Addresses a customer request from Arlo who needs to track in-app form interactions in their existing analytics infrastructure.

## New Public API

- `Klaviyo.registerFormLifecycleCallback(callback: FormLifecycleCallback)` — Register a callback for form lifecycle events (replaces any previous callback)
- `Klaviyo.unregisterFormLifecycleCallback()` — Remove the callback

Java-friendly statics via `KlaviyoFormLifecycleCallbacks.registerFormLifecycleCallback(callback)` / `unregisterFormLifecycleCallback()`.

## Events

`FormLifecycleEvent` is a sealed interface with three subclasses, each carrying form metadata:

| Event | Properties | Fires when |
|-------|-----------|------------|
| `FormShown` | `formId`, `formName` | Form is presented to the user |
| `FormDismissed` | `formId`, `formName` | Form is dismissed (any reason) |
| `FormCtaClicked` | `formId`, `formName`, `buttonLabel`, `deepLinkUrl` | User taps a CTA button |

## Implementation Details

- Sealed interface pattern allows event-specific data (e.g. `buttonLabel` only on CTA events)
- All callbacks invoked on UI thread via `Registry.threadHelper.runOnUiThread()`
- Functional interface (`fun interface`) — Kotlin lambdas and Java SAM both work
- `FormLifecycleEvent.FormShown` fires after the overlay activity is created (not before)
- `FormLifecycleEvent.FormDismissed` is guarded against duplicate firing (e.g. timeout + bridge close)
- Uses `FormContext` internally to pass `formId`/`formName` through the presentation layer
- Bridge messages (`FormWillAppear`, `FormDisappeared`, `OpenDeepLink`) now carry `formId` and `formName`

## Files Changed

**New Files:**
- `FormContext.kt` — Internal data class for form metadata
- `FormLifecycleEvent.kt` — Sealed interface with event subclasses
- `FormLifecycleCallback.kt` — Functional callback interface
- `FormLifecycleUtils.kt` — Shared utility for invoking callbacks
- `FormLifecycleCallbackTest.kt` — Test suite

**Modified Files:**
- `KlaviyoFormsProvider.kt` — Public API extension functions + Java static wrappers
- `KlaviyoNativeBridge.kt` — Fires lifecycle callbacks from bridge messages
- `NativeBridgeMessage.kt` — Added `formName` to `FormWillAppear`/`FormDisappeared`/`OpenDeepLink`, added `buttonLabel` to `OpenDeepLink`
- `KlaviyoPresentationManager.kt` — Fires `FormShown`/`FormDismissed`, uses `FormContext`, guards duplicate dismiss
- `PresentationManager.kt` — Updated interface to accept `FormContext`
- `PresentationState.kt` — States now carry `FormContext` instead of bare `FormId`
- `SampleApplication.kt` — Demo of all three events
- `InAppFormsJavaApiTest.java` — Java API compatibility tests
- `KlaviyoNativeBridgeTest.kt`, `NativeBridgeMessageTest.kt`, `KlaviyoPresentationManagerTest.kt` — Updated/expanded tests

## Test Plan

- [x] All lifecycle callback tests pass
- [x] All existing forms tests pass (no regressions)
- [x] Sample app demonstrates all three events with toast notifications
- [x] Verified callbacks run on UI thread
- [x] Verified callback can be registered/unregistered multiple times
- [x] Verified forms work without callback registered
- [x] Verified Java API compatibility

## Example Usage

```kotlin
Klaviyo.registerFormLifecycleCallback { event ->
    when (event) {
        is FormLifecycleEvent.FormShown ->
            Analytics.track("Form Shown", mapOf("formId" to event.formId, "formName" to event.formName))
        is FormLifecycleEvent.FormDismissed ->
            Analytics.track("Form Dismissed", mapOf("formId" to event.formId))
        is FormLifecycleEvent.FormCtaClicked ->
            Analytics.track("Form CTA", mapOf("button" to event.buttonLabel, "url" to event.deepLinkUrl))
    }
}
```


sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoNativeBridge.kt
```diff
@@ -29,6 +32,12 @@ import com.klaviyo.forms.webview.WebViewClient
  */
 internal class KlaviyoNativeBridge() : NativeBridge {
 
+    /**
+     * Tracks the most recently presented form context so it remains accessible after the form is dismissed.
+     * Needed for v2 protocol where [FormDisappeared] arrives before [OpenDeepLink] (see [handShakeData]).
+     */
+    private var lastFormContext: FormContext? = null
```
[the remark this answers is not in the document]
evan-masseau: `KlaviyoPresentationManager` has a `presentationState` property that tracks the current open form ID, I wonder if we should augment that rather than add a second place that we're tracking "form context"

evan-masseau (commented):
Will be testing this out against https://github.com/klaviyo/fender/pull/56023 locally when i get a second today

evan-masseau:
gonna have to do a gross rebase on this, its behind from the last android release, including the refactor that split the forms module into forms and forms-core causing some of the files we touched here to be deleted.

Easiest way was to flatten all these commits into one and fix the rebase in one shot, otherwise it woulda been 25 consecutive cleanups :gross:

sample/src/main/java/com/klaviyo/sample/SampleApplication.kt
```diff
@@ -22,6 +27,26 @@ class SampleApplication : Application() {
                 // If not using a deep link handler, Klaviyo will send an Intent to your app with the deep link in intent.data
                 showToast("Deep link to: $uri")
             }
+            .registerFormLifecycleCallback { event ->
+                // OPTIONAL SETUP NOTE: Register a callback to receive form lifecycle events
+                // This allows you to track when forms are shown, dismissed, or when CTAs are clicked
+                when (event) {
+                    is FormShown -> {
+                        Registry.log.debug("Form shown: ${event.formId} ${event.formName}")
+                        showToast("Form shown: ${event.formId} ${event.formName}")
```
dan-peluso: nit could we make this configurable as a toggle so you don't get a toast every time a form action happens? just feels a bit noisy if we're also doing a log

sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoNativeBridge.kt
```diff
@@ -92,8 +95,11 @@ internal class KlaviyoNativeBridge() : NativeBridge {
     /**
      * Notify the client that the webview should be shown
      */
-    private fun show(bridgeMessage: FormWillAppear) = Registry.get<PresentationManager>()
-        .present(bridgeMessage.formId)
+    private fun show(bridgeMessage: FormWillAppear) {
+        Registry.get<PresentationManager>().present(
+            FormContext(bridgeMessage.formId, bridgeMessage.formName)
```
dan-peluso: This probably isn't the place for this comment but is it worth adding the new form id from this context to any of our analytics calls? Or do those already exist there

dan-peluso: nevermind we already send it

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoPresentationManager.kt
```diff
@@ -105,10 +110,16 @@ internal class KlaviyoPresentationManager() : PresentationManager {
...
+    override fun dismiss(formContext: FormContext?) = overlayActivity?.let { activity ->
         clearTimers()
         Registry.get<WebViewClient>().detachWebView()
         activity.finish()
+        presentationState.takeIfNot<PresentationState, Hidden>()?.let {
+            val context = formContext ?: it.formContext
+            invokeFormLifecycleCallback(
+                FormLifecycleEvent.FormDismissed(context?.formId, context?.formName)
+            )
+        }
```
dan-peluso: would we want to do any of the above actions if the presentation state is hidden? Like why would we still try and finish the activity or detach the webview / timers if we're coming from a hidden state? What I'm getting at is that if this is a race condition maybe we outta throw here or not do some of the above, but I'll admit it's a bit tricky to think about that case. I just don't want a duplicate accidental message from onsite to get these misaligned

sdk/forms/src/main/java/com/klaviyo/forms/presentation/PresentationState.kt
```diff
@@ -1,27 +1,30 @@
...
     data class Presented(
-        override val formId: FormId?
+        override val formContext: FormContext?
     ) : PresentationState {
         override fun toString(): String = "Presented:$formId"
     }
 
     data object Hidden : PresentationState {
-        override val formId = null
+        override val formContext = null
```
dan-peluso: I like this new class

sdk/forms/src/main/java/com/klaviyo/forms/FormContext.kt
```diff
@@ -0,0 +1,10 @@
+package com.klaviyo.forms
+
+/**
+ * Internal contextual metadata about the in-app form being presented.
+ * Used by [com.klaviyo.forms.presentation.PresentationState] to track the current form.
+ */
+internal data class FormContext(
+    val formId: String?,
```
dan-peluso: I think these should be non nullable, I'm pretty sure a form with no id / name is impossible and I don't want one of these being created with missing information

sdk/forms/src/main/java/com/klaviyo/forms/FormLifecycleEvent.kt
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Use [formId] and [formName] to identify the form associated with any event.
+ * For CTA-specific data, match on [FormCtaClicked] to access [FormCtaClicked.buttonLabel]
+ * and [FormCtaClicked.deepLinkUrl].
+ */
+sealed interface FormLifecycleEvent {
+    /**
+     * The form ID of the form associated with this event, or null if unavailable.
+     */
+    val formId: String?
```
dan-peluso: feeling similarly here about keeping this and the name nullable. Always prefer non nullable types in kotlin if you can

sdk/forms/src/main/java/com/klaviyo/forms/FormLifecycleUtils.kt
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Invoke the registered form lifecycle callback on the UI thread, if one is registered.
+ *
+ * Shared utility used by both [com.klaviyo.forms.bridge.KlaviyoNativeBridge]
+ * and [com.klaviyo.forms.presentation.KlaviyoPresentationManager].
+ */
+internal fun invokeFormLifecycleCallback(event: FormLifecycleEvent) {
+    Registry.getOrNull<FormLifecycleCallback>()?.let { callback ->
+        Registry.threadHelper.runOnUiThread {
```
dan-peluso: what are your thoughts on maybe making this the default runner but having devs be able to pass something else in for which thread to run it on? I know in our sample we're posting stuff to UI but I'm curious what the customers will be using it for, and whether or not they'll like that we're forcing this on the UI thread when they could have a forms heavy observation being used for something unrelated to UI

sdk/forms/src/test/java/com/klaviyo/forms/presentation/KlaviyoPresentationManagerTest.kt
```diff
@@ -19,6 +22,7 @@ import io.mockk.slot
 import io.mockk.unmockkObject
 import io.mockk.verify
 import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
 import org.junit.Test
 
 class KlaviyoPresentationManagerTest : BaseTest() {
```
dan-peluso, line 26: do we have anything for testing mismatched IDs / names? It's a bit hard to follow the test case coverage so I apologize if you already have this but I am concerned that with all the state management of form currently being shown that there might be some cases we don't have covered. Honestly though I'm not sure what kind of logic we're adding to prevent multiple form versions / names / ids messes so might not be too big an issue. I'm just worried about dynamic forms (like lets say one constantly being updated or getting a new name) coming from some very active marketers bleeding into SDK form mismatches.

dan-peluso (commented):
mostly questions, sorry been a while since I looked at the first version and wasn't sure if Ajay or Evan was working on this

sample/src/main/java/com/klaviyo/sample/SampleApplication.kt
ajaysubra: Gotcha. I'll remove this toast.

ajaysubra: Fixed in `d4cb1804c`. Removed the toast calls from the lifecycle callback — logs are kept for diagnostics. The `showToast` function remains since the deep link handler still uses it.

-Claude

sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoNativeBridge.kt
ajaysubra: Acknowledged — sounds like this is already covered since the analytics calls already include formId. No changes needed.

-Claude

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoPresentationManager.kt
ajaysubra: Good question. Looking at the `dismiss()` implementation, there are two guards already protecting against the Hidden-state race:

1. The `overlayActivity?.let` gate — when state is Hidden, `overlayActivity` is null (set to null on every dismiss), so the entire detach/finish block is skipped and we fall through to `clearTimers()` + log.
2. The `takeIfNot<Hidden>()` guard inside the activity block — even if we somehow had an overlayActivity reference, the FORM_DISMISSED callback only fires when transitioning *out* of a non-Hidden state.

So a duplicate `formDisappeared` message arriving after the timeout has already dismissed will be a no-op: no detach, no finish, no duplicate callback. Added a test for the mismatched context scenario in `d4cb1804c` to make this coverage more explicit.

-Claude

sdk/forms/src/main/java/com/klaviyo/forms/presentation/PresentationState.kt
ajaysubra: Thanks! The sealed interface fits nicely here since each state carries its own formContext and the compiler enforces exhaustive matching.

-Claude

sdk/forms/src/main/java/com/klaviyo/forms/FormContext.kt
ajaysubra: Fixed in `d4cb1804c`. Made `formId` and `formName` non-nullable (`String`) in `FormContext`. At bridge construction sites where the JSON values may be absent, we use `orEmpty()` to convert nulls to empty strings. Since `FormContext` is `internal`, the blast radius is contained within the SDK.

-Claude

sdk/forms/src/main/java/com/klaviyo/forms/FormLifecycleEvent.kt
ajaysubra: Fixed in `d4cb1804c`. Made `formId` and `formName` non-nullable (`String`) in the `FormLifecycleEvent` sealed interface and all its subtypes (`FormShown`, `FormDismissed`, `FormCtaClicked`). The callback will always receive populated strings — empty string in the (unlikely) case the data is missing from JS.

-Claude

sdk/forms/src/main/java/com/klaviyo/forms/FormLifecycleUtils.kt
ajaysubra: Good design question. The current approach always dispatches on the UI thread because the primary use case is updating UI in response to form events (as the sample app demonstrates). Making the runner configurable adds API surface for what may be a niche need — customers doing non-UI work in the callback can always dispatch to their own thread/coroutine from within the callback itself.

That said, if we see demand for it, we could add an optional `Executor` or `CoroutineDispatcher` parameter to `registerFormLifecycleCallback` in a future iteration without breaking the current API. Happy to revisit if you feel strongly about doing it now.

-Claude

sdk/forms/src/test/java/com/klaviyo/forms/presentation/KlaviyoPresentationManagerTest.kt
ajaysubra: Added two tests in `d4cb1804c` to cover this scenario:

1. `dismiss with different formContext uses provided context over state` — verifies that when `dismiss()` is called with a formContext that differs from what was presented, the callback receives the *provided* context (i.e., the one from the bridge message, not stale state).
2. `dismiss without formContext falls back to state context` — verifies the fallback path.

Regarding dynamic forms with constantly changing names/IDs: the current architecture only tracks one form at a time (Fender only sends one `formWillAppear`), and each present/dismiss cycle is atomic. A new `formWillAppear` while already presenting is rejected (the `takeIf<Hidden>` guard). So stale/mismatched state from rapidly changing forms should not be an issue in practice.

-Claude

ajaysubra:
bugbot run

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoPresentationManager.kt
ajaysubra: Good point. It's hard to solve for the race condition here since there is no good conflict resolution mechanism in onsite module and it does show both and the SDK isn't privy to which one got shown. Once we have a better way to more deterministically show one form at a time and also once we have message priority implemented this problem should go away.  

Also, we'll do a manual test and see what happens when two forms are shown at once.

sdk/forms/src/main/java/com/klaviyo/forms/FormContext.kt
ajaysubra: Will check fender to make sure this change makes sense

sdk/forms/src/main/java/com/klaviyo/forms/FormLifecycleUtils.kt
ajaysubra: Thanks for calling this out. 

couple points,
* We might already be on the main thread since this is invoked from the web view.
* if they want it on a particular thread they could do it inside the closure 
* looking forward if we add in the functionality of allowing them to block a form from being shown we would want that on the UI/main thread.

sdk/forms/src/test/java/com/klaviyo/forms/presentation/KlaviyoPresentationManagerTest.kt
ajaysubra: We removed the bigger state management logic we had in the initial version of this PR and treating the SDK mostly as a thin pass through layer.

sdk/forms/src/main/java/com/klaviyo/forms/FormContext.kt
dan-peluso: Yeah I'd definitely want this to match the fender models

sdk/forms/src/main/java/com/klaviyo/forms/FormLifecycleUtils.kt
dan-peluso: Would like to discuss this further - I don't think it'd be that difficult a change and I generally like to opt for more configurable design around thread use. I could see lifecycle events for forms being used outside of UI only applications - anyone who wants to internally manage state I think could run into this immediately

sample/src/main/java/com/klaviyo/sample/SampleApplication.kt
```diff
@@ -22,6 +27,23 @@ class SampleApplication : Application() {
...
+                    }
+                    is FormDismissed -> {
+                        Registry.log.debug("Form dismissed: ${event.formId} ${event.formName}")
+                    }
+                    is FormCtaClicked -> {
+                        Registry.log.debug(
+                            "Form CTA: ${event.buttonLabel} -> ${event.deepLinkUrl}"
+                        )
+                    }
+                }
```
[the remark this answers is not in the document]
evan-masseau: non-issue: sample app overrides to verbose

sdk/forms/src/main/java/com/klaviyo/forms/FormLifecycleUtils.kt
evan-masseau: I do like the convenience of passing it in as an arg. Looking ahead though, we will be adding at least one hook quite soon that passes a _callback_, something like

```kotlin
FormWillShow(formId: String, formName: String, allowForm: (Boolean) -> Unit)
```

Allowing the developer to invoke `allowForm(false)` to block the form. It'll be important that we keep that on the thread on which the JS bridge message came from (which should be main), so ~I think we should both remove the `runOnUiThread` jump that we have here now, and not complicate the API by adding an optional dispatcher/executor param here, because we'd have to differentially use it for this `FormWillShow` event.~ EDIT see below

We should call out the thread sensitivity in method docs/readme.

evan-masseau: Actually, sorry, strike that. On modern webview implementations, these would automatically be on main thread (making the call to `runOnUiThread` a no-op), but on older versions that use an outdated version of web/native bridging, it may be called on a background thread. For full consistency, and to provide the best outcome for when we add that `allowForm` caIlback feature, think we actually do need to keep the `runOnUiThread` here. 

Again, we can call it out in the docs. Any analytics library worth its salt will already be moving any heavy lifting to a background thread, but worth being clear. 

I really went full circle there

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoPresentationManager.kt
```diff
@@ -139,9 +139,10 @@ internal class KlaviyoPresentationManager() : PresentationManager {
...
         /**
          * Grace period to close a form with animation, before we just dismiss
-         *  the overlay activity without waiting for formDisappeared event
-         *  ~350ms for the animation, and a little padding.
+         *  the overlay activity without waiting for formDisappeared event.
+         *  Must exceed the JS close animation duration (~400ms) or the activity
+         *  teardown will kill the animation before formDisappeared is sent.
          */
-        private const val CLOSE_TIMEOUT = 400L
+        private const val CLOSE_TIMEOUT = 500L
```
evan-masseau: Preferred to have this shorter, but found that at 400ms, about 75% of the time, the webview didn't have enough time to finish the animation and send the `formDisappeared` event. Didn't look visually different to the user, but the FormDismissed hook event wouldn't fire. 500ms was sufficient that I didn't see any missed events over several tests... added an extra 100ms buffer for safety.

sdk/forms/src/main/java/com/klaviyo/forms/presentation/PresentationState.kt
```diff
@@ -1,27 +1,16 @@
...
-        override val formId: FormId?
-    ) : PresentationState {
-        override fun toString(): String = "Presented:$formId"
-    }
+    /** The overlay activity has been launched but is not yet created */
+    data object Presenting : PresentationState
 
-    data object Hidden : PresentationState {
-        override val formId = null
-        override fun toString(): String = "Hidden"
```
evan-masseau, line 25: Did we lose a `toString` method here? At least as compared to what is in master?

sdk/forms/src/main/assets/klaviyo-js-bridge.js
```diff
@@ -93,19 +93,16 @@ window.openForm = function (formId) {
...
  */
 window.closeForm = function (formId) {
     document.head.dispatchEvent(
         new CustomEvent(
             'closeForm',
             {
-                detail: {
-                    formId: formId
-                }
+                detail: {}
```
evan-masseau, line 105: just a note: fender, annoyingly, needs the detail property. fixing that separately, when that lands this could just be `CustomEvent('closeForm')`. Not a big deal, but I am fixing that in a [wee fender PR](https://github.com/klaviyo/fender/pull/57029/files)

sdk/forms/src/test/java/com/klaviyo/forms/FormLifecycleCallbackTest.kt
```diff
@@ -0,0 +1,185 @@
...
+        Klaviyo.registerFormLifecycleCallback(callback)
+
+        val message =
+            """{"type":"formWillAppear", "data":{"formId":"$testFormId","formName":"$testFormName"}}"""
+        nativeBridge.postMessage(message)
+
+        verify { mockPresentationManager.present() }
+        assertTrue(capturedEvent is FormLifecycleEvent.FormShown)
+        assertEquals(testFormId, capturedEvent!!.formId)
+        assertEquals(testFormName, capturedEvent!!.formName)
```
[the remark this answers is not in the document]
evan-masseau: We do try to avoid this

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoPresentationManager.kt
```diff
@@ -139,9 +139,10 @@ internal class KlaviyoPresentationManager() : PresentationManager {
...
         /**
          * Grace period to close a form with animation, before we just dismiss
-         *  the overlay activity without waiting for formDisappeared event
-         *  ~350ms for the animation, and a little padding.
+         *  the overlay activity without waiting for formDisappeared event.
+         *  Must exceed the JS close animation duration (~350ms) with a buffer for signaling delays
+         *  or else teardown will kill the webview's rendering before formDisappeared can be sent.
          */
-        private const val CLOSE_TIMEOUT = 400L
+        private const val CLOSE_TIMEOUT = 600L
```
evan-masseau, line 146: Note: I preferred to have this shorter, but found that at 400ms about 75% of the time, the webview didn't have enough time to finish the animation and send the formDisappeared event. Didn't look visually different to the user, but the FormDismissed hook event wouldn't fire. 500ms was sufficient that I didn't see any missed events over several tests... so added an extra 100ms buffer for safety.

sdk/forms/src/test/java/com/klaviyo/forms/FormLifecycleCallbackTest.kt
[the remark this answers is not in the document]
evan-masseau: Fixed in 7f4c27f — replaced `assertTrue` + `!!` with a smart cast (`as FormLifecycleEvent.FormShown`), which both asserts the type and enables safe property access without force-unwrap.

sdk/forms/src/main/java/com/klaviyo/forms/presentation/PresentationState.kt
evan-masseau: The explicit `toString()` methods were removed along with `formId` when these changed from `data class` to `data object`. Kotlin's `data object` auto-generates `toString()` returning the simple class name (e.g. `"Presenting"`, `"Hidden"`), which matches the old behavior minus the formId that was intentionally removed from the state. The log statements using `$presentationState` still produce clean output.

sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoNativeBridge.kt
```diff
@@ -115,19 +121,56 @@ internal class KlaviyoNativeBridge() : NativeBridge {
...
+
+        DeepLinking.handleDeepLink(deepLinkUri)
+        invokeFormLifecycleHandler(
+            FormLifecycleEvent.FormCtaClicked(
+                formId = message.formId,
+                formName = message.formName,
+                buttonLabel = message.buttonLabel,
+                deepLinkUrl = deepLinkUri
+            )
+        )
```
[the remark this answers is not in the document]
evan-masseau: yeah this isn't a "real world" scenario, and besides, if that is really the case, then its what we do with the deep link handling, so it would be appropriate to also not emit the lifecycle event.

evan-masseau (approved):

dan-peluso (approved):

PR #429, opened 2026-04-01
fix: downgrade androidx.core to 1.16.0 for compileSdk 35 compat

evan-masseau:
# Description

`core-ktx:1.17.0` requires `compileSdk 36`. By bumping to it in 4.3.1, we inadvertently broke customers still on `compileSdk 35` — Gradle's global dependency version resolution pulls `1.17.0` into their build transitively, causing compilation failures.

This downgrades the SDK modules back to `1.16.0` (the prior version) and pins the sample app to `1.17.0` explicitly so it can keep using APIs that need it.

## Due Diligence

## Changelog / Code Overview

- `versions.properties`: `version.androidx.core` reverted from `1.17.0` to `1.16.0`
- `sample/build.gradle.kts`: Replaced `AndroidX.core.ktx` (refreshVersions-managed, now resolving to 1.16.0) with a hardcoded `"androidx.core:core-ktx:1.17.0"` so the sample app stays on the newer version independently of the SDK modules.

## Test Plan

- CI build passes (no compileSdk change needed — the SDK already uses compileSdk 36 internally, this only affects transitive resolution for consumers)
- Verify a consumer project on `compileSdk 35` can resolve dependencies without being forced to 1.17.0

## Related Issues/Tickets

Resolves MAGE-448


evan-masseau:
bugbot run

sample/build.gradle.kts
```diff
@@ -133,7 +133,7 @@ dependencies {
 
     // Other app and testing dependencies (refreshVersions syntax)
     implementation(platform(AndroidX.compose.bom))
-    implementation(AndroidX.core.ktx)
+    implementation("androidx.core:core-ktx:1.17.0")
```
evan-masseau, line 136: Only reason we even bumped it was to be able to demonstrate a thing in the sample apps that only exists on API level 36+. Usually androidx is explicitly designed to be backward compatible, but guess not in this case. Anyway, this allows us to keep demo-ing that, while leaving the SDK's version of this lib at what it was.

ndurell (approved):

PR #435, opened 2026-04-07
fix: skip profile reset when setProfile called with same identifiers

jason-myers-klaviyo:
## Summary

Fixes excessive `POST /client/push-tokens` requests caused by `setProfile()` being called repeatedly with the same identifiers (e.g. on every screen load).

**Root cause:** `setProfile()` unconditionally called `reset()` whenever identifiers existed in state, even if the incoming identifiers were identical. Each `reset()` regenerated the anonymous ID, which triggered the flush-on-anonymous-ID-change guard in `StateSideEffects.onUserStateChange`. When a push token was in state, this produced **2 push-token API requests per `setProfile()` call** instead of the expected 1 (or 0 for unchanged profiles). Under rate limiting, this compounded into a request storm.

**Fix:** `setProfile()` now compares incoming identifiers against current state and only calls `reset()` when they actually differ. Anonymous ID is the lowest-order identifier — there's no reason to regenerate it when higher-order identifiers (externalId, email, phone) haven't changed. `resetProfile()` remains available for explicitly clobbering all state.

Single-file change in `KlaviyoState.kt`. All existing behavior for the on-resume push token refresh, push state deduplication, profile debouncing, and profile-consolidation-via-push-token-endpoint is preserved.

**Swift SDK counterpart:** klaviyo/klaviyo-swift-sdk#543
**Slack thread:** https://klaviyo.slack.com/archives/C09DKBLQ3PX/p1775509483748149

## Test plan

- [ ] Verify unit tests pass in CI
- [ ] `setProfile(same identifiers)` repeatedly — confirm no spurious push-token requests (identifiers unchanged → no reset → PersistentObservableProperty suppresses no-op sets)
- [ ] `setProfile(different identifiers)` — confirm reset still fires, anonymous ID regenerated, old profile flushed
- [ ] `setProfile(same identifiers, different attributes)` — confirm attributes still update and flush via debounce
- [ ] `resetProfile()` — confirm still clobbers all state as before
- [ ] Verify rate limiting behavior improves for affected customer

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
@@ -79,13 +79,6 @@ internal object KlaviyoApiClient : ApiClient {
 
     override fun enqueuePushToken(token: String, profile: Profile): ApiRequest =
         PushTokenApiRequest(token, profile).also {
-            // Supersede any previously queued push token request.
-            // Only one push-token request should be in the queue at a time;
-            // the latest one always carries the most up-to-date state.
-            apiQueue.filterIsInstance<PushTokenApiRequest>().forEach { existing ->
-                apiQueue.remove(existing)
-                Registry.dataStore.clear(existing.uuid)
-            }
             Registry.log.verbose("Enqueuing Push Token request")
             enqueueRequest(it)
         }
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
@@ -79,6 +79,13 @@ internal object KlaviyoApiClient : ApiClient {
 
     override fun enqueuePushToken(token: String, profile: Profile): ApiRequest =
         PushTokenApiRequest(token, profile).also {
+            // Supersede any previously queued push token request.
+            // Only one push-token request should be in the queue at a time;
+            // the latest one always carries the most up-to-date state.
+            apiQueue.filterIsInstance<PushTokenApiRequest>().forEach { existing ->
+                apiQueue.remove(existing)
+                Registry.dataStore.clear(existing.uuid)
+            }
             Registry.log.verbose("Enqueuing Push Token request")
             enqueueRequest(it)
         }
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
@@ -108,9 +108,17 @@ internal class KlaviyoState : State {
      */
     override fun setProfile(profile: Profile) {
         if (!externalId.isNullOrEmpty() || !email.isNullOrEmpty() || !phoneNumber.isNullOrEmpty()) {
-            // If a profile with external identifiers is already in state, we must reset.
-            // This conditional is important to preserve merging with an anonymous profile.
-            reset()
+            // Only reset if the incoming profile has different identifiers.
+            // Anonymous ID is the lowest-order identifier, so there's no reason to regenerate it
+            // when higher-order identifiers haven't changed. Resetting with the same identifiers
+            // causes unnecessary anonymous ID churn, which triggers spurious API requests.
+            // resetProfile() remains available for explicitly clobbering all state.
+            val identifiersChanged = profile.externalId != externalId ||
+                profile.email != email ||
+                profile.phoneNumber != phoneNumber
+            if (identifiersChanged) {
+                reset()
+            }
         }
 
         // Move any identifiers and attributes to their specified state variables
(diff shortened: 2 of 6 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -79,6 +79,13 @@ internal object KlaviyoApiClient : ApiClient {
 
     override fun enqueuePushToken(token: String, profile: Profile): ApiRequest =
         PushTokenApiRequest(token, profile).also {
+            // Supersede any previously queued push token request.
+            // Only one push-token request should be in the queue at a time;
+            // the latest one always carries the most up-to-date state.
+            apiQueue.filterIsInstance<PushTokenApiRequest>().forEach { existing ->
```
evan-masseau: this could cause data loss, we'd at _least_ have to merge requests together, since the push-token API request could contain other profile data that hasn't made it to klaviyo yet.

sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
```diff
@@ -108,9 +108,17 @@ internal class KlaviyoState : State {
...
         if (!externalId.isNullOrEmpty() || !email.isNullOrEmpty() || !phoneNumber.isNullOrEmpty()) {
-            // If a profile with external identifiers is already in state, we must reset.
-            // This conditional is important to preserve merging with an anonymous profile.
-            reset()
+            // Only reset if the incoming profile has different identifiers.
+            // Anonymous ID is the lowest-order identifier, so there's no reason to regenerate it
+            // when higher-order identifiers haven't changed. Resetting with the same identifiers
+            // causes unnecessary anonymous ID churn, which triggers spurious API requests.
+            // resetProfile() remains available for explicitly clobbering all state.
+            val identifiersChanged = profile.externalId != externalId ||
```
evan-masseau: i wonder if this should even be sensitive to an _additive_ change vs a change of an existing identifier. Like if you already had an email, and are now getting the same email + a phone, we probably don't have to reset either.

evan-masseau: @ndurell We should probably talk that over, we've historically avoided being too "smart" about stuff like that on the SDK side and leave it to the server to figure out the merging etc.

ndurell: Yah would we miss any important signal as a result of this? Can't set profile have other data that might be important to consider here?

sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
```diff
@@ -108,9 +108,17 @@ internal class KlaviyoState : State {
...
+                profile.email != email ||
+                profile.phoneNumber != phoneNumber
+            if (identifiersChanged) {
+                reset()
+            }
```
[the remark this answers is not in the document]
jason-myers-klaviyo: Claude says:

>No, this isn't a real issue. The scenario described is "setProfile called with identical identifiers" — which means the caller is not switching profiles. There's no "old profile" vs "new profile" to flush
  between.

>  The prior behavior (reset + re-identify with the same values) only appeared to have a "flush old, start new" contract. In reality it was flushing an empty anonymous profile created by the unnecessary reset — that
   was the bug. Any pending setAttribute() data sitting in the debounce timer will still flush when the timer fires, exactly as it would have before. The attributes are still in pendingProfile, the timer is still
  ticking, nothing changed for that path.

>  The bot is pattern-matching on "skipped code path = lost behavior" without understanding that the skipped path was producing garbage requests, not useful flushes.

sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
evan-masseau: The setters below are all observed, so any identifier change or attributes mutation will get propagated to an API queue, this will just prevent calling reset _first_, the thing causing the unecessary anon ID change. 

And to answer my own question: technically yes, if they were making only an additive identifier change, we don't need to change anonymous ID. But also that'd make no difference as far as API calls, because the additional identifier would need to be communicated anyway, and the anon ID change would effectively be dropped on the backend.

sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
```diff
@@ -108,9 +108,17 @@ internal class KlaviyoState : State {
...
-            reset()
+            // Only reset if the incoming profile has different identifiers.
+            // Anonymous ID is the lowest-order identifier, so there's no reason to regenerate it
+            // when higher-order identifiers haven't changed. Resetting with the same identifiers
+            // causes unnecessary anonymous ID churn, which triggers spurious API requests.
+            // resetProfile() remains available for explicitly clobbering all state.
+            val identifiersChanged = profile.externalId != externalId ||
+                profile.email != email ||
+                profile.phoneNumber != phoneNumber
+            if (identifiersChanged) {
```
evan-masseau: An SDK engineer will add unit test coverage of this before merging (and manually test obvs too)

dan-peluso: Agree that we should probably add some unit tests here

evan-masseau (approved):
Did a little readability cleanup, added unit tests, and tested it manually a whooole bunch. 
Created a patch release branch, and retargeted this PR to it.

I'll get a second MAGE approval before I hit merge

ajaysubra (approved):
easier to review with iOS and Android code matching this closely.

PR #437, opened 2026-04-10
fix(forms): validate required bridge message fields (MAGE-484)

evan-masseau:
## Summary

- Add `requireString()` extension on `JSONObject` that throws `IllegalArgumentException` when a required field (`formId`, `formName`, `buttonLabel`) is missing or empty from a bridge message
- The existing `catch` in `KlaviyoNativeBridge.postMessage()` logs the error and prevents malformed events from reaching the host app
- Updated all tests to supply required fields and added new tests for missing-field error paths

Part of MAGE-484

Motivated by feedback from Dan on the Flutter SDK PR (#37) where missing fields from the JS bridge were silently creating events with empty strings, which could confuse host app developers.

## Test plan

- [x] `./gradlew :sdk:forms:testDebugUnitTest` passes (137 tests, 0 failures)
- [x] `./gradlew ktlintCheck` passes
- [ ] Verify on device that forms still render and dismiss correctly
- [ ] Verify deep link CTA events still fire with proper context

### DIFF
--- a/sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoNativeBridge.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoNativeBridge.kt
@@ -138,7 +138,7 @@ internal class KlaviyoNativeBridge : NativeBridge {
 
         DeepLinking.handleDeepLink(deepLinkUri)
 
-        if (message.formId.isEmpty() || message.formName.isEmpty() || message.buttonLabel.isEmpty()) {
+        if (message.formId.isEmpty() || message.formName.isEmpty()) {
             Registry.log.warning(
                 "OpenDeepLink missing required fields, skipping lifecycle callback"
             )
--- a/sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoNativeBridge.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoNativeBridge.kt
@@ -96,6 +96,14 @@ internal class KlaviyoNativeBridge : NativeBridge {
      */
     private fun show(bridgeMessage: FormWillAppear) {
         Registry.get<PresentationManager>().present()
+
+        if (bridgeMessage.formId.isEmpty() || bridgeMessage.formName.isEmpty()) {
+            Registry.log.warning(
+                "FormWillAppear missing required fields, skipping lifecycle callback"
+            )
+            return
+        }
+
         invokeFormLifecycleHandler(
             FormLifecycleEvent.FormShown(bridgeMessage.formId, bridgeMessage.formName)
         )
@@ -129,6 +137,14 @@ internal class KlaviyoNativeBridge : NativeBridge {
         }
 
         DeepLinking.handleDeepLink(deepLinkUri)
+
+        if (message.formId.isEmpty() || message.formName.isEmpty() || message.buttonLabel.isEmpty()) {
+            Registry.log.warning(
+                "OpenDeepLink missing required fields, skipping lifecycle callback"
+            )
+            return
+        }
+
         invokeFormLifecycleHandler(
             FormLifecycleEvent.FormCtaClicked(
                 formId = message.formId,
@@ -144,6 +160,14 @@ internal class KlaviyoNativeBridge : NativeBridge {
      */
     private fun close(bridgeMessage: FormDisappeared) {
         Registry.get<PresentationManager>().dismiss()
+
+        if (bridgeMessage.formId.isEmpty() || bridgeMessage.formName.isEmpty()) {
+            Registry.log.warning(
+                "FormDisappeared missing required fields, skipping lifecycle callback"
+            )
+            return
+        }
+
         invokeFormLifecycleHandler(
             FormLifecycleEvent.FormDismissed(bridgeMessage.formId, bridgeMessage.formName)
         )
--- a/sdk/forms/src/main/java/com/klaviyo/forms/bridge/NativeBridgeMessage.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/bridge/NativeBridgeMessage.kt
@@ -115,9 +115,20 @@ internal sealed class NativeBridgeMessage {
         }
 
         /**
+         * Require a non-empty string value for the given key, or throw.
+         * The caller ([KlaviyoNativeBridge.postMessage]) catches and logs the exception.
+         */
+        private fun JSONObject.requireString(key: String): String =
+            optString(key).takeIf { it.isNotEmpty() }
+                ?: throw IllegalArgumentException(
+                    "Required field '$key' missing from bridge message"
+                )
+
+        /**
          * Parse a native bridge message string into a [NativeBridgeMessage]
          *
          * @throws IllegalStateException for unexpected message strings
+         * @throws IllegalArgumentException for messages missing required fields
          */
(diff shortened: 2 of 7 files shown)


sdk/forms/src/main/java/com/klaviyo/forms/bridge/NativeBridgeMessage.kt
```diff
@@ -146,14 +157,14 @@ internal sealed class NativeBridgeMessage {
...
+                    formId = jsonData.requireString("formId"),
+                    formName = jsonData.requireString("formName"),
+                    buttonLabel = jsonData.requireString("buttonLabel")
                 )
 
                 keyName<FormDisappeared>() -> FormDisappeared(
-                    formId = jsonData.optString("formId"),
-                    formName = jsonData.optString("formName")
+                    formId = jsonData.requireString("formId"),
+                    formName = jsonData.requireString("formName")
```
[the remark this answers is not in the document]
evan-masseau: Fixed in 32baceee1. `FormDisappeared` now uses `optString` with empty-string defaults instead of `requireString`, so `dismiss()` always fires even if the JS bridge sends malformed data. This prevents the user from getting permanently stuck behind the overlay. Good catch.

sdk/forms/src/test/java/com/klaviyo/forms/bridge/KlaviyoNativeBridgeTest.kt
```diff
@@ -114,23 +114,45 @@ internal class KlaviyoNativeBridgeTest : BaseTest() {
...
-        verify { mockPresentationManager.present() }
+        val message = """{"type":"formWillAppear"}"""
+        postMessage(message)
+        verify(exactly = 0) { mockPresentationManager.present() }
+        verify {
+            spyLog.error(
+                "Failed to relay webview message: $message",
+                any<IllegalArgumentException>()
+            )
+        }
```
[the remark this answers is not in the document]
evan-masseau: Fixed in 32baceee1. All test assertions now verify log level + exception type using `any()` matchers instead of exact message strings, consistent with the project guidelines and the pattern already used in `FormLifecycleHandlerTest`.

sdk/forms/src/main/java/com/klaviyo/forms/bridge/NativeBridgeMessage.kt
```diff
@@ -129,8 +140,8 @@ internal sealed class NativeBridgeMessage {
                 keyName<HandShook>() -> HandShook
 
                 keyName<FormWillAppear>() -> FormWillAppear(
-                    formId = jsonData.optString("formId"),
-                    formName = jsonData.optString("formName")
+                    formId = jsonData.requireString("formId"),
```
evan-masseau: Hm did not consider that this would mean the whole formWillAppear event would get blocked. I think the validation on ID and name etc should only be applied to the lifecycle even broadcast, so I guess it can't go in the json parsing but the LifecycleEvent construction instead

sdk/forms/src/main/java/com/klaviyo/forms/bridge/NativeBridgeMessage.kt
```diff
@@ -129,8 +129,8 @@ internal sealed class NativeBridgeMessage {
                 keyName<HandShook>() -> HandShook
 
                 keyName<FormWillAppear>() -> FormWillAppear(
-                    formId = jsonData.optString("formId"),
-                    formName = jsonData.optString("formName")
+                    formId = jsonData.optString("formId", ""),
```
evan-masseau: you don't need to change this... its already the behavior of optString without a second arg

sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoNativeBridge.kt
```diff
@@ -96,6 +96,14 @@ internal class KlaviyoNativeBridge : NativeBridge {
      */
     private fun show(bridgeMessage: FormWillAppear) {
         Registry.get<PresentationManager>().present()
+
+        if (bridgeMessage.formId.isEmpty() || bridgeMessage.formName.isEmpty()) {
```
evan-masseau, line 100: These are still exceedingly unlikely edge cases, but Dan pointed out that in the event of a failure like this, it probably makes more sense to handle it internally by logging error messages, rather than pass the ambiguity of a form with no ID/Name along to the host app.

sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoNativeBridge.kt
```diff
@@ -129,6 +137,14 @@ internal class KlaviyoNativeBridge : NativeBridge {
         }
 
         DeepLinking.handleDeepLink(deepLinkUri)
+
+        if (message.formId.isEmpty() || message.formName.isEmpty()) {
```
evan-masseau, line 141: Intentionally letting button name be empty, because it isn't technically required in the editor either.

sdk/forms/src/test/java/com/klaviyo/forms/FormLifecycleHandlerTest.kt
```diff
@@ -154,19 +154,20 @@ internal class FormLifecycleHandlerTest : BaseTest() {
...
     }
 
     @Test
-    fun `formDisappeared without data delegates dismiss with empty context fields`() {
+    fun `formDisappeared without data still dismisses`() {
         Klaviyo.registerFormLifecycleHandler(FormLifecycleHandler { _ -> })
 
         nativeBridge.postMessage("""{"type":"formDisappeared"}""")
 
+        // FormDisappeared uses tolerant parsing so dismiss always fires
```
evan-masseau, line 171: So this test should verify that the hook did NOT fire... or if that case is already covered in tests, perhaps this test can just be removed.

evan-masseau: Fixed in c27f7f8. The test was only verifying that `dismiss()` fired but wasn't checking that the lifecycle callback was skipped — so it would pass even if the callback erroneously fired. Now it tracks whether the callback was invoked and asserts it wasn't, plus verifies the warning log. Kept the test rather than removing it since `KlaviyoNativeBridgeTest` covers the same case with a mock handler, but this one exercises a real lambda callback through the full bridge path.

ajaysubra (approved):
LGTM

dan-peluso (approved):
appreciate you making the changes here

sdk/forms/src/test/java/com/klaviyo/forms/FormLifecycleHandlerTest.kt
```diff
@@ -154,20 +154,25 @@ internal class FormLifecycleHandlerTest : BaseTest() {
...
+    fun `formDisappeared without data still dismisses but skips lifecycle callback`() {
+        var callbackInvoked = false
+        Klaviyo.registerFormLifecycleHandler(FormLifecycleHandler { _ -> callbackInvoked = true })
 
         nativeBridge.postMessage("""{"type":"formDisappeared"}""")
 
+        // FormDisappeared uses tolerant parsing so dismiss always fires
         verify { mockPresentationManager.dismiss() }
+        // Missing formId/formName means the lifecycle callback should be skipped
+        assert(!callbackInvoked) { "Lifecycle callback should not fire when form metadata is missing" }
```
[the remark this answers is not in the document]
evan-masseau: well this simply can't be true

PR #442, opened 2026-04-15
fix(push): dismiss notification on action button tap

evan-masseau:
## Summary
- `setAutoCancel(true)` only dismisses notifications on body tap, not action button taps (standard Android behavior)
- Action button intents now carry the notification tag as an extra (`NOTIFICATION_TAG_EXTRA`)
- `handlePush()` cancels the notification via `NotificationManagerCompat` when it sees this extra
- Notification tag is computed once in `displayNotification` and threaded through to `buildNotification` → `addActionButtons` → `createButtonAction`

## Test plan
- [ ] Existing tests pass (2 updated for new `buildNotification` signature)
- [ ] New test: action button intent includes notification tag from payload
- [ ] New test: action button intent includes generated tag when payload has no tag
- [ ] Manual: send push with action buttons → tap button → notification dismisses
- [ ] Manual: send push without action buttons → tap body → still auto-cancels (no regression)

### DIFF
--- a/sdk/core/src/main/java/com/klaviyo/core/Constants.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/Constants.kt
@@ -18,4 +18,10 @@ object Constants {
      * Klaviyo push messages contain metadata to associate an event with its original transmission
      */
     const val TRACKING_PARAMETER = "_k"
+
+    /**
+     * Intent extra key for the notification tag, used to dismiss the notification
+     * when an action button is tapped and [handlePush] processes the intent
+     */
+    const val NOTIFICATION_TAG_EXTRA = PACKAGE_PREFIX + "notification_tag"
 }
--- a/sdk/core/src/main/java/com/klaviyo/core/Constants.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/Constants.kt
@@ -24,4 +24,10 @@ object Constants {
      * when an action button is tapped and [handlePush] processes the intent
      */
     const val NOTIFICATION_TAG_EXTRA = PACKAGE_PREFIX + "notification_tag"
+
+    /**
+     * Fixed notification ID used in all notify/cancel calls.
+     * Notifications are uniquely identified by their string tag, not this ID.
+     */
+    const val NOTIFICATION_ID = 0
 }
--- a/sdk/core/src/main/java/com/klaviyo/core/Constants.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/Constants.kt
@@ -21,9 +21,13 @@ object Constants {
 
     /**
      * Intent extra key for the notification tag, used to dismiss the notification
-     * when an action button is tapped and [handlePush] processes the intent
+     * when an action button is tapped and [handlePush] processes the intent.
+     *
+     * Uses [INTERNAL_PREFIX] instead of [PACKAGE_PREFIX] to avoid being swept into
+     * analytics event properties by [appendKlaviyoExtras].
      */
-    const val NOTIFICATION_TAG_EXTRA = PACKAGE_PREFIX + "notification_tag"
+    private const val INTERNAL_PREFIX = "_klaviyo."
+    const val NOTIFICATION_TAG_EXTRA = INTERNAL_PREFIX + "notification_tag"
 
     /**
      * Fixed notification ID used in all notify/cancel calls.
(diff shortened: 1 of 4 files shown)


sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
```diff
@@ -310,6 +316,7 @@ class KlaviyoNotification(private val message: RemoteMessage) {
             }
         }?.apply {
             addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+            putExtra(Constants.NOTIFICATION_TAG_EXTRA, notificationTag)
```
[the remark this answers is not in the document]
evan-masseau: Yeah this sounds valid... we could make use a different prefix for the tag? else exclude it explicitly i guess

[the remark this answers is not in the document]
evan-masseau: Fixed in 5a4620b — changed the prefix from `com.klaviyo.` to `_klaviyo.` so `appendKlaviyoExtras` won't sweep it into event properties.

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
```diff
@@ -153,7 +159,10 @@ class KlaviyoNotification(private val message: RemoteMessage) {
      * @param context
      * @return [Notification.Builder] to display
      */
-    internal fun buildNotification(context: Context): NotificationCompat.Builder {
+    internal fun buildNotification(
+        context: Context,
+        notificationTag: String = message.notificationTag ?: generateId().toString()
```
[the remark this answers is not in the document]
evan-masseau: Fixed in 83eb2d0 — removed the default parameter entirely. `buildNotification` now requires the tag, so the only path is through `displayNotification` which computes it once.

sdk/core/src/main/java/com/klaviyo/core/Constants.kt
```diff
@@ -18,4 +18,20 @@ object Constants {
...
+
+    /**
+     * Intent extra key for the notification tag, used to dismiss the notification
+     * when an action button is tapped and [handlePush] processes the intent.
+     *
+     * Uses [INTERNAL_PREFIX] instead of [PACKAGE_PREFIX] to avoid being swept into
+     * analytics event properties by [appendKlaviyoExtras].
+     */
+    private const val INTERNAL_PREFIX = "_klaviyo."
+    const val NOTIFICATION_TAG_EXTRA = INTERNAL_PREFIX + "notification_tag"
```
dan-peluso, line 30: any reason we're keeping this generic and not something more clearly tied to 'action buttons'? It seems like this is just used to distinguish those from normal pushes, but are we trying to future proof it for differing types of non-standard push notifications? I'm thinking things like live notifications, etc

evan-masseau: The key specifically? I made a separate key for it from `com.klaviyo.*` so that it wouldn't get lumped into the the `opened_push` payload, since its only required for local purposes. As for keeping it a generic prefix vs action-button specific... Generic makes sense to me, `notification_tag` is a property of the whole notification, so if we needed to uniquely identify live notifications with a tag, for example, we could still just use this key for that.

dan-peluso: Yeah I guess the value for the key - calling it `notification_tag` versus `action_button_push_tag` or something like that. But if we're planning to use this same pattern for updating other types of pushes (do we have any on the roadmap?) I think this makes sense, as long we we're cool with those also auto-dismissing on open

which actually, come to think of it, might be something we _dont_ want when doing a live notification (the expectation being, it stays in the tray until the 'thing' we track ends) but I might be getting ahead of myself here

evan-masseau: Maybe a little ahead of yourself 😁 
But also, this is not an individual tag per action button, it is the notification's tag, so wouldn't want it to be misleading.

dan-peluso: Right, maybe I'm misunderstanding this though - this is how we distinguish action button style notifications from non-action button style notifications right?

dan-peluso (approved):

sdk/core/src/main/java/com/klaviyo/core/Constants.kt
evan-masseau: No, a notification always has a tag, its just a unique identifier. Here we are passing the tag on the action buttons' intent extras so that when the intent is sent back to the app, we know what notification it came from, and can dismiss that notification from the system tray. This is required for action buttons because they don't dismiss a notification by default, and don't have an option built in to do so either. Unlike the main pending intent of a notification, which has an `autoCancel` property, which if set to true, dismisses the notification immediately on tap, without needing to do so via the application code that receives the intent.

PR #443, opened 2026-04-23
fix(forms): floating form offset clamping + data-klaviyo-device device info

evan-masseau:
# Description

Floating-form rendering tweaks and a new `data-klaviyo-device` attribute on the onsite `<head>` so onsite JS can read device dimensions / safe-area insets / orientation / DPR during synchronous HTML parse, before the webview attaches. Plus a wire-contract rename (`margin` → `offsets`) and a new `addSafeAreaInsetsToOffsets` flag that lets onsite opt out of SDK-managed safe-area math entirely.

## Due Diligence

### Offset clamping
- Clamp floating-form width/height so `offsets` can't drive the computed size negative or push the form off-screen.
- Clamp the bottom-gap used for keyboard-overlap math to `>= 0` so we don't spuriously report overlap when the form sits above the keyboard.

### `data-klaviyo-device` attribute
- New `DeviceInfo` model (`screen`, `safeAreaInsets`, `orientation`, `dpr`) shaped to CSSOM `screen.*` conventions.
- Baked into the HTML template at initial load and pushed via `evaluateJavascript` on orientation / inset changes, so onsite code can read it synchronously before the webview attaches.
- Injection sidesteps JS string-escape concerns by emitting the raw JSON as a JS object expression wrapped in `JSON.stringify(...)` — the engine parses it natively.

### Rotation-lag fix
- `pushDeviceInfo()` now defers to the **next** `ActivityEvent.Resumed` after `onConfigurationChanged`, rather than firing on the config-change callback itself. `ConfigurationChanged` fires while the old activity / display is still being torn down, so reading `Display.rotation` / `decorView.rootWindowInsets` at that moment yields stale values one rotation behind. Uses a one-shot observer with a 1s safety timeout.

### Wire rename + safe-area opt-out
- **`margin` → `offsets`** in the `formWillAppear.layout` payload. The legacy `margin` fallback has been removed now that onsite is fully migrated in production.
- **`addSafeAreaInsetsToOffsets: Boolean`** (optional, defaults to `true`). When `false`, the SDK uses `offsets` as absolute distances from the screen edge — no safe-area inflation in position math, available-space clamping, or keyboard-gap math. Onsite takes responsibility for safe-area policy.

### Misc hardening
- `DeviceInfoProvider.current()` reads UI-thread-only APIs; entire `pushDeviceInfo` chain now dispatches through `runOnUiThread` to avoid silent off-thread degradation.
- `Activity.display` on API 30+ (falling back to `windowManager.defaultDisplay` on 23-29) — avoids blanket `@Suppress("DEPRECATION")` and correctly scopes to the activity's actual display in multi-display setups.
- `pushDeviceInfo` is silent on a null webview (expected idle state during preload / between register-unregister cycles).
- Orientation mapping documented with a KDoc note on the natural-landscape-device caveat (our product scope is natural-portrait phones).

## Wire-contract changes for Fender coordination
- **`margin` → `offsets`** on `formWillAppear.layout`. No backward-compat fallback.
- **New `addSafeAreaInsetsToOffsets: Boolean`** on `formWillAppear.layout`, defaults to `true`.
- **`<head data-klaviyo-device='{…}'>`** attribute with CSSOM-shaped JSON; updated on rotation and inset changes.

## Test Plan

- [x] `./gradlew :sdk:forms:testDebugUnitTest` passes (new coverage in `FormLayoutTest`, `FloatingFormWindowTest`, `DeviceInfoTest`, `KlaviyoPresentationManagerTest`).
- [x] `./gradlew :sdk:forms:ktlintCheck` clean.
- [ ] Manual device verification:
  - [ ] Floating form with large `offsets` clamps to screen bounds instead of overflowing.
  - [ ] `data-klaviyo-device` attribute appears on `<head>` at load and refreshes after rotation (verify via `chrome://inspect` web inspector).
  - [ ] Rotation-lag — rotate twice in rapid succession; the attribute's `orientation` always reflects the final device orientation after the second rotation settles.
  - [ ] `addSafeAreaInsetsToOffsets: false` payload from Fender positions the form at absolute edge-relative offsets (no safe-area padding added).
  - [ ] Fullscreen form still fills the screen regardless of the flag.

## Related Issues/Tickets

Related to [MAGE-541](https://linear.app/klaviyo/issue/MAGE-541)

### DIFF
--- a/sdk/forms/src/main/java/com/klaviyo/forms/bridge/DeviceInfo.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/bridge/DeviceInfo.kt
@@ -70,6 +70,13 @@ internal data class DeviceInfo(
              * Android's rotation values describe the counter-clockwise rotation applied to the
              * natural orientation; pairing that with whether the logical orientation is portrait
              * or landscape gives us enough to pick the `*-primary` vs `*-secondary` flavor.
+             *
+             * Caveat: this mapping assumes a natural-portrait device, which holds for all phones
+             * and is sufficient for our product scope. On natural-landscape devices (some tablets
+             * and foldables) the rotation-to-CSSOM mapping may diverge from
+             * [`screen.orientation.type`](https://drafts.csswg.org/screen-orientation/#dom-screenorientation-type):
+             * for example, `rotation = 90` on a natural-landscape device is reported here as
+             * `portrait-secondary`, where the web platform would report `portrait-primary`.
              */
             fun from(configOrientation: Int, rotation: Int): Orientation {
                 val isPortrait = configOrientation != Configuration.ORIENTATION_LANDSCAPE
--- a/sdk/forms/src/main/java/com/klaviyo/forms/bridge/DeviceInfo.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/bridge/DeviceInfo.kt
@@ -1,0 +1,184 @@
+package com.klaviyo.forms.bridge
+
+import android.content.res.Configuration
+import android.content.res.Resources
+import android.util.DisplayMetrics
+import android.view.Surface
+import androidx.core.view.WindowInsetsCompat
+import androidx.core.view.WindowInsetsCompat.Type.displayCutout
+import androidx.core.view.WindowInsetsCompat.Type.systemBars
+import com.klaviyo.core.Registry
+import kotlin.math.roundToInt
+import org.json.JSONObject
+
+/**
+ * Describes the device's current physical display characteristics, exposed to onsite JS
+ * via the `data-klaviyo-device` attribute on the HTML `<head>` element.
+ *
+ * The shape intentionally mirrors CSSOM conventions so onsite code can treat the payload
+ * as a reliable, orientation-aware substitute for `window.screen.*` during the synchronous
+ * HTML parse phase, before the webview attaches to the view hierarchy.
+ */
+internal data class DeviceInfo(
+    val screenWidthDp: Int,
+    val screenHeightDp: Int,
+    val insetTopDp: Int,
+    val insetBottomDp: Int,
+    val insetLeftDp: Int,
+    val insetRightDp: Int,
+    val orientation: Orientation,
+    val dpr: Int
+) {
+    /**
+     * Serializes the device info into its JSON representation as published on the
+     * `data-klaviyo-device` head attribute.
+     */
+    fun toJson(): String = JSONObject()
+        .put(
+            KEY_SCREEN,
+            JSONObject()
+                .put(KEY_WIDTH, screenWidthDp)
+                .put(KEY_HEIGHT, screenHeightDp)
+        )
+        .put(
+            KEY_SAFE_AREA_INSETS,
+            JSONObject()
+                .put(KEY_TOP, insetTopDp)
+                .put(KEY_BOTTOM, insetBottomDp)
+                .put(KEY_LEFT, insetLeftDp)
+                .put(KEY_RIGHT, insetRightDp)
+        )
+        .put(KEY_ORIENTATION, orientation.cssValue)
+        .put(KEY_DPR, dpr)
+        .toString()
+
+    /**
+     * CSSOM `ScreenOrientation.type` vocabulary.
+     * See https://drafts.csswg.org/screen-orientation/#enumdef-orientationtype
+     */
+    internal enum class Orientation(val cssValue: String) {
+        PortraitPrimary("portrait-primary"),
+        PortraitSecondary("portrait-secondary"),
+        LandscapePrimary("landscape-primary"),
+        LandscapeSecondary("landscape-secondary");
+
+        companion object {
+            /**
+             * Derives the CSSOM orientation label from the Android [Configuration.orientation]
+             * and [android.view.Display.getRotation] values.
+             *
+             * Android's rotation values describe the counter-clockwise rotation applied to the
+             * natural orientation; pairing that with whether the logical orientation is portrait
+             * or landscape gives us enough to pick the `*-primary` vs `*-secondary` flavor.
+             */
+            fun from(configOrientation: Int, rotation: Int): Orientation {
+                val isPortrait = configOrientation != Configuration.ORIENTATION_LANDSCAPE
+                return when (rotation) {
+                    Surface.ROTATION_0 -> if (isPortrait) PortraitPrimary else LandscapePrimary
+                    Surface.ROTATION_90 -> if (isPortrait) PortraitSecondary else LandscapePrimary
+                    Surface.ROTATION_180 -> if (isPortrait) PortraitSecondary else LandscapeSecondary
+                    Surface.ROTATION_270 -> if (isPortrait) PortraitPrimary else LandscapeSecondary
+                    else -> if (isPortrait) PortraitPrimary else LandscapePrimary
+                }
+            }
+        }
+    }
+
+    companion object {
+        private const val KEY_SCREEN = "screen"
+        private const val KEY_WIDTH = "width"
+        private const val KEY_HEIGHT = "height"
+        private const val KEY_SAFE_AREA_INSETS = "safeAreaInsets"
+        private const val KEY_TOP = "top"
+        private const val KEY_BOTTOM = "bottom"
+        private const val KEY_LEFT = "left"
+        private const val KEY_RIGHT = "right"
+        private const val KEY_ORIENTATION = "orientation"
+        private const val KEY_DPR = "dpr"
+
+        /**
+         * Build a [DeviceInfo] snapshot from the given display metrics and insets.
+         *
+         * Separating computation from Android platform lookups keeps the logic pure and
+         * testable. See [DeviceInfoProvider] for the live-device lookup entry point.
+         */
+        fun from(
+            displayMetrics: DisplayMetrics,
+            configuration: Configuration,
+            rotation: Int,
+            insetLeftPx: Int,
+            insetTopPx: Int,
+            insetRightPx: Int,
+            insetBottomPx: Int
+        ): DeviceInfo {
+            // Guard against pathological DisplayMetrics (density <= 0) which would cause
+            // NaN from pxToDp. This most commonly occurs in test doubles.
+            val safeDensity = displayMetrics.density.takeIf { it > 0f } ?: 1f
+            fun pxToDpRounded(px: Int): Int = (px / safeDensity).roundToInt()
+            return DeviceInfo(
+                screenWidthDp = pxToDpRounded(displayMetrics.widthPixels),
+                screenHeightDp = pxToDpRounded(displayMetrics.heightPixels),
+                insetTopDp = pxToDpRounded(insetTopPx),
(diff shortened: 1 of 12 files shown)


evan-masseau:
bugbot run

evan-masseau:
bugbot run

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoPresentationManager.kt
```diff
@@ -420,6 +441,49 @@ internal class KlaviyoPresentationManager() : PresentationManager {
         hostActivityStoppedCleanup = null
     }
 
+    /**
```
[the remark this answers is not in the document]
evan-masseau: Intentional — docs clarified in commit [cb57accfd](https://github.com/klaviyo/klaviyo-android-sdk/pull/443/commits/cb57accfd).

The observer deliberately survives `clearTimers()` because the `data-klaviyo-device` attribute must stay current on the preloaded webview whether or not a form is presenting. Key points:

- `dismiss()` does NOT destroy the webview — it only closes the overlay/floating window. The webview itself is preloaded and persists across presentations. The only teardown path is `destroyWebviewAndListeners` (abort / session timeout).
- `pushDeviceInfo()` is already null-safe: if the webview is gone, it logs at verbose and no-ops. No risk of pushing to "torn-down" state.
- The whole point of deferring to the next Resumed is to keep the attribute fresh on the idle preloaded webview ready for the next form. Cancelling it in `clearTimers` would reintroduce the rotation-lag bug we just fixed.

Resolving.

evan-masseau:
bugbot run

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoPresentationManager.kt
```diff
@@ -428,5 +499,13 @@ internal class KlaviyoPresentationManager() : PresentationManager {
...
          */
         private const val CLOSE_TIMEOUT = 600L
+
+        /**
+         * Upper bound for how long to wait for the next Resumed activity before giving up
+         * on pushing fresh device info after a rotation. Typical rotation-to-Resumed is
+         * well under 300 ms; this bounds the observer lifetime if the activity is
+         * permanently destroyed without replacement.
+         */
+        private const val DEVICE_INFO_PUSH_TIMEOUT = 1_000L
```
evan-masseau, line 509: also to be clear: this isn't any kind of user-visible delay

evan-masseau:
bugbot run

sdk/forms/src/main/java/com/klaviyo/forms/bridge/FormLayout.kt
```diff
@@ -156,19 +155,42 @@ internal data class FormLayout(
...
+
         fun fromJson(json: JSONObject?): FormLayout? {
             if (json == null) return null
 
             val position = FormPosition.fromString(json.optString("position"))
             val width = Dimension.fromJson(json.optJSONObject("width")) ?: return null
             val height = Dimension.fromJson(json.optJSONObject("height")) ?: return null
-            val offsets = Offsets.fromJson(json.optJSONObject("margin"))
+
+            // Prefer the new `offsets` wire key; fall back to legacy `margin` for
```
evan-masseau: we don't need to check for the old key at all anymore, its gone

sdk/forms/src/main/java/com/klaviyo/forms/bridge/FormLayout.kt
```diff
@@ -156,19 +155,42 @@ internal data class FormLayout(
...
+            // backward compatibility with older onsite payloads. Log once per session
+            // when we hit the fallback so we can track deprecation in the wild.
+            val offsetsJson = json.optJSONObject("offsets") ?: json.optJSONObject("margin")?.also {
+                if (loggedMarginDeprecation.compareAndSet(false, true)) {
+                    Registry.log.verbose(
+                        "FormLayout payload used deprecated `margin` key; " +
+                            "expected `offsets`. Update onsite to emit `offsets`."
+                    )
+                }
```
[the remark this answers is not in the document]
evan-masseau: we don't need to reference margin at all, i'm taking this out.

sdk/forms/src/main/java/com/klaviyo/forms/bridge/DeviceInfo.kt
```diff
@@ -0,0 +1,186 @@
...
+                    Surface.ROTATION_180 -> if (isPortrait) PortraitSecondary else LandscapeSecondary
+                    Surface.ROTATION_270 -> if (isPortrait) PortraitPrimary else LandscapeSecondary
+                    else -> if (isPortrait) PortraitPrimary else LandscapePrimary
+                }
+            }
+        }
+    }
+
+    companion object {
+        private const val KEY_SCREEN = "screen"
```
dan-peluso, line 95: just being extra cautious but no chance these would have to accomodate i18n here right? Like if certain web regions would be at risk with these english translations

evan-masseau: no, these are json payload keys

sdk/forms/src/main/java/com/klaviyo/forms/presentation/FloatingFormWindow.kt
```diff
@@ -403,95 +414,141 @@ internal class FloatingFormWindow(private val context: Context) {
...
+            // offsets are treated as the absolute distance from the screen edge and the
+            // SDK does not touch safe-area at all (position math and clamping both skip
+            // the safe-area term). Default (true) preserves the historical behavior of
+            // SDK-managed safe-area insets.
+            val effectiveSafeTop = if (layout.addSafeAreaInsetsToOffsets) safeAreaTop else 0
+            val effectiveSafeBottom = if (layout.addSafeAreaInsetsToOffsets) safeAreaBottom else 0
+            val effectiveSafeLeft = if (layout.addSafeAreaInsetsToOffsets) safeAreaLeft else 0
+            val effectiveSafeRight = if (layout.addSafeAreaInsetsToOffsets) safeAreaRight else 0
+
+            val marginTop = effectiveSafeTop + (layout.offsets.top * density).toInt()
```
evan-masseau: urgh, missed variable renaming again, shuolda been `offset*`

sdk/forms/src/main/java/com/klaviyo/forms/bridge/DeviceInfo.kt
```diff
@@ -0,0 +1,186 @@
...
+ * HTML parse phase, before the webview attaches to the view hierarchy.
+ */
+internal data class DeviceInfo(
+    val screenWidthDp: Int,
+    val screenHeightDp: Int,
+    val insetTopDp: Int,
+    val insetBottomDp: Int,
+    val insetLeftDp: Int,
+    val insetRightDp: Int,
+    val orientation: Orientation,
```
dan-peluso, line 29: we only care about this on the first load right? I can probably see later on in the logic but just wondering how actively this gets set

sdk/forms/src/main/java/com/klaviyo/forms/bridge/DeviceInfo.kt
```diff
@@ -0,0 +1,186 @@
...
+ */
+internal object DeviceInfoProvider {
+
+    /**
+     * Snapshot the current device state. Prefers values from the tracked activity so that
+     * safe-area insets and rotation reflect the actual window placement rather than the raw
+     * display.
+     */
+    fun current(): DeviceInfo {
+        val activity = Registry.lifecycleMonitor.currentActivity
```
dan-peluso, line 154: Just to understand the timing of this better, is this the activity that we launch the webview from or activity that contains the webview?

sdk/forms/src/main/java/com/klaviyo/forms/presentation/FloatingFormWindow.kt
```diff
@@ -403,95 +414,141 @@ internal class FloatingFormWindow(private val context: Context) {
...
-
-                FormPosition.TOP,
-                FormPosition.TOP_LEFT,
-                FormPosition.TOP_RIGHT -> screenHeight - safeAreaTop - topOffset - formHeight
-
-                // TODO: If center-aligned forms support top/bottom offsets in the future,
-                //  this calculation should account for the offset shifting the form from center
-                FormPosition.CENTER -> (screenHeight - formHeight) / 2
+            safeAreaBottom: Int,
+            safeAreaLeft: Int,
```
dan-peluso, line 451: super super super nit but if we do ever come to support RTL for certain countries sometimes folks use 'start' and 'end' instead of left right. Really only applies more for text (I think) and I'm not sure if it would impact where a dialog is offset on a screen but the ole android dev brain had an alarm go off and figure it's worth at least mentioning. Definitely won't be offended if this doesn't switch and likely thinking way too far ahead but I can't help myself in calling it out lol

dan-peluso (approved):
Admittedly I don't have the best understanding of the bug, but I think this approach is generally similar to our other listeners / setters. Mostly just questions, but I did load this into AS and click around to get a better look. One high level thing I was wondering is do we up the handshake contract or is it not necessary since we're only adding an HTML tag, and not modifying the message bus schema. Is this easy to make backwards compatible on the fender side?

belleklaviyo (approved):
from a testing standpoint this seems good and in parity with iOS now! loaded forms side by side on real devices

sdk/forms/src/main/java/com/klaviyo/forms/bridge/DeviceInfo.kt
evan-masseau: the dimension data is updated on rotation as well

sdk/forms/src/main/java/com/klaviyo/forms/presentation/FloatingFormWindow.kt
evan-masseau: yeah i did consider it, after having done enough compose work in the past, just felt like added complexity and we can compensate for it in the future if necessary. Felt like I should stick to same terminology as the editor, onsite etc.

sdk/forms/src/main/java/com/klaviyo/forms/bridge/DeviceInfo.kt
evan-masseau: This would generally be the host app's current activity. doesn't really matter though, we're using it to access the _device_ dimensions

evan-masseau:
> Admittedly I don't have the best understanding of the bug, but I think this approach is generally similar to our other listeners / setters. Mostly just questions, but I did load this into AS and click around to get a better look. One high level thing I was wondering is do we up the handshake contract or is it not necessary since we're only adding an HTML tag, and not modifying the message bus schema. Is this easy to make backwards compatible on the fender side?

I didn't add a handshake spec entry, debatable choice, but i figured the structure is pretty self evident, and we don't really need it for feature-detection sake because you can just react to whether the data attribute is there or not. If we do have a breaking change in the future, we could always add a handshake spec item then.

Most of the other data attributes that are just like, environment vars, we don't explicitly document as handshake spec items, e.g. `forms-environment` or the SDK name/version properties for example.

PR #445, opened 2026-05-04
feat(forms): JWT delivery to IAF WebView via evaluateJavascript

dan-peluso:
# Description

Adds `Klaviyo.setJWT(token)` (and `KlaviyoForms.setJWT(token)` for Java callers) to deliver a JWT to the active In-App Forms WebView by calling `window.klaviyoIAFSetJWT(token)` via `evaluateJavascript`. Implements the Android counterpart to the iOS JWT delivery ticket (MAGE-555).

The call is order-independent: the token is queued at whichever layer it arrives and delivered as soon as the WebView is ready.

| When `setJWT` is called | Behaviour |
|---|---|
| Before `registerForInAppForms` | Queued in `KlaviyoFormsProvider`; forwarded to `WebViewClient` on register |
| After register, before JS handshake | Queued in `KlaviyoWebViewClient`; delivered on handshake completion |
| After handshake | Delivered immediately via `evaluateJavascript` |

JWT delivery is flushed **before** `ProfileEventObserver` starts sending buffered events on handshake, giving onsite a head start on profile fetch before form evaluation is triggered.

## Due Diligence

## Changelog / Code Overview

**New public API (`forms-core`)**
- `fun Klaviyo.setJWT(token: String): Klaviyo` — Kotlin extension
- `KlaviyoForms.setJWT(token: String)` — `@JvmStatic` for Java callers
- `FormsProvider.setJWT(token: String)` added to the service interface

**Internal (`forms`)**
- `WebViewClient.setJWT(token: String)` added to the internal interface
- `KlaviyoWebViewClient`: implements `setJWT` with single-quote escaping and a `pendingJwt` / `isHandshakeCompleted` queue that flushes on `onJsHandshakeCompleted` — JWT is enqueued before `ProfileEventObserver` fires
- `KlaviyoFormsProvider`: implements `setJWT` with a `pendingJwt` field that queues the token when `WebViewClient` is not yet registered, forwarding it into `WebViewClient` during `register()`

**Tests**
- `KlaviyoWebViewClientTest`: ready-webview delivery + pre-handshake queuing
- `InAppFormsTest`: pre-register queuing + post-register delegation
- `InAppFormsNoOpTest`: throws `MissingKlaviyoModule` when forms module absent

## Test Plan
- Call `Klaviyo.setJWT(token)` before `registerForInAppForms()` — verify delivery occurs once the webview handshake completes
- Call `Klaviyo.setJWT(token)` after `registerForInAppForms()` but before the webview has finished loading — verify delivery occurs on handshake
- Call `Klaviyo.setJWT(token)` after handshake — verify `window.klaviyoIAFSetJWT` is called immediately
- Verify tokens containing backslashes or single quotes are correctly escaped in the JS string

## Related Issues/Tickets

- Linear: [MAGE-555](https://linear.app/klaviyo/issue/MAGE-555/jwt-delivery-to-webview-via-evaluatejavascript-android)
- iOS counterpart: MAGE-554
- JS counterpart (defines `window.klaviyoIAFSetJWT`): MAGE-553

### DIFF
--- a/sdk/forms-core/src/main/java/com/klaviyo/forms/FormsProvider.kt
+++ b/sdk/forms-core/src/main/java/com/klaviyo/forms/FormsProvider.kt
@@ -19,4 +19,13 @@ interface FormsProvider {
      * and tearing down the WebView client.
      */
     fun unregister()
+
+    /**
+     * Deliver a JWT to the active IAF webview by calling `window.klaviyoIAFSetJWT`.
+     * If the webview bridge is not yet ready, the token will be queued and delivered once the
+     * handshake completes.
+     *
+     * @param token The JWT to deliver to the webview.
+     */
+    fun setJWT(token: String)
 }
--- a/sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoFormsProvider.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoFormsProvider.kt
@@ -32,6 +32,12 @@ internal fun Klaviyo.reInitializeInAppForms(): Klaviyo {
 }
 
 internal class KlaviyoFormsProvider : FormsProvider {
+
+    /**
+     * JWT queued before [registerForInAppForms] was called; forwarded to [WebViewClient] on register.
+     */
+    private var pendingJwt: String? = null
+
     override fun register(config: InAppFormsConfig) {
         Registry.apply {
             register<InAppFormsConfig>(config)
@@ -44,6 +50,10 @@ internal class KlaviyoFormsProvider : FormsProvider {
             registerOnce<JsBridgeObserverCollection> { KlaviyoObserverCollection() }
         }
         Registry.get<WebViewClient>().initializeWebView()
+        pendingJwt?.let { token ->
+            pendingJwt = null
+            Registry.get<WebViewClient>().setJWT(token)
+        }
     }
 
     override fun unregister() {
@@ -52,6 +62,13 @@ internal class KlaviyoFormsProvider : FormsProvider {
             Registry.get<WebViewClient>().destroyWebView()
         } else {
             Registry.log.warning("Cannot unregister In-App Forms, must be registered first.")
+        }
+    }
+
+    override fun setJWT(token: String) {
+        Registry.getOrNull<WebViewClient>()?.setJWT(token) ?: run {
+            Registry.log.verbose("IAF not yet registered, queuing JWT for delivery on register")
+            pendingJwt = token
         }
     }
 }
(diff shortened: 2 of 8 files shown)


sdk/forms-core/src/main/java/com/klaviyo/forms/FormsProvider.kt
```diff
@@ -19,4 +19,13 @@ interface FormsProvider {
...
     fun unregister()
+
+    /**
+     * Deliver a JWT to the active IAF webview by calling `window.klaviyoIAFSetJWT`.
+     * If the webview bridge is not yet ready, the token will be queued and delivered once the
+     * handshake completes.
+     *
+     * @param token The JWT to deliver to the webview.
+     */
+    fun setJWT(token: String)
```
evan-masseau, line 30: Memory like a sieve... did we say this would live in `forms` or `analytics`? Thought we said `analytics` (and `KlaviyoSwift` on iOS), and then have `forms` observe state to be notified when a JWT token gets set.

sdk/forms/src/main/java/com/klaviyo/forms/KlaviyoFormsProvider.kt
```diff
@@ -32,6 +32,12 @@ internal fun Klaviyo.reInitializeInAppForms(): Klaviyo {
 }
 
 internal class KlaviyoFormsProvider : FormsProvider {
+
+    /**
+     * JWT queued before [registerForInAppForms] was called; forwarded to [WebViewClient] on register.
+     */
+    private var pendingJwt: String? = null
```
evan-masseau, line 39: If moved to `analytics` you could put this in the `KlaviyoState` object as an observable property

ajaysubra (commented):
Just adding a comment for the bot to stop pining me but looks like evan is looking into it.

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
```diff
@@ -132,11 +148,32 @@ internal class KlaviyoWebViewClient() : AndroidWebViewClient(), WebViewClient, J
         }
     }
 
+    /**
+     * Deliver a JWT to the webview, or queue it if the bridge handshake has not yet completed.
+     */
+    override fun setJWT(token: String) {
+        val escapedToken = token.replace("\\", "\\\\").replace("'", "\\'")
```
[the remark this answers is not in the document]
evan-masseau: Can a JWT even contain any unsafe characters? Maybe instead of this, we should just validate the shape of the JWT that was passed in, at which point, we know it doesn't even need escaping here. https://www.jwt.io/introduction#what-is-json-web-token-structure 
Seems like even a simple regex could validate the general shape of the JWT, which would be sufficient to know that we won't need any escaping here.

evan-masseau (changes requested):
Left some thoughts, also gonna "request changes" so we don't mistakenly merge this to master even when approved. Plz start a `feat/*` branch off of master first and target that for now.

sdk/forms-core/src/main/java/com/klaviyo/forms/FormsProvider.kt
dan-peluso: Yeah this needs to go in analytics for sure

dan-peluso:
Closing this after today's discussion - need some design updates here

PR #449, opened 2026-05-20
feat(analytics): JWT parser and validator for auth token system

amber-klaviyo:
## Summary
Foundational layer for personalized in-app-forms auth-token system. Adds a pure-Kotlin JWT parser that extracts `exp` and `iat` (RFC 7519 NumericDate) claims, rejects malformed/expired tokens with 15-second clock-skew leeway, and returns a `ValidatedToken` the downstream `AuthTokenManager` can use for refresh scheduling.

## Test plan
- ✅ 25 test cases covering: happy path (valid token, unknown claims, fractional NumericDate), malformed structure (empty, 1/2/4 segments), malformed Base64, malformed JSON (non-JSON, array, exp as string, empty segment), missing claims (exp/iat/both), expiration boundaries (now==exp, now>exp, leeway boundary), Base64URL handling (URL-safe chars, no padding)
- ✅ Android-specific additions: `ValidatedToken.toString()` redaction, logging never leaks token contents
- ✅ Full analytics test suite passes; no regressions
- ✅ ktlint clean

## Notes
- No signature verification — the Klaviyo backend is the security boundary
- Logging via `Registry.log.warning(...)` with literal messages; token contents are **never** logged
- `ValidatedToken` has a defensive `toString()` override to prevent accidental JWT leakage if logged
- Base64URL decoding normalizes to standard base64 (matches Swift ref impl, avoids test fixture modification)
- All types are `internal` — no public API surface change in this ticket

## Related
- Reference: [Swift PR 575](https://github.com/klaviyo/klaviyo-swift-sdk/pull/575) (merged)
- Blocks: [MAGE-618](https://linear.app/klaviyo/issue/MAGE-618) (AuthTokenManager + public setAuthTokenProvider API)
- Plan: https://linear.app/storage/files/JWT_MAGE_617_Plan

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/auth/JWTParser.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/auth/JWTParser.kt
@@ -6,7 +6,7 @@ import org.json.JSONException
 import org.json.JSONObject
 
 internal object JWTParser {
-    const val DEFAULT_LEEWAY_SECONDS: Long = 15L
+    const val DEFAULT_LEEWAY_SECONDS: Long = 30L
 
     fun parseAndValidate(
         token: String,
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/auth/JWTParser.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/auth/JWTParser.kt
@@ -1,0 +1,89 @@
+package com.klaviyo.analytics.auth
+
+import android.util.Base64
+import com.klaviyo.core.Registry
+import org.json.JSONException
+import org.json.JSONObject
+
+internal object JWTParser {
+    const val DEFAULT_LEEWAY_SECONDS: Long = 15L
+
+    fun parseAndValidate(
+        token: String,
+        nowEpochSeconds: Long = Registry.clock.currentTimeMillis() / 1000L,
+        leewaySeconds: Long = DEFAULT_LEEWAY_SECONDS
+    ): JWTValidationResult {
+        val segments = token.split('.')
+        if (segments.size != 3) {
+            Registry.log.warning("JWT validation failed: malformed structure")
+            return JWTValidationResult.MalformedStructure
+        }
+
+        val payloadBytes = base64UrlDecode(segments[1]) ?: run {
+            Registry.log.warning("JWT validation failed: malformed base64URL payload")
+            return JWTValidationResult.MalformedBase64
+        }
+
+        val claims = try {
+            JSONObject(String(payloadBytes, Charsets.UTF_8))
+        } catch (e: JSONException) {
+            Registry.log.warning("JWT validation failed: malformed JSON payload")
+            return JWTValidationResult.MalformedJson
+        }
+
+        val expRaw = claims.numericClaim("exp")
+        if (expRaw == null) {
+            Registry.log.warning("JWT validation failed: missing exp claim")
+            return JWTValidationResult.MissingExpClaim
+        }
+        if (expRaw.isNaN()) {
+            Registry.log.warning("JWT validation failed: malformed JSON payload")
+            return JWTValidationResult.MalformedJson
+        }
+
+        val iatRaw = claims.numericClaim("iat")
+        if (iatRaw == null) {
+            Registry.log.warning("JWT validation failed: missing iat claim")
+            return JWTValidationResult.MissingIatClaim
+        }
+        if (iatRaw.isNaN()) {
+            Registry.log.warning("JWT validation failed: malformed JSON payload")
+            return JWTValidationResult.MalformedJson
+        }
+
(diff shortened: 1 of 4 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/auth/JWTParser.kt
```diff
@@ -0,0 +1,89 @@
...
+        if (iatRaw == null) {
+            Registry.log.warning("JWT validation failed: missing iat claim")
+            return JWTValidationResult.MissingIatClaim
+        }
+        if (iatRaw.isNaN()) {
+            Registry.log.warning("JWT validation failed: iat claim non-numeric")
+            return JWTValidationResult.MalformedJson
+        }
+
+        if (nowEpochSeconds.toDouble() >= expRaw - leewaySeconds.toDouble()) {
```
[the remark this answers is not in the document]
amber-klaviyo: the terminology may not match, but rejecting BEFORE expiration is the desired behavior. If the token is close to expiration we would rather the SDK reject and refresh in advance, rather than pass along a token that may be expired by the time it is checked.

sdk/analytics/src/main/java/com/klaviyo/analytics/auth/JWTParser.kt
```diff
@@ -0,0 +1,89 @@
+package com.klaviyo.analytics.auth
+
+import android.util.Base64
+import com.klaviyo.core.Registry
+import org.json.JSONException
+import org.json.JSONObject
+
+internal object JWTParser {
+    const val DEFAULT_LEEWAY_SECONDS: Long = 15L
```
ab1470:
```suggestion
    const val DEFAULT_LEEWAY_SECONDS: Long = 30L
```

sorry, I updated the spec to make this 30 seconds instead of 15. I haven't updated the ticket yet (I'm waiting on final spec approval to update all of the tickets

ab1470 (commented):
This LGTM, and I asked Claude to compare it against my [iOS PR](https://github.com/klaviyo/klaviyo-swift-sdk/pull/575/changes) and it says this PR has parity with the iOS one ✅ 

That being said, I'm not going to approve it yet because there's one pending change in the spec that may affect this ticket. I will get back to you on that

sdk/analytics/src/main/java/com/klaviyo/analytics/auth/ValidatedToken.kt
```diff
@@ -0,0 +1,15 @@
...
+internal data class ValidatedToken(
+    val rawToken: String,
+    val expiresAtEpochSeconds: Long,
+    val issuedAtEpochSeconds: Long
+) {
+    // Defensive redaction: prevents the JWT from being exposed if any future call
+    // site ever logs a ValidatedToken (e.g. `log.debug("got $token")`). Token
+    // contents must never be passed to the logger under any circumstances.
+    override fun toString(): String =
+        "ValidatedToken(rawToken=<redacted>, " +
```
evan-masseau, line 12: smart

sdk/analytics/src/main/java/com/klaviyo/analytics/auth/JWTParser.kt
```diff
@@ -0,0 +1,89 @@
+package com.klaviyo.analytics.auth
+
+import android.util.Base64
+import com.klaviyo.core.Registry
+import org.json.JSONException
+import org.json.JSONObject
+
+internal object JWTParser {
+    const val DEFAULT_LEEWAY_SECONDS: Long = 30L
```
evan-masseau, line 9: Just want to ask whether 30 or 15 is the intended default, PR description says 15 and I think i saw 15 in the Swift PR.

evan-masseau (approved):

sdk/analytics/src/main/java/com/klaviyo/analytics/auth/JWTParser.kt
evan-masseau: oops, i see Andrew's comment above, nvm.

ab1470:
> This LGTM, and I asked Claude to compare it against my [iOS PR](https://github.com/klaviyo/klaviyo-swift-sdk/pull/575/changes) and it says this PR has parity with the iOS one ✅
> 
> That being said, I'm not going to approve it yet because there's one pending change in the spec that may affect this ticket. I will get back to you on that

Ok update on this:

sorry for the churn, but I did update this ticket. You can view the change in Linear by clicking the triple-dot menu and clicking "View description history"
Screenshot 2026-05-24 at 15 54 33

But the change is just about where the code should live. Initially the ticket said the code should live in `analytics`, but after discussing with Evan I updated the spec and the ticket to have the code live in `core`.

Hopefully Claude can one-shot the update?

amber-klaviyo:
Closing in favor of https://github.com/klaviyo/klaviyo-android-sdk/pull/458. Same logic, just implemented in the core module as per @ab1470 request due to updated spec

PR #450, opened 2026-05-20
feat(analytics): AuthTokenManager skeleton and registerAuthTokenProvider public API

amber-klaviyo:
# Description

Foundational layer for personalized in-app form authentication. Introduces the `AuthTokenProvider` callback API and the internal `AuthTokenManager` service that the SDK uses to acquire JWTs from the host app. Happy-path acquisition only — concurrency hardening, refresh scheduling, and observer/reset wiring are deferred to sibling sub-issues so this can land focused.

## Due Diligence
- [ ] I have tested this on an emulator and/or a physical device. _(skeleton + unit-tested only; e2e validation lands with [MAGE-619](https://linear.app/klaviyo/issue/MAGE-619) WebView injection)_
- [ ] I have adjusted or added new test cases to team test docs, if applicable. _(deferred until full auth flow lands)_

- [x] `Minor` Contains changes to the public API. _(adds `Klaviyo.registerAuthTokenProvider` + `AuthTokenProvider` type)_

## Changelog / Code Overview

**New public API**
- `Klaviyo.registerAuthTokenProvider(provider: AuthTokenProvider): Klaviyo` — `@JvmStatic`, chainable, fire-and-forget from the host's perspective.
- `fun interface AuthTokenProvider { fun fetchToken(callback: Callback) }` with nested `Callback { onSuccess(jwt) / onFailure(error) }`.

**Internal scaffolding**
- `AuthTokenManager` interface (public for `Registry` lookup from outside `analytics`, e.g. `forms` in [MAGE-619](https://linear.app/klaviyo/issue/MAGE-619)).
- `KlaviyoAuthTokenManager` (`internal`) implementing happy-path `registerProvider` + `currentToken`:
  - Owns `CoroutineScope(SupervisorJob() + Registry.dispatcher)`.
  - `Mutex` guards `provider` / `cachedToken` mutations (held only around state reads/writes — not across the suspend provider invocation).
  - `suspendCancellableCoroutine` bridge with `isActive` guard on both callback paths — late or double callbacks from a buggy host are harmlessly dropped.
  - Returned JWTs are validated via `JWTParser.parseAndValidate`; failure reasons logged at ERROR with the variant name. Token contents are never logged.
  - `inFlightFetch: Deferred<String>?`, `refreshJob: Job?`, and `refreshObservers` are declared as `@Suppress(\"unused\")` placeholders so siblings can fill them in without churning the class signature.
- `Registry.registerOnce<AuthTokenManager>` wired into `Klaviyo.initialize()` next to the existing `State` registration.

**Tests**
- `KlaviyoAuthTokenManagerTest` (9 cases): no-provider error, first-fetch happy path, cache hit, re-register replaces cache, eager fetch on register, `onFailure` propagation, malformed-JWT validation failure (with `Registry.log.error` verification), expired-on-receipt rejection, idempotent late-callback safety.
- `KlaviyoJavaApiTest.testRegisterAuthTokenProvider` confirms Java callers can register via lambda and that the method chains.
- `KlaviyoMock` updated with the new method's mock setup + `verifyRegisterAuthTokenProviderCalled` helper for downstream Java tests.

**Out of scope (sibling sub-issues)**
- Concurrent-caller deduplication + 5s/500ms timeouts → [MAGE-628](https://linear.app/klaviyo/issue/MAGE-628)
- Proactive refresh scheduling → [MAGE-629](https://linear.app/klaviyo/issue/MAGE-629)
- `onTokenRefresh` / `offTokenRefresh` observers and `Klaviyo.resetProfile()` reset wiring → [MAGE-630](https://linear.app/klaviyo/issue/MAGE-630)
- WebView injection consumer → [MAGE-619](https://linear.app/klaviyo/issue/MAGE-619)

## Test Plan

Local (already run):
- `./gradlew :sdk:analytics:testDebugUnitTest` ✅ — full analytics suite green
- `./gradlew :sdk:analytics:testDebugUnitTest --tests \"com.klaviyo.analytics.auth.KlaviyoAuthTokenManagerTest\"` ✅
- `./gradlew :sdk:analytics:testDebugUnitTest --tests \"com.klaviyo.analytics.KlaviyoJavaApiTest.testRegisterAuthTokenProvider\"` ✅
- `./gradlew ktlintCheck` ✅
- `./gradlew :sdk:analytics:build` ✅ (includes lint)

End-to-end validation lands with [MAGE-619](https://linear.app/klaviyo/issue/MAGE-619) when the WebView injection consumer can exercise `currentToken()` against a live in-app form.

## Related Issues/Tickets

- [MAGE-627](https://linear.app/klaviyo/issue/MAGE-627) — this PR
- Parent: [MAGE-618](https://linear.app/klaviyo/issue/MAGE-618) (AuthTokenManager class and public API)
- Spec: [Personalized In-App Forms — Auth Token Integration](https://linear.app/klaviyo/document/personalized-in-app-forms-auth-token-integration-03d3495f6093)

## Base branch note

Stacked on `MAGE-617` (PR #449, in review) because this PR consumes `ValidatedToken` from that change. GitHub will auto-retarget the base to `master` once #449 merges.

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
@@ -6,6 +6,9 @@ import android.content.Intent
 import android.net.Uri
 import androidx.core.app.NotificationManagerCompat
 import androidx.core.net.toUri
+import com.klaviyo.analytics.auth.AuthTokenManager
+import com.klaviyo.analytics.auth.AuthTokenProvider
+import com.klaviyo.analytics.auth.KlaviyoAuthTokenManager
 import com.klaviyo.analytics.linking.DeepLinkHandler
 import com.klaviyo.analytics.linking.DeepLinking
 import com.klaviyo.analytics.model.Event
@@ -28,6 +31,7 @@ import com.klaviyo.core.config.Config
 import com.klaviyo.core.config.LifecycleException
 import com.klaviyo.core.safeApply
 import com.klaviyo.core.safeCall
+import com.klaviyo.core.safeLaunch
 import com.klaviyo.core.utils.JSONUtil.toHashMap
 import com.klaviyo.core.utils.takeIf
 import java.io.Serializable
@@ -103,6 +107,8 @@ object Klaviyo {
             }
         }
 
+        Registry.registerOnce<AuthTokenManager> { KlaviyoAuthTokenManager() }
+
         Registry.get<ApiClient>().startService()
 
         Registry.get<State>().apiKey = apiKey
@@ -139,6 +145,21 @@ object Klaviyo {
     @JvmStatic
     fun unregisterDeepLinkHandler() = safeApply {
         Registry.unregister<DeepLinkHandler>()
+    }
+
+    /**
+     * Register an [AuthTokenProvider] that supplies JWTs to authenticate personalized Klaviyo
+     * features (such as in-app forms that target a known profile).
+     *
+     * Re-registering replaces the previously registered provider and discards any cached token.
+     * The SDK will invoke [AuthTokenProvider.fetchToken] whenever a fresh token is needed; the
+     * host MUST invoke exactly one of [AuthTokenProvider.Callback.onSuccess] or
+     * [AuthTokenProvider.Callback.onFailure] per call.
+     */
+    @JvmStatic
+    fun registerAuthTokenProvider(provider: AuthTokenProvider) = safeApply {
+        val manager = Registry.get<AuthTokenManager>()
+        manager.coroutineScope.safeLaunch { manager.registerProvider(provider) }
     }
 
     /**
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/auth/AuthTokenManager.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/auth/AuthTokenManager.kt
@@ -1,0 +1,34 @@
+package com.klaviyo.analytics.auth
+
+import kotlinx.coroutines.CoroutineScope
+
+/**
+ * Manages the lifecycle of the host-supplied [AuthTokenProvider] and the resulting JWTs used by
+ * personalized Klaviyo features.
+ *
+ * Public for [com.klaviyo.core.Registry] lookup from outside the analytics module (e.g. the forms
+ * module's WebView injection layer). Implementation is internal.
+ */
+interface AuthTokenManager {
+
+    /**
+     * Coroutine scope owned by the manager. The public [com.klaviyo.analytics.Klaviyo] API uses
+     * this scope to launch fire-and-forget calls to the manager's suspending methods.
+     */
+    val coroutineScope: CoroutineScope
+
+    /**
+     * Replace the registered [AuthTokenProvider] (if any), discard any cached token, and eagerly
+     * fetch a fresh token via the new provider.
+     */
+    suspend fun registerProvider(provider: AuthTokenProvider)
+
+    /**
+     * Return a currently-valid JWT, fetching from the registered provider if no cached token is
+     * available or the cached token has expired.
+     *
+     * @throws IllegalStateException if no provider is registered, the provider invokes
+     *   [AuthTokenProvider.Callback.onFailure], or the returned token fails validation.
+     */
+    suspend fun currentToken(): String
+}
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/auth/KlaviyoAuthTokenManager.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/auth/KlaviyoAuthTokenManager.kt
@@ -84,8 +84,9 @@ internal class KlaviyoAuthTokenManager : AuthTokenManager {
             is JWTValidationResult.Valid -> result.token
             else -> {
                 val reason = result::class.simpleName ?: "Unknown"
-                Registry.log.error("Auth token validation failed: $reason")
-                throw IllegalStateException("Auth token validation failed: $reason")
+                val error = IllegalStateException("Auth token validation failed: $reason")
+                Registry.log.error("Auth token validation failed: $reason", error)
+                throw error
             }
         }
 
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/auth/KlaviyoAuthTokenManager.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/auth/KlaviyoAuthTokenManager.kt
@@ -1,0 +1,96 @@
+package com.klaviyo.analytics.auth
+
+import com.klaviyo.core.Registry
+import com.klaviyo.core.safeLaunch
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+
+internal class KlaviyoAuthTokenManager : AuthTokenManager {
+
+    override val coroutineScope: CoroutineScope =
+        CoroutineScope(SupervisorJob() + Registry.dispatcher)
+
+    private val mutex = Mutex()
+
+    private var provider: AuthTokenProvider? = null
+    private var cachedToken: ValidatedToken? = null
+
+    // Reserved for MAGE-628 (concurrent-caller deduplication).
+    @Suppress("unused")
+    private var inFlightFetch: Deferred<String>? = null
+
+    // Reserved for MAGE-629 (proactive refresh scheduling).
+    @Suppress("unused")
+    private var refreshJob: Job? = null
+
+    // Reserved for MAGE-630 (onTokenRefresh/offTokenRefresh observers); typed as a function for
+    // now — the formal observer interface will be introduced alongside the public API.
+    @Suppress("unused")
+    private val refreshObservers = mutableListOf<(String) -> Unit>()
+
+    override suspend fun registerProvider(provider: AuthTokenProvider) {
+        mutex.withLock {
+            this.provider = provider
+            this.cachedToken = null
+        }
+        Registry.log.info("AuthTokenProvider registered")
+
+        coroutineScope.safeLaunch {
+            runCatching { currentToken() }
+                .onFailure { Registry.log.warning("Eager auth token fetch failed", it) }
+        }
+    }
+
+    override suspend fun currentToken(): String {
+        val cached = mutex.withLock { cachedToken }
+        if (cached != null && isStillValid(cached)) {
+            return cached.rawToken
+        }
+
+        val jwt = invokeProvider()
+        val token = validateOrThrow(jwt)
+
+        mutex.withLock { cachedToken = token }
+        Registry.log.info(
+            "Auth token acquired (exp=${token.expiresAtEpochSeconds}, iat=${token.issuedAtEpochSeconds})"
+        )
+        return token.rawToken
+    }
+
+    private suspend fun invokeProvider(): String = suspendCancellableCoroutine { continuation ->
+        val callback = object : AuthTokenProvider.Callback {
+            override fun onSuccess(jwt: String) {
+                if (continuation.isActive) continuation.resume(jwt)
+            }
+
+            override fun onFailure(error: Throwable) {
+                if (continuation.isActive) continuation.resumeWithException(error)
+            }
+        }
+        provider?.fetchToken(callback) ?: continuation.resumeWithException(
+            IllegalStateException("No auth token provider registered")
+        )
+    }
+
+    private fun validateOrThrow(jwt: String): ValidatedToken =
+        when (val result = JWTParser.parseAndValidate(jwt)) {
+            is JWTValidationResult.Valid -> result.token
+            else -> {
+                val reason = result::class.simpleName ?: "Unknown"
+                Registry.log.error("Auth token validation failed: $reason")
+                throw IllegalStateException("Auth token validation failed: $reason")
+            }
+        }
+
+    private fun isStillValid(token: ValidatedToken): Boolean {
+        val now = Registry.clock.currentTimeMillis() / 1000L
+        return now < token.expiresAtEpochSeconds - JWTParser.DEFAULT_LEEWAY_SECONDS
+    }
+}
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/auth/KlaviyoAuthTokenManager.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/auth/KlaviyoAuthTokenManager.kt
@@ -4,6 +4,7 @@ import com.klaviyo.core.Registry
 import com.klaviyo.core.safeLaunch
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Job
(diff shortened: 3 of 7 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/auth/AuthTokenManager.kt
```diff
@@ -0,0 +1,34 @@
+package com.klaviyo.analytics.auth
```
evan-masseau, line 1: This may not have gotten into the linear ticket, but I'd prefer to move the entire `com.klaviyo.analytics.auth` package to `com.klaviyo.core.auth`. The long term plan for SDK arch is for more services to be moved from `analytics` to `core`, so net new stuff should just go there to begin with.

Should be an easy enough move (the rest of this is claude-written but me-approved):

**Zero analytics dependencies in this package.** Every file only imports `com.klaviyo.core.Registry`, `com.klaviyo.core.safeLaunch`, `kotlinx.coroutines.*`, `android.util.Base64`, and `org.json.*` — all already available in `:sdk:core`.

**Cross-cutting fit.** JWT parsing, auth provider, token validation — same category as `Registry.clock` / `Registry.log` / `Registry.dispatcher`. None of it is analytics-specific in the event-tracking/profile-state sense. Also aligns with the long-term direction of moving more services (e.g. `ApiClient`) into core.

**What stays in analytics:** just the public facade glue in `Klaviyo.kt` — the three `import com.klaviyo.core.auth.*` lines, the `@JvmStatic fun registerAuthTokenProvider(...)` public method. Analytics remains the public entry point, since we don't want `core` to be a direct dependency that the host app declares.

**Cost is small:** six file moves + package rename, ~three import lines in `Klaviyo.kt` re-targeted. `KlaviyoAuthTokenManagerTest` moves to `:sdk:core:test` cleanly (only depends on `:sdk:fixtures`, which core already pulls in as `testImplementation`). `KlaviyoMock` doesn't change — it mocks the `Klaviyo` facade method, which stays in analytics. No public API impact (only `AuthTokenManager` and `AuthTokenProvider` are non-internal).

sdk/analytics/src/main/java/com/klaviyo/analytics/auth/KlaviyoAuthTokenManager.kt
```diff
@@ -0,0 +1,104 @@
...
+    private var inFlightFetch: Deferred<String>? = null
+
+    // Reserved for MAGE-629 (proactive refresh scheduling).
+    @Suppress("unused")
+    private var refreshJob: Job? = null
+
+    // Reserved for MAGE-630 (onTokenRefresh/offTokenRefresh observers); typed as a function for
+    // now — the formal observer interface will be introduced alongside the public API.
+    @Suppress("unused")
+    private val refreshObservers = mutableListOf<(String) -> Unit>()
```
evan-masseau, line 37: When this gets fleshed out in MAGE-630, please match the established observer-collection pattern used everywhere else in the SDK:

1. **Typealias for the observer signature** — e.g.
   ```kotlin
   typealias TokenRefreshObserver = (jwt: ValidatedToken) -> Unit
   ```
   Precedent: `NetworkObserver`, `ActivityObserver`, `StoreObserver`, `PermissionObserver`, `GeofenceObserver`, `ApiObserver`, `StateChangeObserver`.

2. **`CopyOnWriteArrayList` for thread safety** — observer notifications happen on whatever coroutine resolves the refresh; adds/removes can come from app code on any thread. Iterating a plain `mutableListOf` while another thread mutates it is a `ConcurrentModificationException` waiting to happen.
   ```kotlin
   private val refreshObservers = CopyOnWriteArrayList<TokenRefreshObserver>()
   ```
   Precedent: every observer collection in the SDK uses this — see `KLog`, `KlaviyoNetworkMonitor`, `KlaviyoLifecycleMonitor`, `SharedPreferencesDataStore`, `KlaviyoPermissionMonitor`, `KlaviyoLocationManager`, `KlaviyoApiClient`, `KlaviyoState`.

Not a blocker for this PR since the field is unused — but worth fixing the placeholder now so the pattern is right when MAGE-630 wires it up (or just remove the placeholder until then, per the other cursor-bot suggestion).

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -103,6 +107,8 @@ object Klaviyo {
             }
         }
 
+        Registry.registerOnce<AuthTokenManager> { KlaviyoAuthTokenManager() }
```
evan-masseau, line 110: Related to my below comment, this `Registry.registerOnce` could be moved from `Klaviyo.initialize` to a core-level `ContentProvider` auto-registration, matching the pattern used in `forms-core` and `location-core`. That would be a nice clean way for core dependencies to be registered from the package that owns them.

sdk/analytics/src/main/java/com/klaviyo/analytics/auth/AuthTokenManager.kt
```diff
@@ -0,0 +1,34 @@
...
+     * Coroutine scope owned by the manager. The public [com.klaviyo.analytics.Klaviyo] API uses
+     * this scope to launch fire-and-forget calls to the manager's suspending methods.
+     */
+    val coroutineScope: CoroutineScope
+
+    /**
+     * Replace the registered [AuthTokenProvider] (if any), discard any cached token, and eagerly
+     * fetch a fresh token via the new provider.
+     */
+    suspend fun registerProvider(provider: AuthTokenProvider)
```
evan-masseau, line 24: No `unregister`? Or coming later?

sdk/analytics/src/main/java/com/klaviyo/analytics/auth/AuthTokenManager.kt
```diff
@@ -0,0 +1,34 @@
...
+ * Public for [com.klaviyo.core.Registry] lookup from outside the analytics module (e.g. the forms
+ * module's WebView injection layer). Implementation is internal.
+ */
+interface AuthTokenManager {
+
+    /**
+     * Coroutine scope owned by the manager. The public [com.klaviyo.analytics.Klaviyo] API uses
+     * this scope to launch fire-and-forget calls to the manager's suspending methods.
+     */
+    val coroutineScope: CoroutineScope
```
evan-masseau, line 18: Exposing `coroutineScope` as a public property of the interface is a recognized structured-concurrency anti-pattern. Roman Elizarov (Kotlin coroutines lead) [calls this out explicitly](https://elizarov.medium.com/coroutine-scopes-and-cancellation-1e3f2eaca4eb): a class should either expose suspend functions (callers bring their own scope) OR internalize launching and expose non-suspend entry points (callers don't see scopes at all). Doing both — "here's a suspend function and *also* here's my scope so you can launch it yourself" — is the worst of both worlds.

The proximate reason it's here: `Klaviyo.registerAuthTokenProvider` is non-suspend public API and needs *some* scope to launch `manager.registerProvider(...)` into. But that's solvable without leaking the scope onto the interface.

**Concrete concern for MAGE-619:** the forms WebView consumer will face the choice of:
```kotlin
// Bad — ties forms work to the token manager's lifecycle:
manager.coroutineScope.launch { val jwt = manager.currentToken() ... }

// Good — forms uses its own scope:
formsScope.launch { val jwt = manager.currentToken() ... }
```
If `coroutineScope` is on the interface, lazy implementations will reach for the convenient option. Then forms work is bound to the token manager's SupervisorJob — wrong lifecycle.

**Suggested shape:**
```kotlin
interface AuthTokenManager {
    // Non-suspend fire-and-forget: manager handles launching internally.
    fun register(provider: AuthTokenProvider)

    // Suspend: caller brings their own scope.
    suspend fun currentToken(): String
}

internal class KlaviyoAuthTokenManager : AuthTokenManager {
    private val scope = CoroutineScope(SupervisorJob() + Registry.dispatcher)

    override fun register(provider: AuthTokenProvider) {
        scope.safeLaunch { registerSuspending(provider) }
    }

    private suspend fun registerSuspending(provider: AuthTokenProvider) {
        mutex.withLock { this.provider = provider; this.cachedToken = null }
        // ... eager fetch
    }

    override suspend fun currentToken(): String { ... }
}
```

Then `Klaviyo.kt` becomes:
```kotlin
fun registerAuthTokenProvider(provider: AuthTokenProvider) = safeApply {
    Registry.get<AuthTokenManager>().register(provider)
}
```

The manager owns its concurrency model, `Klaviyo` doesn't need to know about coroutines, and MAGE-619 is forced to do the right thing (use its own scope, call `currentToken()` from it).

sdk/analytics/src/main/java/com/klaviyo/analytics/auth/KlaviyoAuthTokenManager.kt
```diff
@@ -0,0 +1,104 @@
...
+        )
+    }
+
+    private fun validateOrThrow(jwt: String): ValidatedToken =
+        when (val result = JWTParser.parseAndValidate(jwt)) {
+            is JWTValidationResult.Valid -> result.token
+            else -> {
+                val reason = result::class.simpleName ?: "Unknown"
+                val error = IllegalStateException("Auth token validation failed: $reason")
+                Registry.log.error("Auth token validation failed: $reason", error)
```
evan-masseau, line 95: nit: duplicated error message, could you do `error.message` here at least?

sdk/analytics/src/main/java/com/klaviyo/analytics/auth/KlaviyoAuthTokenManager.kt
```diff
@@ -0,0 +1,104 @@
...
+        Registry.log.info("AuthTokenProvider registered")
+
+        coroutineScope.safeLaunch {
+            try {
+                currentToken()
+            } catch (e: CancellationException) {
+                // Preserve structured concurrency by rethrowing cancellation.
+                throw e
+            } catch (e: Exception) {
+                Registry.log.warning("Eager auth token fetch failed", e)
```
evan-masseau, line 53: nit: looks like we'd get duplicative error logging from this warning, and the `log.error` in `validateOrThrow`

evan-masseau: Could just demote this to `debug` and drop the exception argument maybe

sdk/analytics/src/main/java/com/klaviyo/analytics/auth/AuthTokenManager.kt
evan-masseau: Actually looking closer, its concerning that registering the provider is an async call at all, for example, if we promote a chained call like this:

```
Klaviyo.initialize()
    .registerAuthTokenProvider()   // provider is registered async
    .registerForInAppForms()        // in app forms starts up sync
```

Then i'd be worried that we're creating a race condition unnecessarily. Simply registering the provider should be safe to be a sync call, just might need to change the usage of the `Mutex` which is maybe overkill for the situation. I had to ask Claude about this, here's what it suggested. 

Suggest switching provider and cachedToken to @Volatile or AtomicReference, probably just Volatile (we've used it elsewhere, see `FloatingFormWindow.isDismissed`). The Mutex stays useful for currentToken's read-validate-write transition (and gets subsumed by MAGE-628's Deferred design), but it shouldn't be on the registration path.
```
  @Volatile private var provider: AuthTokenProvider? = null
  @Volatile private var cachedToken: ValidatedToken? = null

  // Non-suspend — returns when the provider reference is set
  fun register(provider: AuthTokenProvider) {
      this.cachedToken = null     // clear first
      this.provider = provider    // then swap
      Registry.log.info("AuthTokenProvider registered")
      internalScope.safeLaunch { tryEagerFetch() }  // eager fetch stays async
  }
```

Then Klaviyo.kt is just 
```
  @JvmStatic
  fun registerAuthTokenProvider(provider: AuthTokenProvider): Klaviyo = safeApply {
      Registry.get<AuthTokenManager>().register(provider)
  }   
```

ab1470:
Similar to #449 and #454, I updated the Linear ticket for this. Here's Claude's summary of what needs to change:

> the manager class itself moves from `analytics` to `core`. Package path changes (`com.klaviyo.analytics.auth` → `com.klaviyo.core.auth`), and the new `lifecycleMonitor: LifecycleMonitor = Registry.lifecycleMonitor` constructor parameter needs to be added now even though it's unused at this stage.

sdk/analytics/src/main/java/com/klaviyo/analytics/auth/AuthTokenManager.kt
ab1470: sorry both @amber-klaviyo and @evan-masseau, I created this ticket before [our Linear conversation](https://linear.app/klaviyo/document/personalized-in-app-forms-auth-token-integration-03d3495f6093#comment-644b3684) about putting this code in `core` instead of `analytics`. I have since updated the ticket; see my comment:
https://github.com/klaviyo/klaviyo-android-sdk/pull/450#issuecomment-4529089943

sorry about the churn :/

hopefully Claude will make it easy to move this into `core`

sdk/analytics/src/main/java/com/klaviyo/analytics/auth/AuthTokenManager.kt
ab1470: sorry, this was my oversight in the Linear ticket. While it would have made sense to add this within MAGE-627, I've already merged the iOS equivalent, so I just created a new 1-point ticket for this. See [MAGE-686](https://linear.app/klaviyo/issue/MAGE-686/android-public-unregisterauthtokenprovider-api)

Should be a very easy add; I bet Claude can one-shot it

sdk/analytics/src/main/java/com/klaviyo/analytics/auth/KlaviyoAuthTokenManager.kt
```diff
@@ -0,0 +1,104 @@
...
+            override fun onSuccess(jwt: String) {
+                if (continuation.isActive) continuation.resume(jwt)
+            }
+
+            override fun onFailure(error: Throwable) {
+                if (continuation.isActive) continuation.resumeWithException(error)
+            }
+        }
+        provider?.fetchToken(callback) ?: continuation.resumeWithException(
+            IllegalStateException("No auth token provider registered")
```
ab1470, line 85: Claude flagged this drift between my (already merged) [iOS-equivalent PR](https://github.com/klaviyo/klaviyo-swift-sdk/pull/576) and this Android PR. I think it's a valid concern:

> Typed error enum vs generic `IllegalStateException`
>
>  iOS has a typed error so consumers can branch on it:
>
>  ```swift
>  package enum AuthTokenError: Error, Equatable {
>      case noProviderRegistered
>      case validationFailed(JWTValidationFailure)
>  }
> ```
>  
>  The doc comment explicitly justifies this:
>
> > "The cases exist so SDK modules that consume `AuthTokenManager` (e.g. KlaviyoForms) can distinguish between 'no provider has been registered yet' and 'the provider returned an invalid token' when shaping their fallback behavior."
>
 > Android throws IllegalStateException for both cases, distinguishable only by string-matching the message:
>  `IllegalStateException("No auth token provider registered")`
>  `IllegalStateException("Auth token validation failed: $reason")`
>  
>  The interface KDoc bundles them together:
>
> > "@throws IllegalStateException if no provider is registered, the provider invokes onFailure, or the returned token fails validation."
>
>  When MAGE-619 (Android WebView injection) consumes `currentToken()`, it'll need to distinguish "auth disabled, fall back silently" from "auth attempted but token invalid, log loudly" — exactly what the iOS comment anticipates. Without typed errors, that becomes message-string matching.
>
>  Recommendation: Add a sealed class hierarchy mirroring the iOS enum:
> ```kotlin
>  sealed class AuthTokenException(message: String) : RuntimeException(message) {
>      object NoProviderRegistered : AuthTokenException("…")
>      data class ValidationFailed(val reason: JWTValidationFailure) : AuthTokenException("…")
>  }
> ```

sdk/analytics/src/main/java/com/klaviyo/analytics/auth/KlaviyoAuthTokenManager.kt
ab1470: FWIW, Claude also added placeholder fields in my equivalent iOS PR, but I ended up removing them so as not to be too prescriptive about implementation before even starting on the relevant tickets. I think Claude was being too eager on scope for this ticket; I think figuring out the appropriate types/patterns for future tickets should be the business of those tickets, not this ticket.

It's up to you as to whether to leave these fields in place or remove them, just letting you know my thinking on the iOS side

amber-klaviyo:
Closing this PR in favor of https://github.com/klaviyo/klaviyo-android-sdk/pull/459 which has the files in core rather than analytics

sdk/analytics/src/main/java/com/klaviyo/analytics/auth/KlaviyoAuthTokenManager.kt
ab1470: marking this as resolved because it's resolved in #459

PR #454, opened 2026-05-22
MAGE-619

amber-klaviyo:
# Description
Phase 1 of [MAGE-619](https://linear.app/klaviyo/issue/MAGE-619): consume the JWT from AuthTokenManager (added in MAGE-617/MAGE-627) at IAF WebView load and inject it into <head data-klaviyo-jwt='…'> so fender's data-klaviyo-jwt MutationObserver ([MAGE-553](https://linear.app/klaviyo/issue/MAGE-553)) can serve personalized form content on first render. Mirrors iOS [PR #577](https://github.com/klaviyo/klaviyo-swift-sdk/pull/577) (initial-injection only — concurrent-fetch dedup + 500ms deadline land with [MAGE-628](https://linear.app/klaviyo/issue/MAGE-628), live refresh with [MAGE-630](https://linear.app/klaviyo/issue/MAGE-630)).

## Due Diligence

- [ ] I have tested this on an emulator and/or a physical device.
- [ ] I have added sufficient unit/integration tests of my changes.
- [ ] I have adjusted or added new test cases to team test docs, if applicable.
- [ ] I am confident these changes are compatible with all Android versions the SDK currently supports.

## Release/Versioning Considerations

- [ ] `Patch` Contains internal changes or backwards-compatible bug fixes.
- [ ] `Minor` Contains changes to the public API. 
- [ ] `Major` Contains **breaking** changes.
- [ ] Contains readme or migration guide changes.
  - If so, please merge to a feature branch so documentation updates only go live upon official release.
- [ ] This is planned work for an upcoming release.
  - If no, author or reviewer should account for this in a release plan, or describe why not below.

PR #460, opened 2026-05-27
feat(forms): inject JWT auth token into IAF WebView at load (MAGE-619)

amber-klaviyo:
## Summary

Phase 1 of MAGE-619: inject the JWT auth token into the In-App Forms WebView's HTML `<head>` at template load time so fender's `data-klaviyo-jwt` MutationObserver (MAGE-553, already merged) can serve personalized content on first render.

- Adds `data-klaviyo-jwt='JWT_TOKEN'` placeholder to `InAppFormsTemplate.html`, following the existing `data-klaviyo-*` substitution convention.
- Refactors `KlaviyoWebViewClient.initializeWebView()` to fetch the auth token asynchronously via `Registry.get<AuthTokenManager>().currentToken()` before loading the WebView template. Synchronous substitutions (SDK metadata, bridge config, device info) still run on the calling (UI) thread; only the auth token requires the async hop.
- Adds `initJob` cancellation in `destroyWebView()` and a stale-ref guard in the `runOnUiThread` block to handle the race where the WebView is destroyed while the token fetch is in flight.
- Logging: `debug` on success, `warning` on unavailability — token contents are never logged.
- Three new tests: token injected, no-token fallback, stale-WebView cancellation. Existing tests updated with `dispatcher.scheduler.advanceUntilIdle()` where `loadTemplate` is now deferred.

**Attribute name:** `data-klaviyo-jwt` — aligned with fender's `JWT_ATTRIBUTE` constant per Daniel's comment on the prior PR.

**iOS counterpart:** https://github.com/klaviyo/klaviyo-swift-sdk/pull/577

**Supersedes:** https://github.com/klaviyo/klaviyo-android-sdk/pull/454 (rebased on `feat/jwts-for-personalized-forms` with logic moved to `forms` module per updated spec; auth module import is now from `core` not `analytics`)

**Blocked portions (TODO comments in code):**
- Best-effort 500ms timeout at form-display time — waiting on **MAGE-628**
- Live-form token refresh via `onTokenRefresh` observer — waiting on **MAGE-630**

## Test plan

- [ ] All `KlaviyoWebViewClientTest` tests pass (`./gradlew :sdk:forms:testDebugUnitTest`)
- [ ] ktlint clean (`./gradlew :sdk:forms:ktlintCheck`)
- [ ] Manual: register an `AuthTokenProvider` in the sample app, trigger an IAF form, confirm `data-klaviyo-jwt` is present in the WebView head with the expected token value
- [ ] Manual: trigger a form without a registered provider; confirm form loads normally (no-token fallback)

### DIFF
--- a/sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
@@ -122,9 +122,9 @@ internal class KlaviyoWebViewClient() : AndroidWebViewClient(), WebViewClient, J
             }
 
             if (authToken != null) {
-                Registry.log.info("Auth token injected at load")
+                Registry.log.debug("Auth token injected at load")
             } else {
-                Registry.log.info("Auth token unavailable at load — proceeding without token")
+                Registry.log.warning("Auth token unavailable at load — proceeding without token")
             }
 
             Registry.threadHelper.runOnUiThread {
--- a/sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
@@ -16,6 +16,7 @@ import androidx.webkit.WebViewFeature.WEB_MESSAGE_LISTENER
 import androidx.webkit.WebViewFeature.isFeatureSupported
 import com.klaviyo.core.Registry
 import com.klaviyo.core.auth.AuthTokenManager
+import com.klaviyo.core.auth.ValidatedToken
 import com.klaviyo.core.config.Clock
 import com.klaviyo.core.safeLaunch
 import com.klaviyo.core.utils.WeakReferenceDelegate
@@ -111,44 +112,57 @@ internal class KlaviyoWebViewClient() : AndroidWebViewClient(), WebViewClient, J
             .replace("DEVICE_INFO", DeviceInfoProvider.current().toJson())
 
         initJob = scope.safeLaunch {
-            // TODO: [MAGE-628] AuthTokenManager will enforce its own 500ms best-effort deadline.
-            // TODO: [MAGE-630] Subscribe to AuthTokenManager.onTokenRefresh for live updates.
-            val authToken: String? = try {
-                Registry.get<AuthTokenManager>().currentToken()
-            } catch (e: CancellationException) {
-                throw e
-            } catch (_: Exception) {
-                null
-            }
+            loadTemplateWithAuthToken(webView, partialHtml, nativeBridge)
+        }
+    }
 
-            if (authToken != null) {
-                Registry.log.debug("Auth token injected at load")
-            } else {
-                Registry.log.warning("Auth token unavailable at load — proceeding without token")
-            }
+    /**
+     * Fetch the auth token, log the outcome, then dispatch back to the UI thread to inject the
+     * (escaped) JWT into [partialHtml] and load it. [webView] is the locally-captured reference
+     * from [initializeWebView]; the body re-checks `this.webView !== webView` after the dispatch
+     * because the field is held weakly and may be cleared by destroy or GC mid-fetch.
+     */
+    private suspend fun loadTemplateWithAuthToken(
+        webView: KlaviyoWebView,
+        partialHtml: String,
+        nativeBridge: NativeBridge
+    ) {
+        // TODO: [MAGE-628] AuthTokenManager will enforce its own 500ms best-effort deadline.
+        // TODO: [MAGE-630] Subscribe to AuthTokenManager.onTokenRefresh for live updates.
+        val token: ValidatedToken? = try {
+            Registry.get<AuthTokenManager>().currentToken()
+        } catch (e: CancellationException) {
+            throw e
+        } catch (_: Exception) {
+            null
+        }
 
-            Registry.threadHelper.runOnUiThread {
-                // webView is held weakly; may be cleared by destroy or GC mid-fetch.
-                if (this@KlaviyoWebViewClient.webView !== webView) {
-                    Registry.log.debug("Skipping IAF template load: WebView no longer current")
-                    return@runOnUiThread
-                }
-                val jwtValue = authToken?.let(::escapeForHtmlAttribute).orEmpty()
-                partialHtml.replace(JWT_TOKEN_PLACEHOLDER, jwtValue).let { html ->
-                    webView.loadTemplate(html, this@KlaviyoWebViewClient, nativeBridge)
-                    handshakeTimer?.cancel()
-                    handshakeTimer = Registry.clock.schedule(
-                        Registry.config.networkTimeout.toLong(),
-                        ::onJsHandshakeTimeout
-                    )
-                }
+        if (token != null) {
+            Registry.log.debug("Auth token injected at load")
+        } else {
+            Registry.log.warning("Auth token unavailable at load — proceeding without token")
+        }
+
+        Registry.threadHelper.runOnUiThread {
+            if (this.webView !== webView) {
+                Registry.log.debug("Skipping IAF template load: WebView no longer current")
+                return@runOnUiThread
+            }
+            val jwtValue = token?.rawToken?.let(::escapeForHtmlAttribute).orEmpty()
+            partialHtml.replace(JWT_TOKEN_PLACEHOLDER, jwtValue).let { html ->
+                webView.loadTemplate(html, this, nativeBridge)
+                handshakeTimer?.cancel()
+                handshakeTimer = Registry.clock.schedule(
+                    Registry.config.networkTimeout.toLong(),
+                    ::onJsHandshakeTimeout
+                )
             }
         }
     }
 
     /**
      * Defense-in-depth escape for the single-quoted `data-klaviyo-jwt` attribute. Valid JWTs
-     * cannot contain these characters, but a custom [AuthTokenProvider] could bypass validation.
+     * cannot contain these characters, but a custom `AuthTokenProvider` could bypass validation.
      */
     private fun escapeForHtmlAttribute(value: String): String =
         value
--- a/sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
@@ -15,6 +15,7 @@ import androidx.webkit.WebViewCompat
 import androidx.webkit.WebViewFeature.WEB_MESSAGE_LISTENER
 import androidx.webkit.WebViewFeature.isFeatureSupported
 import com.klaviyo.core.Registry
+import com.klaviyo.core.auth.AuthTokenException
 import com.klaviyo.core.auth.AuthTokenManager
 import com.klaviyo.core.auth.ValidatedToken
 import com.klaviyo.core.config.Clock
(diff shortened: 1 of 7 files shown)


sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
```diff
@@ -79,16 +109,53 @@ internal class KlaviyoWebViewClient() : AndroidWebViewClient(), WebViewClient, J
...
             .replace("DEVICE_INFO", DeviceInfoProvider.current().toJson())
-            .let { html ->
-                webView.loadTemplate(html, this, nativeBridge)
-                handshakeTimer?.cancel()
-                handshakeTimer = Registry.clock.schedule(
-                    Registry.config.networkTimeout.toLong(),
-                    ::onJsHandshakeTimeout
-                )
+
+        initJob = scope.safeLaunch {
```
evan-masseau, line 115: nit: I'd extract the body of this to its own method

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
```diff
@@ -79,16 +109,53 @@ internal class KlaviyoWebViewClient() : AndroidWebViewClient(), WebViewClient, J
...
+            // TODO: [MAGE-628] AuthTokenManager will enforce its own 500ms best-effort deadline.
+            // TODO: [MAGE-630] Subscribe to AuthTokenManager.onTokenRefresh for live updates.
+            val authToken: String? = try {
+                Registry.get<AuthTokenManager>().currentToken()
```
evan-masseau: Design question — should `currentToken()` return `ValidatedToken` rather than `String`?

A few reasons it might make more sense:

1. **Redacted `toString()` protects against accidental token logging.** `ValidatedToken` already has a redacted `toString()`. Once we unwrap to raw `String`, that protection is gone — any `Registry.log.debug("got token: $token")` (or an unhandled exception with the string in scope) silently leaks the JWT. With the typed wrapper, that's much harder to do by accident.

2. **Metadata is useful to consumers.** `expiresAtEpochSeconds` is directly relevant to upcoming work:
   - MAGE-628 (500ms timeout) could short-circuit if the cached token is still well within its lifetime
   - MAGE-630 (live refresh observer) needs expiry to refresh proactively
   - This call site could decide *not* to inject a token it knows is about to expire mid-session

3. **The function name reads more honestly.** `currentToken(): ValidatedToken` matches what it actually returns internally — the parsed, validated, cached token object. `currentToken(): String` reads like "give me a string" and hides the work that just happened.

API cost is small: `ValidatedToken` goes from `internal` to `public`, callers that only want the string do `.rawToken`. Java interop is the main objection but it's a clean data class.

Not a blocker on this PR — but worth considering before MAGE-628 / MAGE-630 land and we have multiple call sites consuming the raw string.

evan-masseau (approved):
Looks good on the whole

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
amber-klaviyo: `AuthTokenManager.currentToken()` now returns `ValidatedToken`. Callers unwrap to `.rawToken` at the IAF injection site. `ValidatedToken` was promoted from `internal` to `public` for cross-module use.

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
```diff
@@ -79,16 +109,53 @@ internal class KlaviyoWebViewClient() : AndroidWebViewClient(), WebViewClient, J
...
+            } catch (e: CancellationException) {
+                throw e
+            } catch (_: Exception) {
+                null
             }
+
+            if (authToken != null) {
+                Registry.log.debug("Auth token injected at load")
+            } else {
+                Registry.log.warning("Auth token unavailable at load — proceeding without token")
```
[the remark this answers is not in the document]
amber-klaviyo: Good catch — fix in #464. The catch in `loadTemplateWithAuthToken` now distinguishes `AuthTokenException.NoProviderRegistered` (debug, "Auth not enabled — proceeding without token") from any other thrown exception (warning, "Auth token fetch failed — proceeding without token"). Matches the `AuthTokenException.NoProviderRegistered` KDoc intent: *"Callers should treat this as 'auth is not enabled' rather than 'auth failed.'"*

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
```diff
@@ -45,19 +53,40 @@ internal class KlaviyoWebViewClient() : AndroidWebViewClient(), WebViewClient, J
...
+     * Template substitutions that don't require async work (SDK metadata, bridge config, device
+     * info) are applied synchronously on the calling (UI) thread. Auth token injection requires
+     * an async provider call, so the final [KlaviyoWebView.loadTemplate] invocation is deferred
+     * to a background coroutine that calls [AuthTokenManager.currentToken] and then dispatches
+     * back to the UI thread.
      */
-    override fun initializeWebView(): Unit = webView?.let {
-        Registry.log.debug("Klaviyo webview is already initialized")
-    } ?: KlaviyoWebView().let { webView ->
+    override fun initializeWebView() {
```
ab1470, line 79: pasting [my comment](https://github.com/klaviyo/klaviyo-android-sdk/pull/454#discussion_r3298890836) from the now-closed #454:

> is it ok that we're changing this function signature here? Is this just a refactor or does it functionally change things? 
>
> Forgive my Android/Kotlin ignorance...

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
```diff
@@ -79,14 +111,85 @@ internal class KlaviyoWebViewClient() : AndroidWebViewClient(), WebViewClient, J
...
+        // TODO: [MAGE-630] Subscribe to AuthTokenManager.onTokenRefresh for live updates.
+        // NoProviderRegistered is split out from generic fetch failures: the former is the
+        // expected state for apps not using JWT auth (logged at debug), the latter is degraded
+        // functionality with a fallback (logged at warning). Matches the AuthTokenException
+        // KDoc intent — "treat NoProviderRegistered as 'auth is not enabled' rather than 'auth
+        // failed.'"
+        val outcome: AuthOutcome = try {
+            AuthOutcome.Success(Registry.get<AuthTokenManager>().currentToken())
+        } catch (e: CancellationException) {
+            throw e
```
ab1470, line 141: [same comment](https://github.com/klaviyo/klaviyo-android-sdk/pull/454#discussion_r3298907054) that I left on #454:

> if we get a `CancellationException`, will that prevent the web view from launching at all, or will it still launch?

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
ab1470, line 1: [Same comment](https://github.com/klaviyo/klaviyo-android-sdk/pull/454#discussion_r3298939726) as on #454:

> This PR makes a lot of changes to `initializeWebView()`; it adds a `coroutineScope.safeLaunch { ...` wrapper and a `Registry.threadHelper.runOnUiThread { ...` wrapper, among other changes. Are these necessary for the auth token stuff, or are they improvements Claude just caught along the way and threw in with this PR?

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
amber-klaviyo: Yes, a CancellationException will prevent the WebView from loading the template. My understanding (via claude) is that this is the standard coroutine convention — never swallow CancellationException. Let me know if you want a different approach, but my understanding is that aborting is safer than the alternative, which would be silently injecting an empty token while a teardown is in flight.

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
amber-klaviyo: This change allows three things the prior code can't cleanly express: (a) synchronously prep the template on the UI thread, (b) launch a background coroutine for the suspend auth fetch, (c) assign that coroutine to the initJob field so destroyWebView can cancel it. The ?.let / ?: let chain can't hold a field assignment, so a statement-bodied block is required. This is a functional change, but should not negatively impact any existing callers.

Happy to discuss more, or loop in Evan for more Android expertise on this topic.

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
amber-klaviyo: Both wrappers are related to the auth feature. Some of these changes will be modified further as we tackle the rest of the tickets

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
ab1470: what triggers a `CancellationException` here? 

When you talk about a "teardown in flight" are we talking about a teardown of the IAF Web View? A teardown (i.e. deallocation) of the `AuthTokenManager`? Something else?

If there's some kind of problem with the `AuthTokenManager`, I think we wouldn't want to prevent the IAF Web View from launching.

If the cancellation error is because the web view is getting torn down, then yeah I think it makes sense that we're not going to inject a token into a web view that's getting discarded

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
evan-masseau: Yeah this is actually a good code style improvement. A funciton expression that returns `Unit` is functionally the same as a block body with no return, and the latter is more readable.

```kotlin
fun initializeWebView(): Unit = {
    // something
}

fun initializeWebView() {
    // something
}
```

I got into a habit in the past of using function expressions a lot and using chained kotlin expressions, but when they return `Unit` it  is kinda weird/confusing, so eliminating func expressions with `Unit` return types is a good direction.

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
evan-masseau: TL;DR: what Amber said. Cancellation exceptions are supposed to bubble up to the coroutine's scope to cancel properly.

**What triggers it:** `destroyWebView()` calls `initJob?.cancel()`, which cancels the coroutine. If it's suspended at `currentToken()` (waiting on `suspendCancellableCoroutine` for the provider callback), the coroutines runtime throws `CancellationException` at that suspension point. Nothing inside `currentToken()` itself fires it — the trigger is always the external `cancel()` call.

**Why re-throw:** Without the explicit re-throw, `CancellationException` would be caught by `catch (_: Exception)` and treated as `AuthOutcome.FetchFailed`. The coroutine would then proceed to `runOnUiThread`, where the stale-ref guard (`this.webView !== webView`) would block `loadTemplate` anyway — so the observable end result is identical. But eating `CancellationException` breaks cooperative cancellation: `initJob` completes with status finished-normally instead of cancelled, which violates the coroutines contract and can confuse anything inspecting job state.

**Does it crash the app?** No. `CancellationException` transitions the Job to Cancelled state (not Failed) and is explicitly excluded from `CoroutineExceptionHandler` by the coroutines library. `SupervisorJob` on the scope means child cancellation does not propagate anywhere either.

The test `destroyWebView cancels in-flight initJob during auth token fetch` distinguishes the two paths: with re-throw, the coroutine dies at the suspension point and never reaches `runOnUiThread`; without re-throw it would reach `runOnUiThread` and be stopped by the stale-ref guard. The absence of the 'no longer current' log is the proof that cancellation won cleanly at the suspension point.

evan-masseau (approved):
re-approving in light of some more recent changes and comment, thanks for running it past me again

dan-peluso:
Could we add the jwt mutation to the handshake spec? I needed this for my integration to work:
image

ab1470 (approved):

PR #466, opened 2026-06-01
feat(forms): add jwtMutation to JsBridge handshake spec

amber-klaviyo:
## Summary

- Adds `jwtMutation` to the `HelperFunction` enum in `KlaviyoJsBridge`
- Adds `HandshakeSpec(type = "jwtMutation", version = 1)` to the handshake list
- Updates `KlaviyoJsBridgeTest` expected JSON accordingly

The handshake is the capability advertisement the native SDK sends to the fender JS layer at WebView init. Without `jwtMutation` in the spec, fender doesn't know the SDK supports pushing token updates and won't wire up that path. 

The full live-refresh implementation (`window.jwtMutation(newToken)` called on token refresh via the observer) is tracked in MAGE-630 and is out of scope here.

## Test plan

- [x] `./gradlew :sdk:forms:testDebugUnitTest` — all pass
- [x] `./gradlew :sdk:forms:ktlintCheck` — clean

Addresses: https://github.com/klaviyo/klaviyo-android-sdk/pull/460#issuecomment-4578797038

### DIFF
--- a/sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoJsBridge.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoJsBridge.kt
@@ -19,7 +19,8 @@ internal class KlaviyoJsBridge : JsBridge {
         openForm,
         closeForm,
         profileEvent,
-        setSafeArea
+        setSafeArea,
+        jwtMutation
     }
 
     override val handshake: List<HandshakeSpec> = listOf(
@@ -37,6 +38,10 @@ internal class KlaviyoJsBridge : JsBridge {
         ),
         HandshakeSpec(
             type = HelperFunction.profileEvent.name,
+            version = 1
+        ),
+        HandshakeSpec(
+            type = HelperFunction.jwtMutation.name,
             version = 1
         )
     )
--- a/sdk/forms/src/test/java/com/klaviyo/forms/bridge/KlaviyoJsBridgeTest.kt
+++ b/sdk/forms/src/test/java/com/klaviyo/forms/bridge/KlaviyoJsBridgeTest.kt
@@ -35,7 +35,7 @@ class KlaviyoJsBridgeTest : BaseTest() {
     fun `compileJson returns correct JSON for supported functions`() {
         val expectedJson = KlaviyoJsBridge().handshake.compileJson()
         val actualJson = """
-           [{"type":"profileMutation","version":1},{"type":"lifecycleEvent","version":1},{"type":"closeForm","version":1},{"type":"profileEvent","version":1}]
+           [{"type":"profileMutation","version":1},{"type":"lifecycleEvent","version":1},{"type":"closeForm","version":1},{"type":"profileEvent","version":1},{"type":"jwtMutation","version":1}]
         """.trimIndent()
         assertEquals(actualJson, expectedJson)
     }


dan-peluso (approved):

ab1470 (approved):

sdk/forms/src/main/java/com/klaviyo/forms/webview/KlaviyoWebViewClient.kt
```diff
@@ -98,6 +98,23 @@ internal class KlaviyoWebViewClient() : AndroidWebViewClient(), WebViewClient, J
         // Apply all substitutions that can run synchronously on the calling (UI) thread.
         // DeviceInfoProvider.current() reads UI-thread-only APIs (Display.rotation,
         // decorView.rootWindowInsets) — snapshot it here while we are still on the main thread.
+        //
+        // The JS bridge script is inlined rather than referenced via file:// because Android
+        // WebView blocks file:// resource loads when the page is loaded with an HTTPS base URL
+        // (loadDataWithBaseURL uses allowedOrigin as the base URL). Inlining avoids that
+        // restriction while keeping the bridge script in a single maintainable asset file.
+        val bridgeScript = Registry.config.applicationContext.assets
```
dan-peluso: Honestly I thought we only had to do this because of some local issue I was facing - not super convinced it'd necessary. What changes between adding the new handshake method for us to need this?

sdk/forms/src/main/java/com/klaviyo/forms/bridge/ProfileMutationObserver.kt
```diff
@@ -10,6 +10,15 @@ import com.klaviyo.core.Registry
...
 internal class ProfileMutationObserver : JsBridgeObserver, StateChangeObserver {
 
+    /**
+     * Start on [NativeBridgeMessage.HandShook] rather than [NativeBridgeMessage.JsReady] so the
+     * initial [injectProfile] call fires *after* fender's onsite-personalization module has
+     * registered its MutationObserver on `data-klaviyo-profile`. If we inject earlier (JsReady),
+     * the attribute is already set by the time fender's observer is wired up and fender never sees
+     * the change, so the authenticated profiles fetch is never triggered.
+     */
+    override val startOn: NativeBridgeMessage get() = NativeBridgeMessage.HandShook
```
dan-peluso: @evan-masseau any thoughts on this change? We might be able to design around it on the fender side if you had concerns, I think you're probably the most familiar with the handshake arch. Just want to be sure we're covering any downstream effects of this

evan-masseau: I think @amber-klaviyo mentioned that she wasn't sure how much of these changes were Claude reaching for a fix vs actually required. I'd prefer to know for sure what is required changes. I'd be rather surprised if this, or the change to how the script is loaded, are truly needed. For example this is just not true/not how the onsite side works. The onsite side should be reading initial value and _then_ attaching a listener for any subsequent changes.

> If we inject earlier (JsReady), the attribute is already set by the time fender's observer is wired up and fender never sees the change, so the authenticated profiles fetch is never triggered.

sdk/forms/src/main/assets/InAppFormsTemplate.html
```diff
@@ -19,8 +19,8 @@
 
     <title>Klaviyo In-App Form Template</title>
 
-    <!-- Load in JS helper functions from assets directory -->
-    <script type="text/javascript" src="file:///android_asset/klaviyo-js-bridge.js"></script>
+    <!-- JS helper functions inlined at load time by KlaviyoWebViewClient (file:// access blocked from HTTPS origin) -->
```
evan-masseau: If this were broken, in-app forms wouldn't be working all this time. So I'd prefer not to change it without more compelling reason

sdk/forms/src/main/java/com/klaviyo/forms/bridge/ProfileMutationObserver.kt
dan-peluso: Agree on the loading aspect, but I think we might need to make a change on Fender for us to avoid this one. This was something I had to do when testing to get the jwt mutation to propagate to the onsite personalization module.

evan-masseau: Lets take another look at the onsite side first.
Is the personalization code perhaps only listening for the event, without checking the current value of the data attribute?

evan-masseau: I believe that the risk of this change is timing:

1. `registerForInAppForms` called
2. HTML Template gets filled with current profile data
3. Klaviyo.js loads
4. Native SDK receives the handshake confirmation from the JS

If a profile mutation occurs in the 1-2 seconds between 2 and 4, then the profile data attribute does not get updated, and that change never makes it into the webview at all.

evan-masseau: Hm wait `ProfileMutationObserver.startObserver` updates on start, so maybe it wouldn't have that timing issue. Still curious about why the onsite side can't just read initial profile data from the data attribute tho. IIRC we sync the profile data right into the existing onsite tracking module's state (cookie i think?). Personalization should probably be reading profile data out of onsite tracking's state. That'll give the proper compatibility for web forms anyway

dan-peluso: I think one issue was that we never actually defined js function in klaviyo-js-bridge.js lol

evan-masseau: What JS function? For setting the JWT, or the profile?

dan-peluso: So the current design is that we're setting the JWT once in `loadTemplateWithAuthToken` but we need to support a similar 'use js to set HTML tags' like we do for profile mutations (`window.profileMutation = function`) so that we can update JWT on an active webview rather than just load.

evan-masseau: Yes you would indeed need the JS function to inject updates

dan-peluso: Yeah I think that was the real issue here - the bridge function load change made it so that the JWT and profile HTML tag sets happen in a different order (profile first, then jwt) and onsite code has a check that basically says 'once a jwt is valdated check the html for a profile email' and because that had not been set yet, we didn't make the call. I'm going to cut a new ticket describing how we should refactor the webivew html injection and add the restriction that we need to set the profile before the JWT if we want the JWT to be able to actually make a profile call.

dan-peluso: https://linear.app/klaviyo/issue/MAGE-724/refactorandroid-deliver-jwt-via-windowjwtmutation-js-bridge-instead-of cc @amber-klaviyo

amber-klaviyo:
I've reverted the changes that prompted discussion- we now have a dedicated ticket to fix these issues in a better way: https://linear.app/klaviyo/issue/MAGE-724/refactorandroid-deliver-jwt-via-windowjwtmutation-js-bridge-instead-of 

This change will just implement the handshake, which will be required regardless of the discussion outcomes

PR #467, opened 2026-06-02
feat(core): AuthTokenManager deduplication and timeouts (MAGE-628)

amber-klaviyo:
# Description

Hardens `KlaviyoAuthTokenManager` against concurrent callers and slow/unresponsive providers. Concurrent calls to `currentToken()` that miss the cache now share a single in-flight `Deferred<ValidatedToken>` rather than each triggering an independent provider invocation. Per-caller timeout budgets are enforced without cancelling the underlying task, so a later caller with a larger budget still benefits if the provider eventually responds.

Android counterpart to [klaviyo-swift-sdk#586](https://github.com/klaviyo/klaviyo-swift-sdk/pull/586).

## Due Diligence

> **Note:** `AuthTokenManager.currentToken()` gains a `timeoutMs` parameter with a default value — source-compatible, no callers need to change unless they want the interactive (500 ms) budget.

### `AuthTokenException.kt`
- Added `TimedOut` — new sealed variant thrown when the provider doesn't respond within the caller's budget. Mirrors iOS `AuthTokenError.timedOut`.

### `AuthTokenManager.kt` (interface)
- `currentToken(timeoutMs: Long = BACKGROUND_FETCH_TIMEOUT_MS)` — parameter with default; existing callers unchanged.
- Two public constants: `BACKGROUND_FETCH_TIMEOUT_MS = 5_000L` (proactive/eager fetches) and `INTERACTIVE_FETCH_TIMEOUT_MS = 500L` (form-display time). Accessible from both `core` and `forms` without duplication.

### `KlaviyoAuthTokenManager.kt`
- `inFlightFetch: Deferred<ValidatedToken>?` is now active (`@Volatile` for lock-free reads in `registerProvider`/`invokeOnCompletion` to be visible to the mutex-guarded path in `currentToken`).
- `currentToken()` — optimistic cache check → mutex-guarded deferred read-or-create → per-caller `withTimeoutOrNull`. The mutex re-checks the cache after acquiring (double-checked locking pattern) to avoid a redundant fetch if a concurrent caller already populated it while waiting for the lock.
- `invokeOnCompletion` uses a reference-identity check (`if (inFlightFetch === d)`) so a stale deferred completing after a provider swap doesn't clear a freshly-created slot.
- `registerProvider()` cancels any in-flight deferred before swapping. Two complementary guards prevent stale tokens: (1) `isActive` in `suspendCancellableCoroutine` drops late provider callbacks; (2) `mutex.withLock` in `doFetch` is a cancellation checkpoint that catches the case where the callback already fired but the cache write hasn't run.
- `doFetch()` extracted from `currentToken()` — runs inside `scope.async`, owns the full invoke→validate→cache→log flow.

### `KlaviyoWebViewClient.kt`
- Calls `currentToken(INTERACTIVE_FETCH_TIMEOUT_MS)` — 500 ms best-effort budget at form-display time. Removed `MAGE-628` TODO.

## Test Plan

All tests pass (`./gradlew :sdk:core:testDebugUnitTest :sdk:forms:testDebugUnitTest`). Five new tests in `KlaviyoAuthTokenManagerTest`:

| Test | What it proves |
|------|----------------|
| `concurrent callers share a single provider invocation` | N callers → exactly 1 provider call; all receive same token |
| `currentToken throws TimedOut when provider does not respond within timeout` | `TimedOut` thrown and logged at ERROR within the budget |
| `timeout does not cancel underlying fetch so later callers still benefit` | First caller times out; provider resolves; second caller gets token from cache; `callCount == 1` |
| `failure clears in-flight slot so next call re-invokes provider` | `invokeOnCompletion` fires on failure; slot cleared; subsequent call triggers fresh provider invocation |
| `provider swap cancels in-flight fetch so stale token cannot poison cache` | Provider A's late callback is dropped after swap; cache holds Provider B's token |

Existing tests migrated to `runTest(dispatcher)` so `scope.async` and `withTimeoutOrNull` share a single `TestCoroutineScheduler`.

## Related Issues/Tickets

Part of MAGE-628 — https://linear.app/klaviyo/issue/MAGE-628

### DIFF
--- a/sdk/core/src/main/java/com/klaviyo/core/auth/AuthTokenException.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/auth/AuthTokenException.kt
@@ -24,4 +24,13 @@ sealed class AuthTokenException(message: String) : RuntimeException(message) {
     data class ValidationFailed(val reason: String) : AuthTokenException(
         "Auth token validation failed: $reason"
     )
+
+    /**
+     * The provider did not return a token within the caller's timeout budget.
+     * The underlying provider invocation may still be in-flight; a later caller with a longer
+     * budget can still receive the result without triggering a second provider call.
+     *
+     * Mirrors iOS `AuthTokenError.timedOut`.
+     */
+    data object TimedOut : AuthTokenException("Auth token request timed out")
 }
--- a/sdk/core/src/main/java/com/klaviyo/core/auth/AuthTokenException.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/auth/AuthTokenException.kt
@@ -30,7 +30,6 @@ sealed class AuthTokenException(message: String) : RuntimeException(message) {
      * The underlying provider invocation may still be in-flight; a later caller with a longer
      * budget can still receive the result without triggering a second provider call.
      *
-     * Mirrors iOS `AuthTokenError.timedOut`.
      */
     data object TimedOut : AuthTokenException("Auth token request timed out")
 }
--- a/sdk/core/src/main/java/com/klaviyo/core/auth/AuthTokenManager.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/auth/AuthTokenManager.kt
@@ -27,7 +27,6 @@ interface AuthTokenManager {
         /**
          * Best-effort timeout budget used at form-display time, where latency is user-visible.
          * A timeout here degrades gracefully — the form loads without a token rather than blocking.
-         * Matches iOS `FetchMode.interactive`.
          */
         const val INTERACTIVE_FETCH_TIMEOUT_MS = 500L
     }
--- a/sdk/core/src/main/java/com/klaviyo/core/auth/AuthTokenManager.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/auth/AuthTokenManager.kt
@@ -17,6 +17,21 @@ internal typealias TokenRefreshObserver = (token: ValidatedToken) -> Unit
  */
 interface AuthTokenManager {
 
+    companion object {
+        /**
+         * Timeout budget for proactive / background fetches (registration-time warm-up, scheduled
+         * refresh). Matches iOS `FetchMode.background`.
+         */
+        const val BACKGROUND_FETCH_TIMEOUT_MS = 5_000L
+
+        /**
+         * Best-effort timeout budget used at form-display time, where latency is user-visible.
+         * A timeout here degrades gracefully — the form loads without a token rather than blocking.
+         * Matches iOS `FetchMode.interactive`.
+         */
+        const val INTERACTIVE_FETCH_TIMEOUT_MS = 500L
+    }
+
     /**
      * Replace the registered [AuthTokenProvider] (if any), discard any cached token, and
      * asynchronously pre-warm the cache with a fresh token via the new provider.
@@ -34,9 +49,18 @@ interface AuthTokenManager {
      * directly. [ValidatedToken.toString] is redacted, so the wrapper is safer to handle than
      * the raw string.
      *
+     * Concurrent callers that arrive while a provider fetch is already in-flight share the result
+     * of that single fetch rather than each triggering a new provider invocation. Each caller's
+     * [timeoutMs] budget is enforced independently — a caller that times out does not cancel the
+     * underlying fetch, so a later caller with a larger budget can still receive the result.
+     *
+     * @param timeoutMs Maximum milliseconds to wait for the provider to return a token.
+     *   Defaults to [BACKGROUND_FETCH_TIMEOUT_MS]. Pass [INTERACTIVE_FETCH_TIMEOUT_MS] at
+     *   form-display time for a best-effort, non-blocking fetch.
      * @throws [AuthTokenException.NoProviderRegistered] if no provider has been registered.
      * @throws [AuthTokenException.ValidationFailed] if the returned token fails validation.
+     * @throws [AuthTokenException.TimedOut] if the provider does not respond within [timeoutMs].
      * @throws Throwable whatever error the provider passed to [AuthTokenProvider.Callback.onFailure].
      */
-    suspend fun currentToken(): ValidatedToken
+    suspend fun currentToken(timeoutMs: Long = BACKGROUND_FETCH_TIMEOUT_MS): ValidatedToken
 }
--- a/sdk/core/src/main/java/com/klaviyo/core/auth/AuthTokenManager.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/auth/AuthTokenManager.kt
@@ -20,7 +20,7 @@ interface AuthTokenManager {
     companion object {
         /**
          * Timeout budget for proactive / background fetches (registration-time warm-up, scheduled
-         * refresh). Matches iOS `FetchMode.background`.
+         * refresh).
          */
         const val BACKGROUND_FETCH_TIMEOUT_MS = 5_000L
 
--- a/sdk/core/src/main/java/com/klaviyo/core/auth/AuthTokenManager.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/auth/AuthTokenManager.kt
@@ -54,10 +54,8 @@ interface AuthTokenManager {
      * underlying fetch, so a later caller with a larger budget can still receive the result.
      *
      * @param timeoutMs Maximum milliseconds to wait for the provider to return a token. Must be
-     *   positive. Defaults to [BACKGROUND_FETCH_TIMEOUT_MS] (note: Swift's default is
-     *   [INTERACTIVE_FETCH_TIMEOUT_MS] — pass the desired budget explicitly when cross-platform
-     *   parity matters). Pass [INTERACTIVE_FETCH_TIMEOUT_MS] at form-display time for a
-     *   best-effort, non-blocking fetch.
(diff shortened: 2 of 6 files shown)


sdk/core/src/main/java/com/klaviyo/core/auth/AuthTokenManager.kt
```diff
@@ -34,9 +49,21 @@ interface AuthTokenManager {
...
      * the raw string.
      *
+     * Concurrent callers that arrive while a provider fetch is already in-flight share the result
+     * of that single fetch rather than each triggering a new provider invocation. Each caller's
+     * [timeoutMs] budget is enforced independently — a caller that times out does not cancel the
+     * underlying fetch, so a later caller with a larger budget can still receive the result.
+     *
+     * @param timeoutMs Maximum milliseconds to wait for the provider to return a token. Must be
+     *   positive. Defaults to [BACKGROUND_FETCH_TIMEOUT_MS] (note: Swift's default is
+     *   [INTERACTIVE_FETCH_TIMEOUT_MS] — pass the desired budget explicitly when cross-platform
```
ab1470: > note: Swift's default is [INTERACTIVE_FETCH_TIMEOUT_MS]

any reason for diverging from Swift here?

sdk/core/src/main/java/com/klaviyo/core/auth/AuthTokenManager.kt
```diff
@@ -34,9 +49,21 @@ interface AuthTokenManager {
...
+     *   parity matters). Pass [INTERACTIVE_FETCH_TIMEOUT_MS] at form-display time for a
+     *   best-effort, non-blocking fetch.
      * @throws [AuthTokenException.NoProviderRegistered] if no provider has been registered.
      * @throws [AuthTokenException.ValidationFailed] if the returned token fails validation.
+     * @throws [AuthTokenException.TimedOut] if the provider does not respond within [timeoutMs].
+     * @throws IllegalArgumentException if [timeoutMs] is not positive.
      * @throws Throwable whatever error the provider passed to [AuthTokenProvider.Callback.onFailure].
      */
-    suspend fun currentToken(): ValidatedToken
+    suspend fun currentToken(timeoutMs: Long = BACKGROUND_FETCH_TIMEOUT_MS): ValidatedToken
```
ab1470, line 65: I'm doing this [slightly differently in Swift](https://github.com/klaviyo/klaviyo-swift-sdk/pull/586/changes#diff-6c5a80439270bd9ebea5d5aec04c50f985b09596b7df49155b889696c30ab060R114); instead of taking the timeout in ms as a parameter, I take an enum in which the timeout values are defined.

I think either way is fine, I'm just calling out the difference here. I think(?) enums are more standard practice in Swift than they are in Kotlin?

ab1470 (approved):
LGTM, I had Claude compare this one with my iOS-equivalent PR and it says that this one has parity with the iOS one

sdk/core/src/main/java/com/klaviyo/core/auth/AuthTokenException.kt
```diff
@@ -24,4 +24,13 @@ sealed class AuthTokenException(message: String) : RuntimeException(message) {
     data class ValidationFailed(val reason: String) : AuthTokenException(
         "Auth token validation failed: $reason"
     )
+
+    /**
+     * The provider did not return a token within the caller's timeout budget.
+     * The underlying provider invocation may still be in-flight; a later caller with a longer
+     * budget can still receive the result without triggering a second provider call.
+     *
+     * Mirrors iOS `AuthTokenError.timedOut`.
```
evan-masseau:
```suggestion
```

I find code level comments like this a bit noisy in the long run

sdk/core/src/main/java/com/klaviyo/core/auth/AuthTokenManager.kt
```diff
@@ -17,6 +17,21 @@ internal typealias TokenRefreshObserver = (token: ValidatedToken) -> Unit
  */
 interface AuthTokenManager {
 
+    companion object {
+        /**
+         * Timeout budget for proactive / background fetches (registration-time warm-up, scheduled
+         * refresh). Matches iOS `FetchMode.background`.
```
evan-masseau:
```suggestion
         * refresh).
```

sdk/core/src/main/java/com/klaviyo/core/auth/AuthTokenManager.kt
```diff
@@ -17,6 +17,21 @@ internal typealias TokenRefreshObserver = (token: ValidatedToken) -> Unit
...
+         */
+        const val BACKGROUND_FETCH_TIMEOUT_MS = 5_000L
+
+        /**
+         * Best-effort timeout budget used at form-display time, where latency is user-visible.
+         * A timeout here degrades gracefully — the form loads without a token rather than blocking.
+         * Matches iOS `FetchMode.interactive`.
```
evan-masseau:
```suggestion
```

sdk/core/src/main/java/com/klaviyo/core/auth/AuthTokenManager.kt
evan-masseau: Yea using constants is a pretty typical way to do it in Java/Kotlin, I'm thinking of how the android toast library has SHORT and LONG duration constants for example.

sdk/core/src/main/java/com/klaviyo/core/auth/AuthTokenManager.kt
```diff
@@ -34,9 +49,21 @@ interface AuthTokenManager {
...
+     * Concurrent callers that arrive while a provider fetch is already in-flight share the result
+     * of that single fetch rather than each triggering a new provider invocation. Each caller's
+     * [timeoutMs] budget is enforced independently — a caller that times out does not cancel the
+     * underlying fetch, so a later caller with a larger budget can still receive the result.
+     *
+     * @param timeoutMs Maximum milliseconds to wait for the provider to return a token. Must be
+     *   positive. Defaults to [BACKGROUND_FETCH_TIMEOUT_MS] (note: Swift's default is
+     *   [INTERACTIVE_FETCH_TIMEOUT_MS] — pass the desired budget explicitly when cross-platform
+     *   parity matters). Pass [INTERACTIVE_FETCH_TIMEOUT_MS] at form-display time for a
+     *   best-effort, non-blocking fetch.
```
evan-masseau:
```suggestion
     *   positive. Defaults to [BACKGROUND_FETCH_TIMEOUT_MS], pass [INTERACTIVE_FETCH_TIMEOUT_MS] 
     *  at form-display time for a best-effort, non-blocking fetch.
```

Would think using the same default as swift would make sense yeah. And just calling out another explicit name check of swift SDK in comments. I just worry about drift from referencing another repo

sdk/core/src/main/java/com/klaviyo/core/auth/KlaviyoAuthTokenManager.kt
```diff
@@ -55,26 +71,77 @@ internal class KlaviyoAuthTokenManager(
...
+        // would otherwise propagate to callers as if THEIR coroutine was cancelled, which breaks
+        // structured-concurrency semantics. We catch it, use ensureActive() to distinguish "our
+        // coroutine was cancelled" (rethrow — normal teardown) from "the deferred was cancelled
+        // by a provider swap" (retry — pick up the new provider's fetch transparently).
+        return withTimeoutOrNull(timeoutMs) {
+            try {
+                deferred.await()
+            } catch (e: CancellationException) {
+                currentCoroutineContext().ensureActive()
+                currentToken(timeoutMs)
```
evan-masseau, line 129: I think bugbot is asking about this too, but yeah I'm not sure I understand the situation that would lead to recursively calling `currentToken` here after a cancellation.

evan-masseau: I guess its for when the provider gets changed _while_ a request for the token is inflight. Seems a pretty unusual edge case, I suppose falling back on still trying to fetch the token is OK.

sdk/core/src/main/java/com/klaviyo/core/auth/KlaviyoAuthTokenManager.kt
```diff
@@ -55,26 +71,77 @@ internal class KlaviyoAuthTokenManager(
...
+            }
```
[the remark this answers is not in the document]
evan-masseau: This sounds like it'd be such an edge case it hardly seems worth thinking too much about, but, the caller would care more about the timeout relative to its original calltime i'd think, so this seems like the _right_ behavior

sdk/core/src/test/java/com/klaviyo/core/auth/KlaviyoAuthTokenManagerTest.kt
```diff
@@ -192,6 +205,212 @@ class KlaviyoAuthTokenManagerTest : BaseTest() {
         assertEquals(token, result.rawToken)
     }
 
+    // MARK: - MAGE-628: Deduplication and Timeouts
```
evan-masseau:
```suggestion
    // MARK: - Deduplication and Timeouts
```

I typically avoid putting ticket numbers in except for TODOs

evan-masseau:
I didn't see a test that fires a provider swap WHILE a caller with is waiting on current token, to verify that the outer `withTimeoutOrNull` caps the total wait correctly even though the recursive retry gets a fresh budget. Would be easier to test that with unit tests than try to repro manually.

On the whole looks good though, AI wants to find all the little timing edge cases which is useful i guess but only up to a point. I think the proof will in part be in the real world testing here.

evan-masseau (approved):

PR #469, opened 2026-06-02
feat(core): AuthTokenManager proactive refresh scheduling (MAGE-629)

amber-klaviyo:
## Summary

- Schedules a proactive token refresh at 90% of the token's lifetime using `Registry.clock.schedule()` (Java `Timer`-backed, wall-clock-aware — not coroutine `delay`)
- Refresh target formula: `iat + 0.9 * (exp - iat)`, clamped to `[now + 5s, exp − leeway]`
- `FirstStarted` lifecycle observer reconciles state on foreground transitions: expired token → clear + eager fetch, missed refresh → fire immediately, still valid → no-op
- `performScheduledRefresh` clears the cache before calling `currentToken()` so the optimistic cache read falls through and a fresh provider invocation is forced; dedup via `inFlightFetch` applies as normal
- `refreshAtWallClockMs` is intentionally not cleared in the timer callback — a failed refresh leaves the stale target visible so `handleForegroundTransition` case 2 can detect and retry exactly once
- `computeRefreshTarget` extracted as an `internal` companion function for deterministic unit testing
- Registers lifecycle observer via named method reference `::onLifecycleEvent` + `takeIf<ActivityEvent.FirstStarted>()` filter (established codebase pattern)
- TODO left at `performScheduledRefresh` success path for MAGE-630 observer notification hook

## Test plan

- [ ] `computeRefreshTarget` pure-function tests (3): ideal 90%, upper clamp, lower clamp
- [ ] Scheduler integration: timer fires at computed target, chains next refresh
- [ ] `registerProvider` cancels pending refresh job before it fires
- [ ] Foreground transitions: still-valid no-op, non-FirstStarted events ignored, missed-refresh retry, expired-token eager fetch
- [ ] All existing `KlaviyoAuthTokenManagerTest` tests continue to pass (44 total)

### DIFF
--- a/sdk/core/src/main/java/com/klaviyo/core/auth/KlaviyoAuthTokenManager.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/auth/KlaviyoAuthTokenManager.kt
@@ -1,8 +1,11 @@
 package com.klaviyo.core.auth
 
 import com.klaviyo.core.Registry
+import com.klaviyo.core.config.Clock
+import com.klaviyo.core.lifecycle.ActivityEvent
 import com.klaviyo.core.lifecycle.LifecycleMonitor
 import com.klaviyo.core.safeLaunch
+import com.klaviyo.core.utils.takeIf
 import java.util.concurrent.CopyOnWriteArrayList
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
@@ -18,17 +21,17 @@ import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withTimeoutOrNull
 
-/**
- * @param lifecycleMonitor declared in the constructor to lock the signature for MAGE-629's
- *   lifecycle-aware refresh work; unused by code in this PR.
- */
 internal class KlaviyoAuthTokenManager(
-    @Suppress("unused") private val lifecycleMonitor: LifecycleMonitor = Registry.lifecycleMonitor
+    private val lifecycleMonitor: LifecycleMonitor = Registry.lifecycleMonitor
 ) : AuthTokenManager {
 
     // Internal (not on the interface) so MAGE-619 consumers are forced to use their own scope
     // when calling currentToken(), binding auth work to the correct lifecycle.
     internal val scope: CoroutineScope = CoroutineScope(SupervisorJob() + Registry.dispatcher)
+
+    init {
+        lifecycleMonitor.onActivityEvent(::onLifecycleEvent)
+    }
 
     // Guards the read-validate-write transition on both cachedToken and inFlightFetch, ensuring
     // exactly one Deferred is created when multiple callers miss the cache simultaneously.
@@ -47,6 +50,12 @@ internal class KlaviyoAuthTokenManager(
     // mutex; @Volatile ensures those writes are visible to the mutex-protected read in currentToken.
     @Volatile private var inFlightFetch: Deferred<ValidatedToken>? = null
 
+    // NOT cleared in the timer callback on firing — a failed refresh leaves this pointing at a
+    // past target so handleForegroundTransition() case 2 can detect the miss and retry once.
+    @Volatile private var refreshJob: Clock.Cancellable? = null
+
+    @Volatile private var refreshAtWallClockMs: Long? = null
+
     // Reserved for MAGE-630 (onTokenRefresh/offTokenRefresh observers). Matches the established
     // SDK observer-collection pattern (CopyOnWriteArrayList for thread-safe iteration while
     // observers add/remove themselves on arbitrary threads).
@@ -63,6 +72,9 @@ internal class KlaviyoAuthTokenManager(
         //      before the write — even when mutex.withLock acquires the lock uncontended.
         inFlightFetch?.cancel()
         inFlightFetch = null
+        refreshJob?.cancel()
+        refreshJob = null
+        refreshAtWallClockMs = null
         cachedToken = null
         this.provider = provider
         Registry.log.info("AuthTokenProvider registered")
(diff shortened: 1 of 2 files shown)


sdk/core/src/main/java/com/klaviyo/core/auth/KlaviyoAuthTokenManager.kt
```diff
@@ -82,20 +112,32 @@ internal class KlaviyoAuthTokenManager(
     }
 
     override suspend fun currentToken(timeoutMs: Long): ValidatedToken {
+        return currentTokenInternal(timeoutMs = timeoutMs, allowCachedToken = true)
+    }
+
+    private suspend fun currentTokenInternal(
```
ab1470:
```suggestion
    /**
     * Shared implementation behind [currentToken]. Split out so the
     * `allowCachedToken` knob stays off the public [AuthTokenManager] interface —
     * external callers always get the cache-honoring behavior.
     *
     * @param allowCachedToken When `true` (the [currentToken] path), a still-valid
     *   cached token short-circuits both the optimistic pre-lock read and the
     *   double-checked read under the mutex. When `false`, both reads are skipped
     *   and the call always resolves through the in-flight fetch, forcing a fresh
     *   provider invocation even if the cache is currently valid.
     *
     *   Only the proactive-refresh path ([performScheduledRefresh]) passes `false`:
     *   a refresh fires *because* the cached token is aging, so returning that
     *   still-valid token would make the refresh a no-op. Note this only bypasses
     *   the *read* — dedup still applies (a concurrent fetch is joined, not
     *   duplicated, via [inFlightFetch]), and the existing cache is left intact so
     *   demand callers keep getting the valid token while the refresh runs.
     */
    private suspend fun currentTokenInternal(
```
I asked Claude to write inline code documentation for this so I could better understand what that paraeter is doing. Feel free to modify the suggested documentation as you see fit, but I do think it would be helpful to have some documentation here

ab1470 (approved):
LGTM. Claude checked for parity with the [Swift PR](https://github.com/klaviyo/klaviyo-swift-sdk/pull/588) and all looks good

dan-peluso (commented):
Apologies on the time it took to review, this definitely got super complex with the concurrency stuff.  I think my only concern is adding some edge case tests for some things: 

1. timeout during a scheduled refresh - performScheduledRefresh calls currentTokenInternal(timeoutMs = BACKGROUND_FETCH_TIMEOUT_MS, ...). If the provider hangs long enough to hit the timeout, withTimeoutOrNull returns null, AuthTokenException.TimedOut is thrown, and it's caught by the catch (e: Exception) block. clearFiredFlagForFailedRefresh is called. But there's no test verifying this path specifically — timeout failure should behave the same as a provider error, including leaving one foreground retry available.

2. Failed foreground retry does not create a second retry - The foreground retries after scheduled timer refresh fails test verifies that a failed timer refresh leaves one foreground retry. But there's no test verifying the inverse: after that foreground retry fires (regardless of outcome), a subsequent foreground transition is a no-op and doesn't retry again. Specifically, refreshAtWallClockMs is cleared in the foreground case 2 branch before the coroutine launches, so a third foreground should be still-valid. This is implied by the code but not asserted.

3. allowCachedToken=false + dedup interaction

4. exp < iat (malformed JWT where expiry predates issuance). iatMs + 0.9 * (expMs - iatMs) would produce a value less than iatMs, which gets clamped to lowerBound. Worth a test since JWTParser.parseAndValidate presumably rejects this — but knowing the math degrades gracefully rather than producing a negative delay would be reassuring.


I bet Claude can crush those, lmk if you had any followups!

sdk/core/src/main/java/com/klaviyo/core/auth/KlaviyoAuthTokenManager.kt
```diff
@@ -82,20 +112,50 @@ internal class KlaviyoAuthTokenManager(
...
+     *   provider invocation even if the cache is currently valid.
+     *
+     *   Only the proactive-refresh path ([performScheduledRefresh]) passes `false`:
+     *   a refresh fires *because* the cached token is aging, so returning that
+     *   still-valid token would make the refresh a no-op. Note this only bypasses
+     *   the *read* — dedup still applies (a concurrent fetch is joined, not
+     *   duplicated, via [inFlightFetch]), and the existing cache is left intact so
+     *   demand callers keep getting the valid token while the refresh runs.
+     */
+    private suspend fun currentTokenInternal(
```
evan-masseau: I don't think this method's name communicates what it actually does. I'd either break it up a little, or rename it to something like `getOrFetchToken` so it's clearer that this is multipurpose (return cached token OR force a fresh provider fetch).

sdk/core/src/main/java/com/klaviyo/core/auth/KlaviyoAuthTokenManager.kt
```diff
@@ -47,6 +51,27 @@ internal class KlaviyoAuthTokenManager(
...
+    // NOT cleared in the timer callback on firing — a failed refresh leaves this pointing at a
+    // past target so handleForegroundTransition() case 2 can detect the miss and retry once.
+    // @Volatile because registerProvider writes without holding the mutex; @Volatile ensures those
+    // writes are visible to subsequent mutex-protected reads in handleForegroundTransition().
+    @Volatile private var refreshJob: Clock.Cancellable? = null
+
+    // @Volatile for the same reason as refreshJob: registerProvider clears without holding mutex.
+    @Volatile private var refreshAtWallClockMs: Long? = null
+
+    // Set by the timer callback while holding [mutex] before refresh work begins. This prevents a
```
evan-masseau: Nit: this `[mutex]` uses KDoc bracket-link syntax, but it's inside a plain `//` comment so the link is dead — `[...]` only resolves to a clickable symbol inside a `/** */` block. Either drop the brackets here, or (see my review summary) promote the method-level contract comments to KDoc where links like this actually resolve. The private-field rationale comments are correctly left as `//`.

evan-masseau (approved):
Approving — the proactive-refresh state machine is carefully built and the test coverage maps cleanly to every foreground case and the cancel/fire race. Nice work.

Two non-blocking comments left inline:

1. **Naming** (`currentTokenInternal`): the name undersells its dual role (return-cached vs. force-fresh-fetch). Consider `getOrFetchToken` or splitting it.

2. **KDoc vs `//` comments**: the private-field rationale comments are correctly left as `//` — they explain *why* a concurrency primitive was chosen, for a maintainer, on private members Dokka skips. No need to promote those. But the **method-level** `//` blocks on `scheduleRefresh` / `performScheduledRefresh` / `handleForegroundTransition` document behavior + contracts and are the real KDoc candidates — promoting them would also make the `[symbol]` links from `currentTokenInternal`'s KDoc resolve to hover docs. Also flagged one dead `[mutex]` bracket-link sitting in a plain `//` comment (line 63).

None of these block merge.

PR #478, opened 2026-06-10
feat(core): AuthTokenManager connectivity-driven refresh retry (MAGE-684)

amber-klaviyo:
## Summary

- When the proactive refresh path fails with a network-classified error (`IOException`, `UnknownHostException`, `SocketTimeoutException`, `ConnectException`), a single `connectivityWaitJob` is armed on the manager's existing `CoroutineScope` that suspends until the next \"connectivity available\" notification from `Registry.networkMonitor`, then re-triggers `performScheduledRefresh`.
- At most one job is active at a time — rapid connectivity flap is handled via a monotonic generation counter; cancelling the old job before arming a new one prevents observer accumulation.
- `clearTokenState()` (MAGE-630) and `registerProvider()` cancel and clear the job as part of their existing teardown, replacing the `TODO (MAGE-684)` stub left by the prior ticket.
- Non-network failures (HTTP errors, validation failures, host-side bugs) do not arm the connectivity retry.

## Test plan

- [ ] `connectivity retry fires after network comes back online after IOException` — happy path end-to-end
- [ ] `connectivity retry fires after UnknownHostException / SocketTimeoutException / ConnectException` — all classified exception types
- [ ] `connectivity wait job is not armed when connectivity notification is offline` — only `isConnected=true` fires the retry
- [ ] `rapid flap cancels existing connectivity wait job before arming new one` — at-most-one job invariant; generation counter correctness
- [ ] `registering new provider while connectivity wait is armed cancels the job`
- [ ] `non-network exception does not arm connectivity wait job` — `RuntimeException`, `ValidationFailed`
- [ ] `clearTokenState cancels and clears connectivity wait job` — simulates logout / `resetProfile`
- [ ] All prior `KlaviyoAuthTokenManagerRefreshTest` and `KlaviyoAuthTokenManagerTest` tests still pass (187 total)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication token refresh reliability by automatically retrying proactive scheduled refresh after transient network/offline failures recover, while preventing retries during stale session/reset scenarios and handling rapid connectivity changes safely.

* **Tests**
  * Added a connectivity-focused test suite covering multiple network exception types, offline vs online behavior (including immediate retry when already connected), at-most-one retry job behavior, and correct cancellation when refreshing state is cleared or a new provider is registered.

### DIFF
--- a/sdk/core/src/main/java/com/klaviyo/core/auth/KlaviyoAuthTokenManager.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/auth/KlaviyoAuthTokenManager.kt
@@ -4,8 +4,13 @@ import com.klaviyo.core.Registry
 import com.klaviyo.core.config.Clock
 import com.klaviyo.core.lifecycle.ActivityEvent
 import com.klaviyo.core.lifecycle.LifecycleMonitor
+import com.klaviyo.core.networking.NetworkObserver
 import com.klaviyo.core.safeLaunch
 import com.klaviyo.core.utils.takeIf
+import java.io.IOException
+import java.net.ConnectException
+import java.net.SocketTimeoutException
+import java.net.UnknownHostException
 import java.util.concurrent.CopyOnWriteArrayList
 import java.util.concurrent.atomic.AtomicLong
 import kotlin.coroutines.resume
@@ -13,6 +18,7 @@ import kotlin.coroutines.resumeWithException
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.async
 import kotlinx.coroutines.currentCoroutineContext
@@ -89,6 +95,19 @@ internal class KlaviyoAuthTokenManager(
     // (established SDK observer-collection pattern, matches StateChangeObserver, ActivityObserver).
     private val refreshObservers = CopyOnWriteArrayList<TokenRefreshObserver>()
 
+    // A pending coroutine that waits for connectivity to be restored before retrying
+    // performScheduledRefresh. At most one is active at a time; any existing job is cancelled before
+    // arming a new one (prevents job accumulation on rapid flap). Cleared once the wait resolves or
+    // is cancelled. Guarded by mutex writes inside clearTokenState; @Volatile for reads in
+    // armConnectivityWaitJob (outside the mutex) so cancel() is visible to the newly-starting job.
+    // Internal (not private) so tests in this module can inspect job state without reflection.
+    @Volatile internal var connectivityWaitJob: Job? = null
+
+    // Monotonic counter bumped each time armConnectivityWaitJob() arms a new job. Used by the
+    // job's finally block to detect whether a newer arm replaced it; if so the old job must not
+    // null out the new job's reference. AtomicLong for the same reason as refreshGeneration.
+    private val connectivityWaitGeneration = AtomicLong(0L)
+
     override fun registerProvider(provider: AuthTokenProvider) {
         // Cancel any in-flight fetch for the old provider before swapping.
         // Two complementary guards prevent a stale token from reaching the cache:
@@ -103,6 +122,8 @@ internal class KlaviyoAuthTokenManager(
         refreshJob = null
         refreshAtWallClockMs = null
         refreshTimerFired = false
+        connectivityWaitJob?.cancel()
+        connectivityWaitJob = null
         refreshGeneration.incrementAndGet()
         // Advance profileGeneration so any pending clearTokenState(expectedGeneration) from a
         // prior resetProfile() sees the generation mismatch and skips, preserving this new
@@ -150,7 +171,8 @@ internal class KlaviyoAuthTokenManager(
             refreshAtWallClockMs = null
             refreshTimerFired = false
             refreshGeneration.incrementAndGet()
-            // TODO (MAGE-684): connectivityWaitJob?.cancel(); connectivityWaitJob = null
+            connectivityWaitJob?.cancel()
+            connectivityWaitJob = null
             cachedToken = null
             profileGeneration.incrementAndGet()
             // Clear the reset-pending flag so the next successful refresh (from a new or retained
@@ -386,6 +408,9 @@ internal class KlaviyoAuthTokenManager(
         } catch (e: Exception) {
             if (timerGeneration != null) clearFiredFlagForFailedRefresh(timerGeneration)
             Registry.log.warning("Proactive token refresh failed: ${e.javaClass.simpleName}", e)
+            if (isNetworkException(e)) {
+                armConnectivityWaitJob()
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


sdk/core/src/main/java/com/klaviyo/core/auth/KlaviyoAuthTokenManager.kt
```diff
@@ -388,6 +410,9 @@ internal class KlaviyoAuthTokenManager(
         } catch (e: Exception) {
             if (timerGeneration != null) clearFiredFlagForFailedRefresh(timerGeneration)
             Registry.log.warning("Proactive token refresh failed: ${e.javaClass.simpleName}", e)
+            if (isNetworkException(e)) {
```
evan-masseau: We do also have a method in `NetworkMonitor` for checking `isNetworkConnected` that we could use as a pre-flight check instead/in addition to a reactive catch here. @ab1470 did you do a pre-flight connectivity check on iOS?

ab1470: no, on iOS we only respond reactively. If there's no connection we should get a `URLError` quickly, at which point we'll kick off a listener for network connectivity and retry once we get a ping that the network is back

sdk/core/src/main/java/com/klaviyo/core/auth/KlaviyoAuthTokenManager.kt
```diff
@@ -388,6 +410,9 @@ internal class KlaviyoAuthTokenManager(
...
+                armConnectivityWaitJob()
```
evan-masseau: Failure-path cleanup needed here — bundling the related concurrency findings since they share a root cause:

- **Logout/reset race (the important one):** this `catch` arms `connectivityWaitJob` with none of the stale-guards the success branch has — no `provider == null` / `profileResetPending` check. A timer refresh that fails with a network error *concurrently* with `clearTokenState()` / `registerProvider()` can assign the job **after** teardown nulled it, leaving a zombie that re-fetches and notifies observers post-logout. Worth adding: `Registry.dispatcher` is `Dispatchers.IO` (multi-threaded), so this is a true parallel race, not just an interleaving — and the new tests run on a single-threaded `TestDispatcher`, so the suite can't surface it. This is exactly the post-reset notification that MAGE-630's `profileResetPending` machinery was built to prevent; the failure path just doesn't honor it.
- **Serialize the transitions:** `connectivityWaitJob` / `connectivityWaitGeneration` are mutated under the mutex in `clearTokenState` but outside it in `registerProvider` and here. +1 CodeRabbit's suggestion to wrap all transitions in one dedicated lock so the at-most-one-job invariant actually holds.
- **+1 CodeRabbit** on the two related items: the observer registration/cancellation race in `armConnectivityWaitJob` (register before installing `invokeOnCancellation`, re-check `isActive` after), and the missing `currentCoroutineContext().ensureActive()` before the post-connectivity `performScheduledRefresh()`.

None of these block approval, but please get them in before release.

sdk/core/src/main/java/com/klaviyo/core/auth/KlaviyoAuthTokenManager.kt
```diff
@@ -422,6 +447,73 @@ internal class KlaviyoAuthTokenManager(
         }
     }
 
+    /**
+     * Returns `true` for exceptions that indicate genuine offline conditions:
+     * [IOException] and its subtypes [UnknownHostException], [SocketTimeoutException], and
+     * [ConnectException]. HTTP errors, validation failures, and server-side bugs return `false`
+     * and must not trigger a connectivity retry (they won't resolve by waiting for the network).
+     */
+    private fun isNetworkException(e: Exception): Boolean =
```
evan-masseau, line 521: Minor, and I don't think anyone else caught it: `UnknownHostException`, `SocketTimeoutException`, and `ConnectException` are all subclasses of `IOException`, so this whole expression collapses to `e is IOException`. Harmless and arguably self-documenting, but the three explicit disjuncts are dead checks — either reduce to `e is IOException`, or add a one-line comment that the enumerated list is intentional for readability.

sdk/core/src/test/java/com/klaviyo/core/auth/KlaviyoAuthTokenManagerConnectivityTest.kt
```diff
@@ -0,0 +1,524 @@
...
+class KlaviyoAuthTokenManagerConnectivityTest : BaseTest() {
+
+    companion object {
+        private const val NOW_SECONDS = TIME / 1000L
+        private const val IAT_SECONDS = NOW_SECONDS - 60
+        private const val EXP_SECONDS = NOW_SECONDS + 3600
+    }
+
+    private lateinit var fakeNetworkMonitor: FakeNetworkMonitor
+    private val observerSlot = slot<ActivityObserver>()
```
evan-masseau: `observerSlot` is captured by the `onActivityEvent` mock just below but never read by any test. Drop the field and use `any()` in the capture. (CodeRabbit's CLI pass flagged this; it isn't in the inline threads.)

sdk/core/src/test/java/com/klaviyo/core/auth/KlaviyoAuthTokenManagerConnectivityTest.kt
```diff
@@ -0,0 +1,524 @@
...
+            )
+        )
+        val manager = KlaviyoAuthTokenManager()
+
+        manager.registerProvider(provider)
+        dispatcher.scheduler.advanceUntilIdle()
+        assertEquals("initial eager fetch", 1, provider.callCount)
+
+        // Fire the scheduled refresh — it fails with a network error
+        val timerTask = staticClock.scheduledTasks.first()
```
evan-masseau: DRY: this `staticClock.scheduledTasks.first()` + `staticClock.execute(...)` pattern repeats ~11 times across the file — worth an `executeScheduledRefresh()` helper. (Same spirit as the note above about parameterizing the four exception-variant tests.)

evan-masseau (approved):
Approving — the connectivity-driven retry is a solid feature and the happy-path coverage is genuinely thorough. A few actionable items inline plus the existing bot threads; none block the approval, but please land the failure-path concurrency cluster before this rides up to release. I'm also endorsing CodeRabbit's remaining nits already on the threads (`!!` in test assertions, asserting log level over message text, parameterizing the exception-variant tests) — all worth doing, none blocking.

---

Stepping back, a broader concern I want on the record — framed as something to explore, **not** a blocker for this PR:

`KlaviyoAuthTokenManager` now coordinates state with **1 `Mutex`, ~8 `@Volatile` fields, 3 `AtomicLong` generation counters, and a `CopyOnWriteArrayList`**. Every `@Volatile` carries a paragraph explaining exactly which methods write it outside the lock — that documentation load is itself a signal we're near the edge of what's reasonable to reason about. The tell: three independent reviewers (Bugbot, CodeRabbit, and me) each landed on the same race cluster in the *newest* addition. Each feature (MAGE-630 reset, now MAGE-684 connectivity) bolts on another field threaded through both the synchronous public API and the suspend/mutex core, and the state cross-product keeps growing.

Root cause is one tension: the public API is synchronous (`registerProvider`, `invalidate(): Long`) but the core is suspend + mutex-guarded, so we reach for `@Volatile` + generation counters to mutate outside the lock. The `invalidate()` / `clearTokenState(expectedGeneration)` generation handshake exists *only* because those two halves can race today.

**Idea worth exploring** (just an idea): confine all mutable state to an internally single-threaded context — `Registry.dispatcher.limitedParallelism(1)` or an actor-style channel. Public methods enqueue; nothing interleaves by construction. That would let us drop most of the `@Volatile`s and likely all three generation counters (they exist to detect interleaving that could no longer happen). The cost is that synchronous methods become enqueue-and-return — but the real work is already async, and it removes the API shapes that exist only to manage the races.

I'm honestly not certain it nets out simpler, and I don't want to derail this PR to find out. But I am concerned about long-term maintainability as-is. **Proposal: a dedicated pre-release review of the JWT workstream focused specifically on simplifying the parallelism story.** Happy to prototype the limited-parallelism approach together and help guide the call — planting the flag now while it's fresh, before the surface area grows again.

PR #448, opened 2026-05-15
feat(push): add open_url action for push tap and action buttons

morrilltim:
## Summary

Adds `open_url` support across two surfaces in the Android SDK:

- **Push tap:** when the FCM payload contains a `web_url` field (http/https), tapping the notification opens the URL in the default browser instead of the host app
- **Action buttons:** new `ActionButton.OpenUrl` sealed-class variant handles `"action": "open_url"` in the button payload, also routing to the default browser

Action buttons for `deep_link` and `open_app` already shipped in v4.4.0. This PR is purely additive — existing deep-link behavior is unchanged.

Per [PUSH-630](https://linear.app/klaviyo/issue/PUSH-630), the `web_url` field is a **separate top-level key** in the push payload; presence of a valid http(s) `web_url` is the SDK's discriminator. `open_action` is not used.

Both surfaces route through a new internal `KlaviyoTrampolineActivity` (`exported=false`, translucent, `noHistory=true`, `excludeFromRecents=true`) which fires `Klaviyo.handlePush(intent)` for `$opened_push` tracking and notification dismissal, then dispatches the browser intent. Without it, the browser would swallow the intent and the host app would never see the tap.

> [!NOTE]
> **Dual-field precedence (edge case):** if a payload contains both `url` (deep link) and `web_url`, the SDK routes to the deep link and logs a warning. This should not happen in normal customer flows — the fender composer ([`PushDeepLinkingSection.tsx`](https://github.com/klaviyo/fender/blob/master/client/shared/mobile-content-builder/src/push/components/editorPanel/PushDeepLinkingSection.tsx)) uses a single `openAction` SelectBox with mutually-exclusive conditional rendering, so a marketer cannot populate both fields on the same variation. The rule only matters for direct Send-Push API calls or test tooling like `fcm-push-cli`. iOS counterpart aligned in [klaviyo-swift-sdk#615](https://github.com/klaviyo/klaviyo-swift-sdk/pull/615).

A new `makeBrowserIntent(uri)` helper on `DeepLinking` builds `Intent.ACTION_VIEW` with `CATEGORY_BROWSABLE` + `FLAG_ACTIVITY_NEW_TASK` and no `setPackage`, so the OS routes to the default browser.

[...]

PR #493, opened 2026-06-23
feat(push): KlaviyoTrampolineActivity with opt-in automatic push tracking [MAGE-767]

glenn-klaviyo:
# Description

Routes Klaviyo notification taps (body, `deep_link`, `open_app`) through the transparent `KlaviyoTrampolineActivity` when a host opts in via the `com.klaviyo.automatic_push_tracking` manifest flag, so the SDK tracks `$opened_push` itself — removing the need for `Klaviyo.handlePush(intent)` in every host Activity. Builds on the trampoline from #448; flag off (default) is unchanged.

## Due Diligence

  - Part of the Automatic Push Open Tracking milestone (MAGE-765 telemetry, MAGE-768 dedup). Targets the `feat/auto-push-tracking` integration branch, not `master`.

## Changelog / Code Overview
- **core:** add `Config.getManifestBoolean` (+ `Context.getManifestBoolean` extension) and the `Constants.AUTOMATIC_PUSH_TRACKING` manifest key, re-exposed as `KlaviyoPushService.METADATA_AUTOMATIC_PUSH_TRACKING`.
- **push-fcm:** extend `KlaviyoTrampolineActivity` with a shared `forDestination` factory, `onNewIntent` (singleTask re-entry), and generalized dispatch — browser URL (unchanged, #448) / deep-link-via-`ACTION_VIEW` when no `DeepLinkHandler` is registered / launch fallback. `KlaviyoNotification` reads the flag once per build and routes body + action-button intents through the trampoline when enabled. Manifest hardened with `singleTask` + `taskAffinity=""` + `excludeFromRecents`.
- Flag **off** → body/`deep_link`/`open_app` intents byte-identical to today; `open_url` always routes through the trampoline regardless (per #448).

**Out of scope (sibling tickets):** the dedup guard (MAGE-768) and any standalone telemetry (MAGE-765). Until MAGE-768 lands, an opt-in host with a leftover manual `handlePush` will double-count opens — expected and acceptable while unreleased.

## Test Plan
- Unit: `:sdk:core` + `:sdk:push-fcm` `testDebugUnitTest` (JDK 17) green; ktlint clean. Covers the trampoline dispatch matrix (handler registered/unregistered, deep link present/absent/unresolvable, non-Klaviyo), flag-on/off notification routing with extras parity, and `getManifestBoolean` across both SDK API paths.
- Manual (pending): sample app on emulator — flag on: body + action-button taps from foreground/background/cold start track the open and land on the correct destination; flag off: `open_url` still tracks and opens the browser.

## Related Issues/Tickets
- MAGE-767 — https://linear.app/klaviyo/issue/MAGE-767

## Summary by CodeRabbit

* **New Features**
  * Added an optional **automatic push tracking** mode that routes notification body taps and action buttons through the unified in-app trampoline flow, controlled via manifest metadata.
* **Bug Fixes**
  * Improved trampoline routing for deep links, unresolved deep links, and standard launches, including more reliable back-stack behavior.
  * Reduced sensitive/verbose logging when an intent can’t be resolved.
* **Tests**
  * Expanded automated coverage for body taps, deep-link actions, open-app actions, and browser links in both tracking modes.
* **Documentation**
  * Added support for reading boolean manifest metadata values with a default.

### DIFF
--- a/sdk/core/src/main/java/com/klaviyo/core/Constants.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/Constants.kt
@@ -31,11 +31,9 @@ object Constants {
 
     /**
      * Manifest `<meta-data>` key a host app sets to opt into automatic push open tracking.
-     * When enabled, Klaviyo notification taps route through `KlaviyoTrampolineActivity`, which
-     * calls `Klaviyo.handlePush` itself — removing the need for `handlePush` in host Activities.
      *
      * Lives in core (not push-fcm) because telemetry's push token request must read it, and core
-     * cannot depend on push-fcm. Re-exposed as `KlaviyoPushService.METADATA_AUTOMATIC_PUSH_TRACKING`.
+     * cannot depend on push-fcm.
      */
     const val AUTOMATIC_PUSH_TRACKING = PACKAGE_PREFIX + "automatic_push_tracking"
 
--- a/sdk/core/src/main/java/com/klaviyo/core/Constants.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/Constants.kt
@@ -30,6 +30,16 @@ object Constants {
     const val NOTIFICATION_TAG_EXTRA = INTERNAL_PREFIX + "notification_tag"
 
     /**
+     * Manifest `<meta-data>` key a host app sets to opt into automatic push open tracking.
+     * When enabled, Klaviyo notification taps route through `KlaviyoTrampolineActivity`, which
+     * calls `Klaviyo.handlePush` itself — removing the need for `handlePush` in host Activities.
+     *
+     * Lives in core (not push-fcm) because telemetry's push token request must read it, and core
+     * cannot depend on push-fcm. Re-exposed as `KlaviyoPushService.METADATA_AUTOMATIC_PUSH_TRACKING`.
+     */
+    const val AUTOMATIC_PUSH_TRACKING = PACKAGE_PREFIX + "automatic_push_tracking"
+
+    /**
      * Fixed notification ID used in all notify/cancel calls.
      * Notifications are uniquely identified by their string tag, not this ID.
      */
--- a/sdk/core/src/main/java/com/klaviyo/core/utils/IntentExtension.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/utils/IntentExtension.kt
@@ -12,8 +12,7 @@ fun Intent.startActivityIfResolved(context: Context) {
     if (activityResolved(context)) {
         context.startActivity(this)
     } else {
-        // Log only the action/component, never the whole intent — its `data` and extras can carry
-        // payload-derived URLs (deep links, browser URLs) with customer identifiers or tokens.
+        // Avoid logging the full intent — data/extras can carry deep-link URLs with PII.
         Registry.log.error(
             "No activity found to handle intent (action=$action, component=$component)"
         )
--- a/sdk/core/src/main/java/com/klaviyo/core/utils/IntentExtension.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/utils/IntentExtension.kt
@@ -12,7 +12,11 @@ fun Intent.startActivityIfResolved(context: Context) {
     if (activityResolved(context)) {
         context.startActivity(this)
     } else {
-        Registry.log.error("No activity found to handle intent: $this")
+        // Log only the action/component, never the whole intent — its `data` and extras can carry
+        // payload-derived URLs (deep links, browser URLs) with customer identifiers or tokens.
+        Registry.log.error(
+            "No activity found to handle intent (action=$action, component=$component)"
+        )
     }
 }
 
--- a/sdk/fixtures/src/main/java/com/klaviyo/fixtures/MockIntent.kt
+++ b/sdk/fixtures/src/main/java/com/klaviyo/fixtures/MockIntent.kt
@@ -2,6 +2,7 @@ package com.klaviyo.fixtures
 
 import android.annotation.SuppressLint
 import android.app.PendingIntent
+import android.content.ComponentName
 import android.content.Intent
 import android.net.Uri
 import android.os.Bundle
@@ -66,9 +67,22 @@ data class MockIntent(
             every { anyConstructed<Intent>().flags } answers {
                 if (flags.isCaptured) flags.captured else 0
             }
-            // No setComponent is used in production (we set class via setClassName); stub the getter
-            // so reads (e.g. error logging) don't throw on the constructed mock.
-            every { anyConstructed<Intent>().component } returns null
+            // Mirror Android: setClassName(pkg, cls) makes getComponent() return the corresponding
+            // ComponentName. When only setPackage was used (e.g. deep-link intents), component is null.
+            every { anyConstructed<Intent>().component } answers {
+                if (className.isCaptured) {
+                    val capturedPkg = if (packageName.isCaptured) packageName.captured else ""
+                    val capturedCls = className.captured
+                    mockk<ComponentName>(relaxed = true) {
+                        every { getPackageName() } returns capturedPkg
+                        every { getClassName() } returns capturedCls
+                        every { flattenToShortString() } returns "$capturedPkg/$capturedCls"
+                        every { toString() } returns "ComponentInfo{$capturedPkg/$capturedCls}"
+                    }
+                } else {
+                    null
+                }
+            }
 
             return mockIntent
         }
--- a/sdk/fixtures/src/main/java/com/klaviyo/fixtures/MockIntent.kt
+++ b/sdk/fixtures/src/main/java/com/klaviyo/fixtures/MockIntent.kt
@@ -66,6 +66,9 @@ data class MockIntent(
             every { anyConstructed<Intent>().flags } answers {
                 if (flags.isCaptured) flags.captured else 0
             }
+            // No setComponent is used in production (we set class via setClassName); stub the getter
+            // so reads (e.g. error logging) don't throw on the constructed mock.
+            every { anyConstructed<Intent>().component } returns null
 
             return mockIntent
         }
--- a/sdk/push-fcm/src/main/AndroidManifest.xml
+++ b/sdk/push-fcm/src/main/AndroidManifest.xml
@@ -13,7 +13,9 @@
             android:name="com.klaviyo.pushFcm.KlaviyoTrampolineActivity"
             android:exported="false"
             android:excludeFromRecents="true"
+            android:launchMode="singleTask"
             android:noHistory="true"
+            android:taskAffinity=""
             android:theme="@android:style/Theme.Translucent.NoTitleBar" />
     </application>
 </manifest>
--- a/sdk/push-fcm/src/main/AndroidManifest.xml
+++ b/sdk/push-fcm/src/main/AndroidManifest.xml
@@ -9,6 +9,11 @@
                 <action android:name="com.google.firebase.MESSAGING_EVENT" />
             </intent-filter>
         </service>
+        <!--
+            Trampoline activity that tracks the push open then forwards the intent. Configured to be
+            invisible and transient: translucent theme (no UI), noHistory + excludeFromRecents (never
+            in the back stack or recents), taskAffinity="" + singleTask (no task of its own).
+        -->
         <activity
             android:name="com.klaviyo.pushFcm.KlaviyoTrampolineActivity"
             android:exported="false"
--- a/sdk/push-fcm/src/main/AndroidManifest.xml
+++ b/sdk/push-fcm/src/main/AndroidManifest.xml
@@ -10,9 +10,14 @@
             </intent-filter>
         </service>
         <!--
-            Trampoline activity that tracks the push open then forwards the intent. Configured to be
-            invisible and transient: translucent theme (no UI), noHistory + excludeFromRecents (never
-            in the back stack or recents), taskAffinity="" + singleTask (no task of its own).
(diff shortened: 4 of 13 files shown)


sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
```diff
@@ -257,11 +264,18 @@ class KlaviyoNotification(private val message: RemoteMessage) {
...
+            // intent data) so it calls handlePush; otherwise target the host directly as before.
+            val intent = if (autoTracking) {
+                KlaviyoTrampolineActivity.forDestination(context, deepLink)
+            } else {
+                makeResolvedDeepLinkIntent(
+                    context,
+                    deepLink,
+                    "Push message contained unsupported deep link: $deepLink"
+                )
+            }
```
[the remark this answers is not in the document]
glenn-klaviyo: Reviewed with Claude:
> Working as intended — not a regression. You're right that the auto-tracking path keeps the deep link on the intent even when no Activity resolves it, whereas makeResolvedDeepLinkIntent strips it. That difference is deliberate.

> The legacy coupling tied handler invocation to Activity resolvability as a side effect of reusing one intent builder for two unrelated questions ("does an Activity exist?" vs. "should we notify the registered handler?"). A DeepLinkHandler exists precisely to route links programmatically — independent of intent-filters — which is the norm for single-Activity Compose/Navigation apps. So delivering every deep link to a registered handler, resolvable or not, is the correct behavior; the legacy path silently dropped those links (handler skipped and only the launcher opened).

> There's no double-navigation risk: when a handler is registered, KlaviyoTrampolineActivity.startDestination deliberately takes the launcher-only branch and lets handlePush deliver the link to the handler. Activity resolvability is still honored in the no-handler branch, matching legacy exactly.

> Added a comment at the forDestination call sites documenting this so it's clear at the source.

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoTrampolineActivity.kt
```diff
@@ -78,16 +109,55 @@ internal class KlaviyoTrampolineActivity : Activity() {
...
+                        context,
+                        copyIntent = intent
+                    )
+                    if (viewIntent.activityResolved(context)) {
+                        Registry.log.verbose("Trampoline dispatching deep link via VIEW: $deepLink")
+                        viewIntent
+                    } else {
+                        Registry.log.error(
+                            "Trampoline could not resolve deep link $deepLink; falling back to launcher"
+                        )
```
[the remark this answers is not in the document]
glenn-klaviyo: Fixed in 72b5ed2e3. Dropped the URL/deep-link interpolation from the three dispatch logs, keeping the branch label so the traces stay debuggable (per the ticket). The ERROR-level one was the real concern since ERROR is production-visible; the VERBOSE ones are debug-only but sanitized for consistency.

-Claude

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoTrampolineActivity.kt
```diff
@@ -78,16 +109,55 @@ internal class KlaviyoTrampolineActivity : Activity() {
...
+                    } else {
+                        Registry.log.verbose("Trampoline dispatching launch intent")
+                    }
+                    DeepLinking.makeLaunchIntent(context, intent.extras)
+                }
+            }
+
+            destination?.apply {
+                addFlags(Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_SINGLE_TOP)
+            }?.startActivityIfResolved(context)
```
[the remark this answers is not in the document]
glenn-klaviyo: Fixed in 72b5ed2e3. Confirmed against the legacy path: the flag-off body deep-link intent ends up SINGLE_TOP | CLEAR_TOP (makeDeepLinkIntent sets SINGLE_TOP, makeOpenedIntent adds CLEAR_TOP), whereas the trampoline was forwarding only NEW_TASK | SINGLE_TOP. Added CLEAR_TOP to the forwarded destination to restore back-stack parity; NEW_TASK stays since we launch from the trampoline's own taskAffinity="" task. Note this also adds CLEAR_TOP to the action-button path (legacy buttons were NEW_TASK | SINGLE_TOP) — that is the canonical notification-nav combo and avoids duplicate stacked activities, so it's an intended minor improvement.

-Claude

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoTrampolineActivity.kt
```diff
@@ -78,16 +109,63 @@ internal class KlaviyoTrampolineActivity : Activity() {
 
         private fun dispatchDestination(intent: Intent, context: Context) {
             intent.getStringExtra(BROWSER_URL_EXTRA)?.let { url ->
+                Registry.log.verbose("Trampoline dispatching browser intent")
                 DeepLinking.makeBrowserIntent(url.toUri()).startActivityIfResolved(context)
```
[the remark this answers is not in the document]
glenn-klaviyo: Fixed in 306816755 — verified: startActivityIfResolved logged the full intent, and Intent.toString() includes the URL in data. Rather than inline a check in just the browser branch, I sanitized the shared core util (IntentExtension.startActivityIfResolved) to log only action+component. The browser branch was not the only leak — the deep-link forward in startDestination also routes through this util with data=deepLink and would leak identically, as do the analytics/forms callers. Fixing it once covers all of them with no duplicated resolve-check logic.

-Claude

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
```diff
@@ -166,8 +166,15 @@ class KlaviyoNotification(private val message: RemoteMessage) {
         notificationTag: String
     ): NotificationCompat.Builder {
         val requestCodeBase = generateId()
+        // When the host opts into automatic push tracking, body/deep_link/open_app taps route
+        // through KlaviyoTrampolineActivity so it can call handlePush itself. open_url already
+        // routes through the trampoline regardless (the browser would otherwise swallow the intent).
+        val autoTracking = Registry.config.getManifestBoolean(
+            KlaviyoPushService.METADATA_AUTOMATIC_PUSH_TRACKING,
+            false
+        )
```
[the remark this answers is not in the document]
glenn-klaviyo: Fixed in 935c88a41 — verified the claim. buildNotification read the flag via Registry.config.getManifestBoolean, which returns the default until Klaviyo.initialize runs, so a host that initializes from an Activity (not Application) and receives a push while backgrounded would build the notification pre-init and silently fall back to legacy routing despite opting in.

The flag is static manifest data readable from any Context, and buildNotification already has a valid one, so it now reads via context.getManifestBoolean(...) — same pattern KlaviyoRemoteMessage already uses for the icon/color meta-data. Works regardless of init state.

-Claude

sdk/fixtures/src/main/java/com/klaviyo/fixtures/MockIntent.kt
```diff
@@ -66,6 +66,9 @@ data class MockIntent(
             every { anyConstructed<Intent>().flags } answers {
                 if (flags.isCaptured) flags.captured else 0
             }
+            // No setComponent is used in production (we set class via setClassName); stub the getter
+            // so reads (e.g. error logging) don't throw on the constructed mock.
+            every { anyConstructed<Intent>().component } returns null
```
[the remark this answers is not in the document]
glenn-klaviyo: Fixed in 935c88a41. Agreed — setClassName(pkg, cls) makes getComponent() non-null in real Android, so the null stub diverged. MockIntent now mirrors that: when setClassName was captured it returns a ComponentName reflecting the package/class; when only setPackage was used (e.g. makeDeepLinkIntent) component stays null, matching Android. Analytics/core/push-fcm suites green.

-Claude

sdk/core/src/main/java/com/klaviyo/core/utils/IntentExtension.kt
```diff
@@ -12,7 +12,11 @@ fun Intent.startActivityIfResolved(context: Context) {
     if (activityResolved(context)) {
         context.startActivity(this)
     } else {
-        Registry.log.error("No activity found to handle intent: $this")
+        // Log only the action/component, never the whole intent — its `data` and extras can carry
+        // payload-derived URLs (deep links, browser URLs) with customer identifiers or tokens.
```
ab1470:
```suggestion
```
Claude can be so verbose with code comments. I appreciate that it explains the reasoning behind its changes but I feel like this isn't important or useful to include in production code

sdk/core/src/main/java/com/klaviyo/core/Constants.kt
```diff
@@ -29,6 +29,16 @@ object Constants {
     private const val INTERNAL_PREFIX = "_klaviyo."
     const val NOTIFICATION_TAG_EXTRA = INTERNAL_PREFIX + "notification_tag"
 
+    /**
+     * Manifest `<meta-data>` key a host app sets to opt into automatic push open tracking.
+     * When enabled, Klaviyo notification taps route through `KlaviyoTrampolineActivity`, which
+     * calls `Klaviyo.handlePush` itself — removing the need for `handlePush` in host Activities.
+     *
+     * Lives in core (not push-fcm) because telemetry's push token request must read it, and core
+     * cannot depend on push-fcm. Re-exposed as `KlaviyoPushService.METADATA_AUTOMATIC_PUSH_TRACKING`.
```
ab1470:
```suggestion
`KlaviyoPushService.METADATA_AUTOMATIC_PUSH_TRACKING`.
```
this seems like an implementation detail that can be removed

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
```diff
@@ -166,8 +167,20 @@ class KlaviyoNotification(private val message: RemoteMessage) {
         notificationTag: String
     ): NotificationCompat.Builder {
         val requestCodeBase = generateId()
+        // When the host opts into automatic push tracking, body/deep_link/open_app taps route
+        // through KlaviyoTrampolineActivity so it can call handlePush itself. open_url already
+        // routes through the trampoline regardless (the browser would otherwise swallow the intent).
+        //
+        // Read the manifest flag from the build-time [context] rather than Registry.config, which
```
ab1470: > [context]

why the `[` `]`?

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoPushService.kt
```diff
@@ -21,6 +22,19 @@ open class KlaviyoPushService : FirebaseMessagingService() {
...
+        /**
+         * Manifest `<meta-data>` key to opt into automatic push open tracking. When set, Klaviyo
+         * notification taps route through [KlaviyoTrampolineActivity], which tracks the open itself
+         * so the host app no longer needs to call `Klaviyo.handlePush(intent)` in its Activities:
+         *
+         * ```xml
+         * <meta-data
+         *     android:name="com.klaviyo.automatic_push_tracking"
+         *     android:value="true" />
+         * ```
```
ab1470: this code snippet is a little confusing to me. What are we supposed to do with it? Does it need to be in the documentation for this property?

sdk/push-fcm/src/main/AndroidManifest.xml
```diff
@@ -13,7 +13,9 @@
             android:name="com.klaviyo.pushFcm.KlaviyoTrampolineActivity"
             android:exported="false"
             android:excludeFromRecents="true"
+            android:launchMode="singleTask"
             android:noHistory="true"
+            android:taskAffinity=""
```
ab1470, line 28: what are these for?

ab1470 (approved):
Just some minor things, but LGTM. I don't have a full grasp of Android and Trampoline Activities though so my qualifications to review this are limited

sdk/push-fcm/src/main/AndroidManifest.xml
glenn-klaviyo: these attributes make the trampoline invisible essentially. We want it to be invisible for when a user hits their on device back button, we don't want this activity appear in that stack. It also prevents a blank screen flash between tapping the notification and the destination loading.

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
glenn-klaviyo: very valid, claude being claude... I'll clean up these comments

sdk/core/src/main/java/com/klaviyo/core/utils/IntentExtension.kt
glenn-klaviyo: I agree. trying to find the line between leaving enough comments in there so claude has an easier time understanding things in the future but I realize this can be quite verbose. I'll clean this up. Thanks!

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoPushService.kt
glenn-klaviyo: The snippet is the actual manifest entry an integrating app adds to opt in to the automatic push tracking. On a second glance I suppose it doesn't really belong here. It will be mentioned in the README / docs in a later PR

PR #495, opened 2026-06-29
feat(push): in-memory dedup guard for auto-tracked push opens (MAGE-768)

glenn-klaviyo:
# Description
In-memory dedup guard so a single notification tap tracks exactly one `Opened Push`, even when a host opts into automatic push tracking (the `KlaviyoTrampolineActivity` from #493) but leaves a manual `Klaviyo.handlePush(intent)` call in an Activity. The trampoline tracks first, then forwards the destination intent (carrying the same Klaviyo extras) to the host, whose manual call would otherwise track a second event — this guard short-circuits that second call. Safety net that makes a future default-on flip safe for existing integrators.

## Due Diligence

> Stacked on #493 (`feat/mage-767-trampoline`). Review/merge that first; base will retarget to the default branch once it lands.

## Changelog / Code Overview
- **`sdk/core/.../utils/BoundedIdSet.kt`** (new) — thread-safe, fixed-capacity (default 256) set with FIFO eviction and an atomic `markOnce(id): Boolean` (combined check-and-insert). Duplicate inserts are a no-op that does not advance eviction order.
- **`sdk/core/.../Constants.kt`** — two internal-prefixed intent extras: `NOTIFICATION_AUTO_TRACKED_EXTRA` (marks an intent as routed through the auto-tracking trampoline) and `NOTIFICATION_UID_EXTRA` (SDK-generated per-notification fallback dedup key). Internal prefix keeps both out of analytics event payloads.
- **`sdk/push-fcm/.../KlaviyoTrampolineActivity.kt`** — `forDestination` (the auto-tracking-path intent factory) stamps the auto-tracked marker and a fresh per-notification UUID. `forBrowserUrl` (`open_url`) is intentionally left unmarked — it always trampolines regardless of the flag and the host never receives a forwarded `open_url` tap, so there is nothing to deduplicate.
- **`sdk/analytics/.../Klaviyo.kt#handlePush`** — top-of-function guard, scoped to auto-tracked intents only. Dedup key prefers the per-delivery `tm` from the `_k` payload, falling back to the generated per-notification UUID when `tm` is absent. Both keys ride forward onto the host's destination intent, so the trampoline's call and the host's manual call resolve to the same key. Manual-only integrations carry no marker, so their behavior is unchanged.

### Design notes
- **Flag-off is byte-identical.** The guard engages only when the auto-tracked marker is present, which only the trampoline (auto path) stamps. Manual-only integrations never carry it.
- **Dedup key.** `tm` is a per-delivery ULID present on campaign sends and is the same value the backend associates with the open. When absent (local notifications, previews, non-campaign sends) the generated per-notification UUID is used. The raw `_k` payload is deliberately not used as a fallback — minus `tm` it is per-message metadata shared across deliveries, so it would collide and drop legitimate opens.

## Test Plan
- `BoundedIdSetTest` (core): insert/contains round-trip, `markOnce` true-then-false, FIFO eviction removes oldest, duplicate insert does not advance eviction order, evicted IDs re-insertable, capacity bound, constructor rejects non-positive capacity, concurrent-access safety.
- `KlaviyoTest` (analytics): one event when an auto-tracked delivery is handled twice (distinct intents + same intent object); distinct deliveries tracked independently; `tm`-less dedup via the generated UID with distinct notifications independent; malformed `_k` falls back gracefully; manual (non-auto-tracked) intents never deduped.
- Manual: sample/test app with auto-tracking on + a leftover manual `handlePush` — single `Opened Push` per tap; distinct notifications each track.

## Related Issues/Tickets
- MAGE-768
- Stacked on #493 (MAGE-767)

## Summary by CodeRabbit

* **New Features**
  * Added bounded in-process deduplication for auto-tracked push notification opens to prevent repeated processing.
  * Updated destination trampoline intents to include a per-notification fallback ID to support consistent “handled once” behavior.

* **Bug Fixes**
  * Prevented duplicate `$opened_push` events for the same delivery, while still performing deep-link handling and dismissal for the first occurrence.
  * Gracefully falls back when delivery metadata is missing or malformed, and only dedups when a dedup key is available.

* **Tests / Quality**
  * Expanded coverage for deduplication, fallback behavior, and bounded set eviction/concurrency.

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
@@ -49,17 +49,8 @@ object Klaviyo {
     private val preInitQueue: Queue<Operation<Unit>> = LinkedList()
 
     /**
-     * In-memory dedup guard over push *delivery* IDs already auto-tracked by [handlePush].
-     *
-     * Only auto-tracking-path intents engage this guard: the trampoline stamps
-     * [Constants.NOTIFICATION_AUTO_TRACKED_EXTRA] on its targeted intents and forwards the
-     * destination intent (carrying that marker and the `_k` payload) to the host. The trampoline's
-     * own `handlePush` marks the delivery first; a host that also left a manual `handlePush` call in
-     * place then short-circuits instead of tracking a second `$opened_push` for the same tap. It
-     * also dedupes trampoline `singleTask` re-entry with the same intent.
-     *
-     * Manual-only integrations (auto-tracking disabled) never carry the marker, so they never engage
-     * the guard and behave exactly as before it existed.
+     * Push delivery IDs already auto-tracked within this process, so a single tap records one
+     * `$opened_push`. See [handlePush] for how entries are matched and added.
      */
     private val handledPushDeliveries = BoundedIdSet()
 
@@ -346,55 +337,58 @@ object Klaviyo {
      */
     @JvmStatic
     fun handlePush(intent: Intent?): Klaviyo {
-        // Dedup guard (top of function), scoped to the automatic-tracking path only: a push delivery
-        // we've already auto-tracked means this is a re-delivery of the same tap — a host's leftover
-        // manual handlePush after the trampoline already tracked, or singleTask re-entry.
-        // Short-circuit fully: no event, no dismissal, no deep link dispatch. The marker is absent on
-        // manual-only integrations, so their behavior is unchanged. Unseen delivery → atomically mark
-        // it and proceed.
-        if (intent.isAutoTrackedOpen) {
-            intent.pushDeliveryId?.let { deliveryId ->
-                if (!handledPushDeliveries.markOnce(deliveryId)) {
-                    Registry.log.verbose(
-                        "Ignoring duplicate auto-tracked push open for delivery $deliveryId"
-                    )
-                    return this
-                }
+        // Dedup guard for the automatic-tracking path. The trampoline calls handlePush and then
+        // forwards the same intent to the host, so a leftover manual handlePush call (or singleTask
+        // re-entry) would otherwise track a second open for one tap. Record each delivery the first
+        // time it's seen; if it's already recorded, drop this call entirely — no event, dismissal, or
+        // deep link. Manual-only integrations carry no auto-tracked marker, so they're unaffected.
+        val deliveryId = intent.pushDeliveryId
+        if (
+            intent.isAutoTrackedOpen &&
+            deliveryId != null &&
+            !handledPushDeliveries.markOnce(deliveryId)
+        ) {
+            Registry.log.verbose("Ignoring duplicate auto-tracked push open")
+            return this
+        }
+
+        if (intent == null || !intent.isKlaviyoNotificationIntent) {
+            Registry.log.verbose("Non-Klaviyo intent ignored")
+            return this
+        }
+
+        // Create and enqueue an $opened_push. safeApply(preInitQueue) buffers this for replay if
+        // handlePush runs before initialize(), and guards against unexpected failures.
+        safeApply(preInitQueue) {
+            val state = Registry.get<State>()
+            val event = Event(EventMetric.OPENED_PUSH)
+            event.appendKlaviyoExtras(intent)
+            state.pushToken?.let { event[EventKey.PUSH_TOKEN] = it }
+            // Not using Klaviyo.createEvent here to avoid nesting safeApply calls
+            state.createEvent(event, state.getAsProfile())
+        }
+
+        // Dismiss the notification if opened via an action button. Body taps auto-cancel via
+        // setAutoCancel(true) on the builder; action button taps don't (standard Android behavior).
+        safeApply {
+            val notificationTag = intent.getStringExtra(Constants.NOTIFICATION_TAG_EXTRA)
+            if (notificationTag != null) {
+                NotificationManagerCompat
+                    .from(Registry.config.applicationContext)
+                    .cancel(notificationTag, Constants.NOTIFICATION_ID)
+            }
+        }
+
+        // If the notification carries a deep link and a handler is registered, invoke it. Otherwise
+        // do nothing — the host already received the appropriate intent.
+        safeApply {
+            val deepLink = intent.data
+            if (deepLink != null && DeepLinking.isHandlerRegistered) {
+                DeepLinking.handleDeepLink(deepLink)
             }
         }
 
         return this
-            .takeIf { intent.isKlaviyoNotificationIntent }
-            ?.safeApply(preInitQueue) {
-                // Create and enqueue an $opened_push
-                val event = Event(EventMetric.OPENED_PUSH)
-                event.appendKlaviyoExtras(intent)
-
-                Registry.get<State>().pushToken?.let { event[EventKey.PUSH_TOKEN] = it }
-
-                // Not using createEvent here to avoid nested safeApply calls
-                Registry.get<State>().createEvent(event, Registry.get<State>().getAsProfile())
-            }?.safeApply {
-                // Dismiss the notification if opened via an action button.
-                // Body taps are handled by setAutoCancel(true) on the notification builder,
-                // but action button taps don't trigger auto-cancel (standard Android behavior).
-                intent?.getStringExtra(Constants.NOTIFICATION_TAG_EXTRA)?.let { tag ->
-                    NotificationManagerCompat
-                        .from(Registry.config.applicationContext)
-                        .cancel(tag, Constants.NOTIFICATION_ID)
-                }
-            }?.safeApply {
-                // If a Klaviyo notification is deep linked, invoke the developer's deep link handler
-                // if registered. If not, do nothing. The host already received the appropriate intent.
-                intent?.data?.takeIf {
-                    DeepLinking.isHandlerRegistered
-                }?.let { uri ->
-                    DeepLinking.handleDeepLink(uri)
-                }
-            }
-            ?: apply {
-                Registry.log.verbose("Non-Klaviyo intent ignored")
-            }
     }
 
     /**
@@ -475,20 +469,14 @@ object Klaviyo {
         get() = this?.getBooleanExtra(Constants.NOTIFICATION_AUTO_TRACKED_EXTRA, false) ?: false
 
     /**
-     * The dedup key identifying this intent's push delivery, or `null` if none is available.
+     * Dedup key for this intent's push delivery, or `null` if none is available. Prefers the `tm`
+     * field of the `_k` payload (a per-delivery ULID on campaign sends), else the SDK-generated
+     * [Constants.NOTIFICATION_UID_EXTRA]. Both are copied forward to the host's intent, so the
+     * trampoline call and a leftover manual call for the same tap resolve to the same key while
+     * distinct notifications stay distinct.
      *
-     * Prefers the `tm` field of the `_k` tracking payload — a ULID unique per delivery, present on
-     * real campaign sends. When `tm` is absent (local notifications, previews, non-campaign sends)
-     * it falls back to the SDK-generated [Constants.NOTIFICATION_UID_EXTRA] stamped on auto-tracked
-     * trampoline intents.
-     *
-     * Both keys are copied forward onto the host's destination intent, so the value is identical
-     * across the trampoline's `handlePush` call and a host's leftover manual call for the same tap —
-     * exactly what [handlePush]'s guard needs — while distinct notifications get distinct keys.
-     *
-     * There is deliberately no fallback to the raw `_k` payload: `_k` minus `tm` is per-message
-     * metadata shared by every delivery of the same campaign/message, so using it as a dedup key
-     * would drop legitimate opens across distinct deliveries.
+     * Deliberately not the raw `_k`: minus `tm` it is per-message metadata shared across deliveries,
+     * so it would collapse distinct opens.
      */
     private val Intent?.pushDeliveryId: String?
         get() {
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
@@ -49,7 +49,7 @@ object Klaviyo {
     private val preInitQueue: Queue<Operation<Unit>> = LinkedList()
 
     /**
-     * Push delivery IDs already auto-tracked within this process, so a single tap records one
+     * Push delivery IDs already handled within this process, so a single tap records one
      * `$opened_push`. See [handlePush] for how entries are matched and added.
      */
     private val handledPushDeliveries = BoundedIdSet()
@@ -337,23 +337,19 @@ object Klaviyo {
      */
     @JvmStatic
     fun handlePush(intent: Intent?): Klaviyo {
-        // Dedup guard for the automatic-tracking path. The trampoline calls handlePush and then
-        // forwards the same intent to the host, so a leftover manual handlePush call (or singleTask
-        // re-entry) would otherwise track a second open for one tap. Record each delivery the first
-        // time it's seen; if it's already recorded, drop this call entirely — no event, dismissal, or
-        // deep link. Manual-only integrations carry no auto-tracked marker, so they're unaffected.
-        val deliveryId = intent.pushDeliveryId
-        if (
-            intent.isAutoTrackedOpen &&
-            deliveryId != null &&
-            !handledPushDeliveries.markOnce(deliveryId)
-        ) {
-            Registry.log.verbose("Ignoring duplicate auto-tracked push open")
+        if (intent == null || !intent.isKlaviyoNotificationIntent) {
+            Registry.log.verbose("Non-Klaviyo intent ignored")
             return this
         }
 
-        if (intent == null || !intent.isKlaviyoNotificationIntent) {
-            Registry.log.verbose("Non-Klaviyo intent ignored")
+        // Dedup guard: track each push delivery at most once per process. The trampoline calls
+        // handlePush and forwards the same intent to the host, so a leftover manual handlePush call
+        // (or singleTask re-entry) would otherwise double-track one tap. The first call to see a
+        // delivery records it and proceeds; a later call for the same delivery short-circuits — no
+        // event, dismissal, or deep link.
+        val deliveryId = intent.pushDeliveryId
+        if (deliveryId != null && !handledPushDeliveries.markOnce(deliveryId)) {
+            Registry.log.verbose("Ignoring duplicate push open")
             return this
         }
 
@@ -461,19 +457,11 @@ object Klaviyo {
     fun isKlaviyoNotificationIntent(intent: Intent?): Boolean = intent.isKlaviyoNotificationIntent
 
     /**
-     * Whether this intent was routed through the automatic-open-tracking trampoline, i.e. carries
-     * [Constants.NOTIFICATION_AUTO_TRACKED_EXTRA]. Gates [handlePush]'s dedup guard so it never
-     * affects manual-only integrations.
-     */
-    private val Intent?.isAutoTrackedOpen: Boolean
-        get() = this?.getBooleanExtra(Constants.NOTIFICATION_AUTO_TRACKED_EXTRA, false) ?: false
-
-    /**
      * Dedup key for this intent's push delivery, or `null` if none is available. Prefers the `tm`
      * field of the `_k` payload (a per-delivery ULID on campaign sends), else the SDK-generated
-     * [Constants.NOTIFICATION_UID_EXTRA]. Both are copied forward to the host's intent, so the
-     * trampoline call and a leftover manual call for the same tap resolve to the same key while
-     * distinct notifications stay distinct.
+     * [Constants.NOTIFICATION_UID_EXTRA] stamped on trampoline intents. Both are copied forward to
+     * the host's intent, so the trampoline call and a leftover manual call for the same tap resolve
+     * to the same key while distinct notifications stay distinct.
      *
      * Deliberately not the raw `_k`: minus `tm` it is per-message metadata shared across deliveries,
      * so it would collapse distinct opens.
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
@@ -28,6 +28,7 @@ import com.klaviyo.core.config.Config
 import com.klaviyo.core.config.LifecycleException
 import com.klaviyo.core.safeApply
 import com.klaviyo.core.safeCall
+import com.klaviyo.core.utils.BoundedIdSet
 import com.klaviyo.core.utils.JSONUtil.toHashMap
 import com.klaviyo.core.utils.takeIf
 import java.io.Serializable
@@ -46,6 +47,26 @@ object Klaviyo {
      * Queue of failed operations attempted prior to [initialize]
      */
     private val preInitQueue: Queue<Operation<Unit>> = LinkedList()
+
+    /**
+     * In-memory dedup guard over push *delivery* IDs already auto-tracked by [handlePush].
+     *
+     * Only auto-tracking-path intents engage this guard: the trampoline stamps
+     * [Constants.NOTIFICATION_AUTO_TRACKED_EXTRA] on its targeted intents and forwards the
+     * destination intent (carrying that marker and the `_k` payload) to the host. The trampoline's
+     * own `handlePush` marks the delivery first; a host that also left a manual `handlePush` call in
+     * place then short-circuits instead of tracking a second `$opened_push` for the same tap. It
+     * also dedupes trampoline `singleTask` re-entry with the same intent.
+     *
+     * Manual-only integrations (auto-tracking disabled) never carry the marker, so they never engage
+     * the guard and behave exactly as before it existed.
+     */
+    private val handledPushDeliveries = BoundedIdSet()
+
+    /**
+     * Key within the `_k` tracking payload whose value uniquely identifies a single push delivery.
+     */
+    private const val PUSH_DELIVERY_KEY = "tm"
 
     /**
      * This method is provided for apps that are unable to register their API key immediately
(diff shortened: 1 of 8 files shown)


sdk/core/src/test/java/com/klaviyo/core/utils/BoundedIdSetTest.kt
```diff
@@ -0,0 +1,151 @@
...
+            }
+            // Release all workers at once to maximize contention on the same id.
+            start.countDown()
+            assertTrue(done.await(5, TimeUnit.SECONDS))
+
+            assertEquals(1, winners.size)
+        } finally {
+            pool.shutdownNow()
+        }
+    }
```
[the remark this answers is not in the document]
glenn-klaviyo: Fixed in dda37756e — extracted a `runConcurrently(threadCount, timeoutSeconds, worker)` helper that owns the pool / start-latch / await / shutdown; both stress tests now just pass their worker body and keep their own assertions.

-Claude

sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoTest.kt
```diff
@@ -704,6 +705,127 @@ internal class KlaviyoTest : BaseTest() {
...
     }
 
+    // --- Auto-tracked push-open dedup guard (`_k.tm`, gated by NOTIFICATION_AUTO_TRACKED_EXTRA) ---
+    // Note: Klaviyo is an object, so its in-memory dedup set persists across tests in this class.
+    // Each dedup test uses a distinct delivery ID to stay independent of execution order.
+
+    private fun deliveryExtras(deliveryId: String) = mapOf(
+        "com.klaviyo.body" to "Message body",
+        "com.klaviyo._k" to """{"m":"01GK4P5W6AV4V3APTJ727JKSKQ","tm":"$deliveryId"}"""
+    )
```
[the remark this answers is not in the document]
glenn-klaviyo: Acknowledged — skipping as low-impact. The rest of this file (e.g. `stubIntentExtras` and the other `handlePush` tests) already uses the raw `"com.klaviyo._k"` / `"com.klaviyo.body"` literals, so deriving these from `Constants` in only the new helpers would make the file *less* consistent. A file-wide switch to `Constants.PACKAGE_PREFIX + Constants.TRACKING_PARAMETER` would be a reasonable standalone cleanup if we want it.

-Claude

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -324,38 +345,57 @@ object Klaviyo {
...
+        // manual-only integrations, so their behavior is unchanged. Unseen delivery → atomically mark
+        // it and proceed.
+        if (intent.isAutoTrackedOpen) {
+            intent.pushDeliveryId?.let { deliveryId ->
+                if (!handledPushDeliveries.markOnce(deliveryId)) {
+                    Registry.log.verbose(
+                        "Ignoring duplicate auto-tracked push open for delivery $deliveryId"
+                    )
+                    return this
+                }
```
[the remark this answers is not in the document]
glenn-klaviyo: Verified against `safeCall` (KlaviyoException.kt) — the concern does not hold for the realistic failure path:

- The only failure that actually loses the trampoline’s `handlePush` here is a pre-init invocation, which throws a `KlaviyoException`. `safeApply(preInitQueue)` enqueues that op and replays it on `initialize()`, so the open is still tracked despite the early mark. The host’s manual call short-circuiting is the intended behavior (we want exactly one open, not a duplicate).
- A non-`KlaviyoException` is the only path that isn’t queued — but that is swallowed identically in the legacy manual-only path (and re-thrown in debug builds), and it’s deterministic for a given intent, so a host retry would fail the same way. No net regression vs. legacy.
- Moving the mark to after the enqueue would reintroduce a pre-init double-track: both the trampoline op and the host op buffer before either marks, then replay as two opens.

So the top-of-function mark is the more accurate choice across pre-init and post-init. Leaving as-is, but happy to revisit if we prefer the mark-after-tracking ordering. cc @glenn-klaviyo

-Claude

glenn-klaviyo:
bugbot run

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -426,6 +466,43 @@ object Klaviyo {
...
+            val deliveryId = this?.getStringExtra(PACKAGE_PREFIX + TRACKING_PARAMETER)
+                ?.takeIf { it.isNotEmpty() }
+                ?.let { trackingPayload ->
+                    runCatching { JSONObject(trackingPayload).optString(PUSH_DELIVERY_KEY) }
+                        .getOrNull()
+                        ?.takeIf { it.isNotEmpty() }
+                }
+            return deliveryId ?: this?.getStringExtra(Constants.NOTIFICATION_UID_EXTRA)
+                ?.takeIf { it.isNotEmpty() }
+        }
```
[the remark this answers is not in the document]
glenn-klaviyo: on Android, an activity-launching action collapses the shade and handlePush cancels the notification, so a second interaction on the same notification isn't or shouldn't reachable. Did some manual testing around this and wasn't able to repro.

sdk/core/src/test/java/com/klaviyo/core/utils/BoundedIdSetTest.kt
```diff
@@ -0,0 +1,150 @@
...
+
+            repeat(threadCount) { i ->
+                pool.execute {
+                    start.await()
+                    worker(i)
+                    done.countDown()
+                }
+            }
+            start.countDown()
+            assertTrue(done.await(timeoutSeconds, TimeUnit.SECONDS))
```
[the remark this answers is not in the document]
glenn-klaviyo: Fixed in ca5b0f2dd — workers now count down in a `finally` and any thrown `Throwable` is collected and asserted (with stack traces) after the await, so a failing worker reports its real cause instead of a misleading timeout. Adapted the snippet to our repo: imported `ConcurrentLinkedQueue` rather than an inline FQN, and used JUnit 4 `assertTrue(message, condition)` arg order.

-Claude

glenn-klaviyo:
bugbot run

glenn-klaviyo:
@coderabbitai review

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -324,38 +345,57 @@ object Klaviyo {
...
-                DeepLinking.isHandlerRegistered
-            }?.let { uri ->
-                DeepLinking.handleDeepLink(uri)
+    fun handlePush(intent: Intent?): Klaviyo {
+        // Dedup guard (top of function), scoped to the automatic-tracking path only: a push delivery
+        // we've already auto-tracked means this is a re-delivery of the same tap — a host's leftover
+        // manual handlePush after the trampoline already tracked, or singleTask re-entry.
+        // Short-circuit fully: no event, no dismissal, no deep link dispatch. The marker is absent on
+        // manual-only integrations, so their behavior is unchanged. Unseen delivery → atomically mark
+        // it and proceed.
```
ab1470: > Unseen delivery → atomically mark it and proceed.

nit: what does this mean?

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -324,38 +345,57 @@ object Klaviyo {
...
+        if (intent.isAutoTrackedOpen) {
+            intent.pushDeliveryId?.let { deliveryId ->
+                if (!handledPushDeliveries.markOnce(deliveryId)) {
+                    Registry.log.verbose(
+                        "Ignoring duplicate auto-tracked push open for delivery $deliveryId"
+                    )
+                    return this
+                }
             }
         }
```
ab1470:
```suggestion
        val deliveryId = intent.pushDeliveryId
        
        if (intent.isAutoTrackedOpen && deliveryId != null && !handledPushDeliveries.markOnce(deliveryId)) {
            Registry.log.verbose("Ignoring duplicate auto-tracked push open for delivery $deliveryId")
            return this
        }
```

My Swift-y brain didn't like this nested logic so I asked Gemini to suggest how to make this more readable, and this is what it came up with

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -324,38 +345,57 @@ object Klaviyo {
...
+                // if registered. If not, do nothing. The host already received the appropriate intent.
+                intent?.data?.takeIf {
+                    DeepLinking.isHandlerRegistered
+                }?.let { uri ->
+                    DeepLinking.handleDeepLink(uri)
+                }
+            }
+            ?: apply {
+                Registry.log.verbose("Non-Klaviyo intent ignored")
+            }
```
ab1470:
```suggestion
        // 1. Early exit (The Kotlin equivalent of a guard statement)
        if (!intent.isKlaviyoNotificationIntent) {
            Registry.log.verbose("Non-Klaviyo intent ignored")
            return this
        }
        
        // 2. Extract shared state to a local variable
        val state = Registry.get<State>()
        
        // 3. Keep safeApply only where it dictates a specific execution context/queue
        this.safeApply(preInitQueue) {
            val event = Event(EventMetric.OPENED_PUSH)
            event.appendKlaviyoExtras(intent)
            
            state.pushToken?.let { event[EventKey.PUSH_TOKEN] = it }
            
            state.createEvent(event, state.getAsProfile())
        }
        
        // 4. Flatten the notification dismissal logic
        val notificationTag = intent?.getStringExtra(Constants.NOTIFICATION_TAG_EXTRA)
        if (notificationTag != null) {
            NotificationManagerCompat
                .from(Registry.config.applicationContext)
                .cancel(notificationTag, Constants.NOTIFICATION_ID)
        }
        
        // 5. Flatten the deep link logic
        val deepLinkUri = intent?.data
        if (deepLinkUri != null && DeepLinking.isHandlerRegistered) {
            DeepLinking.handleDeepLink(deepLinkUri)
        }
        
        return this
```

I asked Gemini if this code block was idiomatic Kotlin, and it said no, not really:

> To be completely candid: no, this is not considered idiomatic Kotlin. While it uses a lot of Kotlin-specific features (takeIf, let, apply, and Elvis operators), it falls into a very common trap known as Scope Function Abuse. It’s incredibly easy to do when you first discover Kotlin's functional chains, but it results in code that is hard to read, hard to debug, and fragile.
>
>Here is a breakdown of why this structure is problematic, followed by how to refactor it idiomatically.
>
>What Makes This Unidiomatic?
> - Obscured Control Flow: You are using a long chain (return this.takeIf { ... }?.apply { ... } ?: apply { ... }) to replicate a simple if / else statement. This forces the reader to mentally parse a functional pipeline just to figure out the basic execution path.
> - The Fragile Elvis Operator (?:): The final ?: apply { ... } acts as an "else" block if isKlaviyoNotificationIntent is false. However, if any of those intermediate safeApply blocks were to unexpectedly return null due to an internal error or refactor, it would accidentally trigger the "Non-Klaviyo intent ignored" log, creating a very confusing bug.
> - Repeated Lookups: You call Registry.get<State>() twice in the same block. Fetching it once and storing it in a local variable is cleaner and more performant.
> - Nested Scope Functions: Inside the chain, you have even more scope functions (?.let, ?.takeIf { ... }?.let). This brings back the "pyramid of doom" we avoided in your previous snippet.

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -47,6 +48,26 @@ object Klaviyo {
...
+     * Only auto-tracking-path intents engage this guard: the trampoline stamps
+     * [Constants.NOTIFICATION_AUTO_TRACKED_EXTRA] on its targeted intents and forwards the
+     * destination intent (carrying that marker and the `_k` payload) to the host. The trampoline's
+     * own `handlePush` marks the delivery first; a host that also left a manual `handlePush` call in
+     * place then short-circuits instead of tracking a second `$opened_push` for the same tap. It
+     * also dedupes trampoline `singleTask` re-entry with the same intent.
+     *
+     * Manual-only integrations (auto-tracking disabled) never carry the marker, so they never engage
+     * the guard and behave exactly as before it existed.
+     */
```
ab1470, line 54: I don't want to get super nitpicky about comments again lol, but this seems like a lot of explanation for a single property. It's describing behavior that's enabled by this property but is implemented elsewhere; this property is one small piece of making this behavior work. I feel like it would be better to have this description of the behavior closer to where that business logic is actually happening

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
glenn-klaviyo: ah yes, I realize this is unclear - basically means "if we haven't seen this delivery, record it and continue". I'll improve the readability of this.

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
glenn-klaviyo: a lot of this code was existing logic I mainly just added the `isAutoTrackedOpen` logic. That being said, I'll try and leave the code in a _better_ place :)

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
glenn-klaviyo: comment cleaned up

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
glenn-klaviyo: cleaned up

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -324,39 +336,61 @@ object Klaviyo {
...
-            // Create and enqueue an $opened_push
+    fun handlePush(intent: Intent?): Klaviyo {
+        // Dedup guard for the automatic-tracking path. The trampoline calls handlePush and then
+        // forwards the same intent to the host, so a leftover manual handlePush call (or singleTask
+        // re-entry) would otherwise track a second open for one tap. Record each delivery the first
+        // time it's seen; if it's already recorded, drop this call entirely — no event, dismissal, or
+        // deep link. Manual-only integrations carry no auto-tracked marker, so they're unaffected.
+        val deliveryId = intent.pushDeliveryId
+        if (
+            intent.isAutoTrackedOpen &&
```
evan-masseau: I'd lean toward de-duping all, not just the auto-tracked opens. If we only dedupe on the auto-tracked one, couldn't that create an ordering issue? 

i.e. an autotracked push open first, and host app calls `handlePush` second would still lead to a duplicate request?

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -324,39 +336,61 @@ object Klaviyo {
...
+            !handledPushDeliveries.markOnce(deliveryId)
+        ) {
+            Registry.log.verbose("Ignoring duplicate auto-tracked push open")
+            return this
+        }
+
+        if (intent == null || !intent.isKlaviyoNotificationIntent) {
+            Registry.log.verbose("Non-Klaviyo intent ignored")
+            return this
+        }
```
evan-masseau: nit: I'd suggest keeping this check first in the function

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -324,39 +336,61 @@ object Klaviyo {
...
+
+        // Create and enqueue an $opened_push. safeApply(preInitQueue) buffers this for replay if
+        // handlePush runs before initialize(), and guards against unexpected failures.
+        safeApply(preInitQueue) {
```
evan-masseau, line 358: I gotta take a closer look at the change of chaining here. Might be ok, but want to verify in Android Studio

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -426,6 +460,37 @@ object Klaviyo {
     @JvmStatic
     fun isKlaviyoNotificationIntent(intent: Intent?): Boolean = intent.isKlaviyoNotificationIntent
 
+    /**
+     * Whether this intent was routed through the automatic-open-tracking trampoline, i.e. carries
+     * [Constants.NOTIFICATION_AUTO_TRACKED_EXTRA]. Gates [handlePush]'s dedup guard so it never
+     * affects manual-only integrations.
+     */
+    private val Intent?.isAutoTrackedOpen: Boolean
```
evan-masseau: My claude's gonna send you an idea for extracting logic for these from Klaviyo.kt. Not your fault, something I always regretted putting in here and now that we're adding more, would like to extract it a bit for code organizations sake

sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoTest.kt
```diff
@@ -704,6 +705,127 @@ internal class KlaviyoTest : BaseTest() {
         verify(inverse = true) { mockApiClient.enqueueEvent(any(), any()) }
     }
 
+    // --- Auto-tracked push-open dedup guard (`_k.tm`, gated by NOTIFICATION_AUTO_TRACKED_EXTRA) ---
+    // Note: Klaviyo is an object, so its in-memory dedup set persists across tests in this class.
+    // Each dedup test uses a distinct delivery ID to stay independent of execution order.
+
+    private fun deliveryExtras(deliveryId: String) = mapOf(
```
evan-masseau, line 712: Similar to above, the tests could move to their own file too

sdk/core/src/main/java/com/klaviyo/core/utils/BoundedIdSet.kt
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * When the set is full, inserting a new ID evicts the oldest-inserted one. A duplicate insert is a
+ * no-op that does **not** refresh the ID's position in eviction order, so the set behaves as a
+ * sliding window over the most recently *first-seen* IDs.
+ *
+ * All operations are guarded by an intrinsic lock, so the set is safe to share across threads.
+ *
+ * @param capacity maximum number of IDs retained before the oldest is evicted.
+ */
+class BoundedIdSet(private val capacity: Int = DEFAULT_CAPACITY) {
```
evan-masseau, line 15: This is so overkill, its great.

evan-masseau:
Broadly approve, but some code organization ideas I'm gonna send your way, and I'd like to explore applying this dedupe across the board, e.g. even if I as an app developer called `handlePush` twice on the same push by mistake, would this catch the dupe for me? Or if automated is called first and my manual call came second?

evan-masseau:
Since you're adding more to `Klaviyo.kt` here, a code-organization thought for a follow-up (happy for this to be a **separate PR stacked on `feat/mage-768-dedup-guard`** — no need to bloat this one):

**Extract the intent/URI utils out of `Klaviyo.kt`.** Everything from `Intent.isKlaviyoIntent` down (the `isKlaviyo*Intent` / `isKlaviyoUniversalTracking*` extensions, ~line 398 onward) is pure util code sitting in the middle of the public facade. It'd read a lot better in a `com.klaviyo.analytics.utils` subpackage, mirroring `core.utils.IntentExtension`.

**⚠️ One catch that makes this not a pure move:** these are public members of `object Klaviyo`, and consumers reach them *through the `Klaviyo` qualifier*:
- Kotlin: `import com.klaviyo.analytics.Klaviyo.isKlaviyoNotificationIntent` (see `SampleActivity.kt`)
- Java: `Klaviyo.isKlaviyoIntent(intent)` and `Klaviyo.INSTANCE.isKlaviyoIntent(intent)` (see `KlaviyoJavaApiTest.java`)

Kotlin also won't let you split an `object` declaration across files, so relocating them *out* of the object is the only way to physically move them — and that changes the qualifier for every caller = **source break for Kotlin, source + binary break for Java.** So two ways to play it:

1. **Non-breaking now (extract-and-delegate):** move the real logic to `internal` top-level functions in `analytics/.../utils/IntentExtensions.kt`, and leave thin public members in `Klaviyo` that delegate (keeping the same `@JvmStatic` / `@JvmName`). Public surface stays byte-for-byte identical; you still get the file cleaned up.
2. **Clean top-level public move (drop the `Klaviyo` members):** nicer end state, but it's a breaking change → **major version only.** Worth noting `isKlaviyoIntent` is already `@Deprecated` for next-major removal, so a major is the natural time to fully relocate the rest too.

**Tests too:** `KlaviyoTest.kt` is ~1000 lines. In the same spirit, the intent/URI extension tests could move into a dedicated file (e.g. `utils/IntentExtensionsTest.kt`, or under `linking/`), matching the existing per-subpackage test layout (`state/`, `model/`, `networking/`, `linking/`).

Not blocking this PR at all — just flagging it while the file's top of mind, and the compat constraint so nobody does the naive move and breaks integrators.

sdk/core/src/main/java/com/klaviyo/core/utils/BoundedIdSet.kt
evan-masseau: I did a little poking at the idea with Claude, from what I can tell this is an elegant solution to three-pronged problem without over-engineering (joking comment above aside). Looked at some alternative built in data structs but none solve all 3 (limit capacity, threadsafe concurrency, atomic markOnce)

So ignore the haters (me) this looks good.

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
glenn-klaviyo: Done in e0df0d9fb — dropped the marker gate; handlePush now dedupes every Klaviyo open by delivery id, so ordering can't produce a dupe.

-Claude

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
glenn-klaviyo: Done — isKlaviyoNotificationIntent guard now runs first, dedup guard second.

-Claude

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
glenn-klaviyo: Reworked into early-return guards (no more takeIf/elvis chain) — ready for your Android Studio pass.

-Claude

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
glenn-klaviyo: Deferring the Klaviyo.kt extraction to a follow-up PR to keep this one focused.

-Claude

sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoTest.kt
glenn-klaviyo: Same — moving these with the extraction in the follow-up PR.

-Claude

sdk/core/src/main/java/com/klaviyo/core/utils/BoundedIdSet.kt
glenn-klaviyo: Thanks — also switched it to a private lock monitor per a CodeRabbit nit.

-Claude

glenn-klaviyo:
> Broadly approve, but some code organization ideas I'm gonna send your way, and I'd like to explore applying this dedupe across the board, e.g. even if I as an app developer called `handlePush` twice on the same push by mistake, would this catch the dupe for me? Or if automated is called first and my manual call came second?

I've addressed some of your comments above to remove the auto-tracked market gate, so `handlePush` no longer distinguishes trampoline vs. manual - it dedupes any Klaviyo open on it's delivery id. Auto-first/manual-second, manual-twice-by-mistake, a `singleTask` re-entry: all collapse to **one** open. It does require a real campaign as previews don't have `_k` payload properties. cc @evan-masseau

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoTrampolineActivity.kt
```diff
@@ -85,10 +87,14 @@ internal class KlaviyoTrampolineActivity : Activity() {
...
          * seam as [forBrowserUrl]. Callers append their own Klaviyo extras for parity with the
          * non-trampoline intents they replace.
+         *
+         * Stamps a fresh per-intent [Constants.NOTIFICATION_UID_EXTRA] that `handlePush` uses as a
+         * fallback dedup key for payloads without a `tm` (e.g. local notifications, previews).
          */
         internal fun forDestination(context: Context, deepLink: Uri? = null): Intent = Intent().apply {
             setClassName(context.packageName, KlaviyoTrampolineActivity::class.java.name)
             deepLink?.let { data = it }
+            putExtra(Constants.NOTIFICATION_UID_EXTRA, UUID.randomUUID().toString())
```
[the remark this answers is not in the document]
glenn-klaviyo: Leaving as-is. On Android a notification is dismissed on the first interaction — body taps auto-cancel, and action-button taps collapse the shade while handlePush cancels the notification — so a user can't tap a second target on the same notification, and the multiple-opens case isn't reachable in practice. The guard's real job still holds: the same tap tracked by the trampoline and then a leftover manual handlePush share the same forwarded UUID, so that dedupes. (Consistent with the earlier action-button thread.)

-Claude

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -47,6 +48,17 @@ object Klaviyo {
      */
     private val preInitQueue: Queue<Operation<Unit>> = LinkedList()
 
+    /**
+     * Push delivery IDs already handled within this process, so a single tap records one
+     * `$opened_push`. See [handlePush] for how entries are matched and added.
+     */
+    private val handledPushDeliveries = BoundedIdSet()
```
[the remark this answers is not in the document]
glenn-klaviyo: Leaving as-is for now. The dedup tests avoid cross-test bleed by using a distinct delivery id per test (noted in-file), so there's no failure today. Opting not to add a test-only reset hook onto the public Klaviyo object for this PR — can revisit if we want stricter isolation.

-Claude

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoTrampolineActivity.kt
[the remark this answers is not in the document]
evan-masseau: Actually, this drew my attention, but for different reasons than bugbot's.
If I'm reading correctly, you're only adding this fallback UUID for `tm`-less notifications that are going through the trampoline activity. I think we should add this UUID when the host app is not using trampoline activity too. Goes along with my earlier feedback to have this dedupe guard cover all de-duping of `handlePush` per notification. That'd give de-dupe coverage to all "preview" notifications regardless of whether the app is using trampoline activity.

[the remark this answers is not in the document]
glenn-klaviyo: funny, this was my original thought too, then claude pushed me away from it.. I'll address this.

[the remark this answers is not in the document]
glenn-klaviyo: Done in 444e60efe — moved the fallback uid out of the trampoline factory into buildNotification, so every notification's body + buttons carry it (trampoline or not). With opens now deduped by delivery id across the board, handlePush covers all per-notification de-duping. Made it one uid per notification, so body/buttons share a key and don't split.

-Claude

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
evan-masseau: I think it looks good. the fact that tests were untouched it a good sign too. More readable than all the chaining, thanks for the cleanup.

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -426,6 +456,29 @@ object Klaviyo {
...
+     *
+     * Deliberately not the raw `_k`: minus `tm` it is per-message metadata shared across deliveries,
+     * so it would collapse distinct opens.
+     */
+    private val Intent?.pushDeliveryId: String?
+        get() {
+            val deliveryId = this?.getStringExtra(PACKAGE_PREFIX + TRACKING_PARAMETER)
+                ?.takeIf { it.isNotEmpty() }
+                ?.let { trackingPayload ->
+                    runCatching { JSONObject(trackingPayload).optString(PUSH_DELIVERY_KEY) }
```
evan-masseau, line 474: I find myself wondering why use `tm` at all anymore TBH. Non-blocking, just curious if now that we' are always making a notification uid, is that just the more reliable/consistent thing to use?

evan-masseau (approved):

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
glenn-klaviyo: yeah the UUID approach does handle most of the cases. The one thing claude keeps pointing out is

> dedup across genuinely duplicate deliveries of the same push. If FCM delivers the same message twice (at-least-once semantics) and there's no notification_tag to collapse them, two notifications show — each with its own fresh UUID but the same tm. tm collapses those to one open; the UUID would count two. Narrow (needs dup delivery + no collapse tag), but real. tm is also the server's per-delivery identity, so it's the more "correct" grain semantically.

I'm leaning toward keeping as is given this. It also aligns with iOS right now. If we want to simplify in the future should be an easy lift.

PR #499, opened 2026-07-02
feat(forms): handle IAF openExternalUrl bridge event [PUSH-733]

josephzhou-klaviyo:
# Description
Handle the IAF `openExternalUrl` native bridge event so in-app form CTAs with external/web URLs open in the default browser instead of routing through the deep link handler. Surfaces a new `FormCtaExternalUrlClicked` lifecycle event. Resolves PUSH-733.

Tested on emulator and validated that external url buttons work in in-app forms
## Due Diligence
  - Unit tests pass. **Full end-to-end QA is now complete** (previously gated on the fender-side emitter — see the 2026-07-08 update below): live-validated against a real Fender dev server, local Django, the fender `push-634` companion branch, and a real emulator across `mailto:`, `tel:`, and `javascript:` (rejected).

  - Adds `FormLifecycleEvent.FormCtaExternalUrlClicked`. **No version bump in this PR** — versioning to be assigned at release time.
  - README documents the new lifecycle event. Since it describes an unreleased API, consider merging to a feature/docs branch so docs only go live on release (currently targeting `master` per request).

## Changelog / Code Overview
Self-contained on `master` — **not** dependent on PUSH-637 / `feat/push-trampoline`. Rather than a new `DeepLinking` browser-intent helper, this reuses the existing external-browser intent pattern already in `KlaviyoWebViewClient.shouldOverrideUrlLoading` (`ACTION_VIEW` + `FLAG_ACTIVITY_NEW_TASK`, no `setPackage`).

- `NativeBridgeMessage.kt` — new `OpenExternalUrl` message, decode branch (required `url`), and handshake spec `openExternalUrl` v1 (permits KlaviyoJS to emit it).
- `KlaviyoNativeBridge.kt` — dispatch branch + `openExternalUrl()` handler: opens the browser, then fires `FormCtaExternalUrlClicked` (skips the callback with a warning if `formId`/`formName` are missing). Never touches `DeepLinking`, so existing `openDeepLink` behavior is unchanged.
- `FormLifecycleEvent.kt` — new public `FormCtaExternalUrlClicked(formId, formName, buttonLabel, externalUrl: Uri)` case.
- `README.md` — documents the new event (Kotlin + Java).
- `SampleApplication.kt` — wires the new event into the sample's lifecycle handler.
- Tests: decode + handshake (`NativeBridgeMessageTest`), handler intent + lifecycle + edge cases (`KlaviyoNativeBridgeTest`), Java interop (`InAppFormsJavaApiTest`).

## Update (2026-07-08): scheme allowlist + scheme-specific intent actions

Two follow-up commits extend `openExternalUrl` to match `klaviyo-swift-sdk` PR #623 and achieve parity with the (unmerged) Android push fix in `timmorrill/push-834-handle-non-web-url-schemes` (PUSH-834):

- **`86cd7faa1` — Scheme allowlist gate.** `openExternalUrl` now validates the URL's scheme against an allowlist (`http`, `https`, `mailto`, `tel`, `sms`, `smsto`) before dispatch. Disallowed schemes (`javascript:`, `intent:`, `file:`, `data:`, etc.) are dropped silently with a warning log and never fire `FormCtaExternalUrlClicked`. `smsto` is included per PUSH-834's Android-specific note (Android has a handler for it; iOS omits it only because iOS Messages has no `smsto:` handler).
- **`0b82d48ca` — Scheme-specific intent actions.** `openExternalUrl` now dispatches `ACTION_DIAL` for `tel:`, `ACTION_SENDTO` for `mailto:`/`sms:`/`smsto:`, and `ACTION_VIEW` for `http`/`https` (previously always `ACTION_VIEW`), plus adds the matching `<queries>` declarations to `sdk/core`'s manifest — required so `resolveActivity()` can see the dialer/SMS/mail handlers under Android 11+ package visibility (without this, `tel:`/`mailto:`/etc. silently fail to resolve even when a handler app is installed).

Both changes mirror — and are structured to consolidate cleanly with, once merged — PUSH-834's `DeepLinking`/`KlaviyoRemoteMessage` implementation for push's `open_url`/`web_url` fields: same scheme sets, and the manifest `<queries>` addition is byte-for-byte identical to PUSH-834's, so the two branches will merge without conflict.

**Live-validated end-to-end on a real emulator**, against the real Fender dev server + local Django + fender's `push-634` companion branch (which required a matching `openWebUrl` → `openExternalUrl` rename on the JS side to interoperate with this PR — done and verified):
- `mailto:` → allowlist passes, Gmail opens via `ACTION_SENDTO`, `FormCtaExternalUrlClicked` fires with the correct data.
- `tel:` → allowlist passes, Dialer opens with the number pre-filled via `ACTION_DIAL`, lifecycle event fires.
- `javascript:` → allowlist rejects it: warning logged, zero intent dispatch, zero lifecycle event, no crash.

## Test Plan
1. `./gradlew ktlintCheck` — clean.
2. `./gradlew :sdk:forms:testDebugUnitTest --tests "com.klaviyo.forms.bridge.NativeBridgeMessageTest" --tests "com.klaviyo.forms.bridge.KlaviyoNativeBridgeTest" --tests "com.klaviyo.forms.InAppFormsJavaApiTest"` — pass.
3. `./gradlew :sdk:forms:build :sdk:forms-core:build :sdk:analytics:build` — green.
4. **Manual QA — DONE (see 2026-07-08 update above):** validated against a real fender form with `mailto:`, `tel:`, and `javascript:` (rejected) CTAs. `FormCtaExternalUrlClicked` fires correctly for allowed schemes and is correctly skipped for the rejected one; existing deep-link CTAs unchanged.

> ⚠️ **Dependency (resolved for local testing, still tracks a real fender PR):** requires the fender-side change to emit `openExternalUrl` from KlaviyoJS — confirmed working against fender branch `josephzhou/push-634-iaf-open-url-runtime` (which required its own `openWebUrl` → `openExternalUrl` rename to match this PR's message type). Verify that branch (or whatever supersedes it) actually merges before this ships to production, since this SDK path is inert until it does.

## Related Issues/Tickets
- PUSH-733 (this) — split from PUSH-637 (Android push `open_url`)
- PUSH-732 — iOS parallel (`formExternalUrlClicked`)

## Summary by CodeRabbit

* **New Features**
  * Added support for form CTAs that open external web links in the device browser.
  * Lifecycle callbacks now include details for external-link CTA clicks, such as button label and URL.

* **Bug Fixes**
  * Restricted link handling to safe, supported URL schemes.
  * Added validation so malformed link actions are skipped and logged instead of causing issues.

* **Documentation**
  * Updated form lifecycle guidance with the new external-link CTA behavior.

* **Tests**
  * Expanded coverage for message decoding, browser launch behavior, and Java accessibility.

### DIFF
--- a/sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoNativeBridge.kt
+++ b/sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoNativeBridge.kt
@@ -1,5 +1,6 @@
 package com.klaviyo.forms.bridge
 
+import android.content.Intent
 import android.net.Uri
 import android.webkit.JavascriptInterface
 import android.webkit.WebView
@@ -11,6 +12,7 @@ import com.klaviyo.analytics.Klaviyo
 import com.klaviyo.analytics.linking.DeepLinking
 import com.klaviyo.analytics.networking.ApiClient
 import com.klaviyo.core.Registry
+import com.klaviyo.core.utils.startActivityIfResolved
 import com.klaviyo.forms.FormLifecycleEvent
 import com.klaviyo.forms.FormLifecycleHandler
 import com.klaviyo.forms.bridge.NativeBridgeMessage.Abort
@@ -19,6 +21,7 @@ import com.klaviyo.forms.bridge.NativeBridgeMessage.FormWillAppear
 import com.klaviyo.forms.bridge.NativeBridgeMessage.HandShook
 import com.klaviyo.forms.bridge.NativeBridgeMessage.JsReady
 import com.klaviyo.forms.bridge.NativeBridgeMessage.OpenDeepLink
+import com.klaviyo.forms.bridge.NativeBridgeMessage.OpenExternalUrl
 import com.klaviyo.forms.bridge.NativeBridgeMessage.TrackAggregateEvent
 import com.klaviyo.forms.bridge.NativeBridgeMessage.TrackProfileEvent
 import com.klaviyo.forms.presentation.PresentationManager
@@ -73,6 +76,7 @@ internal class KlaviyoNativeBridge : NativeBridge {
                 is TrackAggregateEvent -> createAggregateEvent(bridgeMessage)
                 is TrackProfileEvent -> createProfileEvent(bridgeMessage)
                 is OpenDeepLink -> deepLink(bridgeMessage)
+                is OpenExternalUrl -> openExternalUrl(bridgeMessage)
                 is FormDisappeared -> close(bridgeMessage)
                 is Abort -> abort(bridgeMessage.reason)
             }
@@ -151,6 +155,41 @@ internal class KlaviyoNativeBridge : NativeBridge {
                 formName = message.formName,
                 buttonLabel = message.buttonLabel,
                 deepLinkUrl = deepLinkUri
+            )
+        )
+    }
+
+    /**
+     * Handle an [OpenExternalUrl] message by opening the URL in the default browser.
+     *
+     * Unlike [deepLink], the intent is not package-scoped (no `setPackage`), so the OS routes it
+     * to the default browser, bypassing any registered deep link handler — mirroring
+     * [com.klaviyo.forms.webview.KlaviyoWebViewClient.shouldOverrideUrlLoading]. The `NEW_TASK`
+     * intent launches independently of the overlay activity, so no grace period is needed.
+     * Fires [FormLifecycleEvent.FormCtaExternalUrlClicked] after dispatch.
+     */
+    private fun openExternalUrl(message: OpenExternalUrl) {
+        val externalUri = message.url.toUri()
+
+        Intent().apply {
+            data = externalUri
+            action = Intent.ACTION_VIEW
+            flags = Intent.FLAG_ACTIVITY_NEW_TASK
+        }.startActivityIfResolved(Registry.config.applicationContext)
+
+        if (message.formId.isEmpty() || message.formName.isEmpty()) {
+            Registry.log.warning(
+                "OpenExternalUrl missing required fields, skipping lifecycle callback"
+            )
+            return
+        }
+
+        invokeFormLifecycleHandler(
+            FormLifecycleEvent.FormCtaExternalUrlClicked(
+                formId = message.formId,
+                formName = message.formName,
+                buttonLabel = message.buttonLabel,
+                externalUrl = externalUri
             )
         )
     }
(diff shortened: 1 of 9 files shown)


sdk/forms/src/main/java/com/klaviyo/forms/bridge/KlaviyoNativeBridge.kt
```diff
@@ -155,6 +159,41 @@ internal class KlaviyoNativeBridge : NativeBridge {
...
+
+        invokeFormLifecycleHandler(
+            FormLifecycleEvent.FormCtaExternalUrlClicked(
+                formId = message.formId,
+                formName = message.formName,
+                buttonLabel = message.buttonLabel,
+                externalUrl = externalUri
+            )
+        )
+    }
```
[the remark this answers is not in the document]
josephzhou-klaviyo: Thanks — declining this one (as a nitpick/defense-in-depth, not a correctness issue).

The suggestion is to mirror "the same scheme policy used by `KlaviyoWebViewClient.shouldOverrideUrlLoading`", but that method doesn't actually validate schemes — it builds `ACTION_VIEW` from `request.url` and launches it with no `http`/`https` check (KlaviyoWebViewClient.kt:207-219), exactly like this handler. So adding the guard here wouldn't make the two paths consistent; it would make `openExternalUrl` *stricter* than the existing external-browser path.

On the merit: the bridge is origin-locked to `Registry.config.baseUrl`, and this event is emitted only by trusted KlaviyoJS whose `IAF_OPEN_URL` runtime emits web URLs by design — so a `content://`/`file://` value effectively can't reach here in practice.

Separately, scheme handling is intentionally being unified in the sibling push work (PUSH-834) via a scheme-aware external-intent helper that will *support* additional schemes (mailto/tel/sms), not restrict to http/https. If we add hardening, the right scope is to route both this handler and `shouldOverrideUrlLoading` through that shared helper together, rather than a one-off `setOf("http","https")` gate in a single method. Tracking that as a follow-up.

morrilltim:
I think we want to this target the push-trampoline branch

sdk/core/src/main/res/values/strings.xml
```diff
@@ -2,7 +2,7 @@
 <resources>
     <!-- Please do not modify / override these values, they are critical for internal versioning  -->
     <string name="klaviyo_sdk_name_override">android</string>
-    <string name="klaviyo_sdk_version_override">4.4.0</string>
```
ndurell, line 5: Maybe check with the team to get the right version.

sdk/core/src/main/AndroidManifest.xml
ndurell, line 17: Maybe check with the team if we want this added by default to all apps that use our sdk.

ndurell: Could add this to the sample instead?

ndurell (approved):

PR #501, opened 2026-07-06
feat(forms): handle BadJWT native bridge message (MAGE-846)

ab1470:
# Description
Fender's forms webview posts `{"type":"BadJWT","data":{}}` over the native bridge when it rejects an injected JWT (e.g. bad signature, or a token that doesn't resolve to a profile). The Android SDK had no case for it, so it threw `IllegalStateException` which got caught and logged at **error** level — misleading, since a rejected JWT is a normal outcome. This handles it gracefully as a warning-level no-op.

## Due Diligence

  - Part of the JWT-for-personalized-forms feature; this PR targets the `feat/jwts-for-personalized-forms` feature branch.

## Changelog / Code Overview
- **`NativeBridgeMessage.kt`** — added a `BadJwt` message type to the sealed class; `decodeWebviewMessage` now maps it instead of throwing. The wire type is PascalCase (`"BadJWT"`), which breaks the lower-camelCase `keyName` convention used by every other type, so it's matched against an explicit `BAD_JWT_TYPE` literal (documented inline).
- **`KlaviyoNativeBridge.kt`** — added the `BadJwt` branch to the dispatch `when`; `badJwt()` logs at **warning** and no-ops, matching iOS's degrade-don't-throw behavior and the repo's logging-severity guidelines.
- **Tests** — decode maps `"BadJWT"` → `BadJwt`; a case-sensitivity regression test asserts near-miss casings still throw; and a dispatch-level test asserts `BadJWT` logs a warning and does **not** hit the error-level catch.

**Not added to the handshake (`handShakeData`), by design.** Verified against fender (`onsite-in-app-forms`, MAGE-553 on master): BadJWT delivery is gated on the SDK advertising `jwtMutation` (`native/index.ts`), never on `BadJWT` itself — `postNativeMessage` does no per-type gating and nothing calls `getSpecVersion(..., 'BadJWT')`. A handshake entry would have zero effect and would force the PascalCase exception into a second location.

**Follow-up (out of scope):** `BadJwt` is currently a terminal log-only no-op — there's no host-app-facing signal or `AuthTokenManager` re-fetch path, so a rejected token leaves the form unauthenticated for the session. Worth a separate ticket alongside the wider fender-protocol audit noted in MAGE-846.

## Test Plan
- `./gradlew :sdk:forms:testDebugUnitTest --tests "com.klaviyo.forms.bridge.NativeBridgeMessageTest" --tests "com.klaviyo.forms.bridge.KlaviyoNativeBridgeTest"` — pass
- `./gradlew :sdk:forms:ktlintCheck` — pass
- Manual repro (from MAGE-846): register an auth-token provider returning a well-formed-but-unsigned JWT, trigger a personalized form; fender rejects it and posts `BadJWT`. Previously logged an error+stacktrace; now logs a single warning and the form continues unauthenticated. _(Emulator repro checkbox left unchecked pending a device pass.)_

## Related Issues/Tickets
Resolves MAGE-846

## Summary by CodeRabbit

* **Bug Fixes**
  * Added handling for a new bridge message indicating an invalid/rejected JWT.
  * The form bridge now logs this as a warning (not an error) and continues without crashing.
* **Tests**
  * Added unit tests confirming the new message type is decoded correctly.
  * Added coverage to ensure type matching is case-sensitive and mismatches throw as expected.

### DIFF
--- a/sdk/forms/src/test/java/com/klaviyo/forms/bridge/KlaviyoNativeBridgeTest.kt
+++ b/sdk/forms/src/test/java/com/klaviyo/forms/bridge/KlaviyoNativeBridgeTest.kt
@@ -453,6 +453,19 @@ internal class KlaviyoNativeBridgeTest : BaseTest() {
     }
 
     @Test
+    fun `BadJWT logs a warning and does not throw`() {
+        /**
+         * A rejected JWT is a normal outcome, so it should degrade gracefully rather than
+         * hitting the error-level catch block.
+         *
+         * @see com.klaviyo.forms.bridge.KlaviyoNativeBridge.badJwt
+         */
+        postMessage("""{"type":"BadJWT","data":{}}""")
+        verify { spyLog.warning(any()) }
+        verify(exactly = 0) { spyLog.error(any(), any<Throwable>()) }
+    }
+
+    @Test
     fun `malformed message throws an error`() {
         postMessage("sawr a warewolf with a chinese menu inhis hands")
         verify { spyLog.error(any(), any<JSONException>()) }
--- a/sdk/forms/src/test/java/com/klaviyo/forms/bridge/KlaviyoNativeBridgeTest.kt
+++ b/sdk/forms/src/test/java/com/klaviyo/forms/bridge/KlaviyoNativeBridgeTest.kt
@@ -454,12 +454,8 @@ internal class KlaviyoNativeBridgeTest : BaseTest() {
 
     @Test
     fun `BadJWT logs a warning and does not throw`() {
-        /**
-         * A rejected JWT is a normal outcome, so it should degrade gracefully rather than
-         * hitting the error-level catch block.
-         *
-         * @see com.klaviyo.forms.bridge.KlaviyoNativeBridge.badJwt
-         */
+        // A rejected JWT is a normal outcome, so it should degrade gracefully rather than
+        // hitting the error-level catch block. See KlaviyoNativeBridge.badJwt.
         postMessage("""{"type":"BadJWT","data":{}}""")
         verify { spyLog.warning(any()) }
         verify(exactly = 0) { spyLog.error(any(), any<Throwable>()) }
--- a/sdk/forms/src/test/java/com/klaviyo/forms/bridge/NativeBridgeMessageTest.kt
+++ b/sdk/forms/src/test/java/com/klaviyo/forms/bridge/NativeBridgeMessageTest.kt
@@ -390,6 +390,28 @@ class NativeBridgeMessageTest : BaseTest() {
     }
 
     @Test
+    fun `test decodeWebviewMessage decodes BadJWT type`() {
+        val badJwtMessage = """{"type": "BadJWT", "data": {}}"""
+
+        assertEquals(
+            NativeBridgeMessage.BadJwt,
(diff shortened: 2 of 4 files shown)


ab1470:
bugbot run

sdk/forms/src/test/java/com/klaviyo/forms/bridge/KlaviyoNativeBridgeTest.kt
```diff
@@ -452,6 +452,19 @@ internal class KlaviyoNativeBridgeTest : BaseTest() {
...
     }
 
+    @Test
+    fun `BadJWT logs a warning and does not throw`() {
+        /**
+         * A rejected JWT is a normal outcome, so it should degrade gracefully rather than
+         * hitting the error-level catch block.
+         *
+         * @see com.klaviyo.forms.bridge.KlaviyoNativeBridge.badJwt
+         */
```
glenn-klaviyo: any reason this style comment `/** */` over just inline `//` or `///`

glenn-klaviyo (approved):
LGTM

sdk/forms/src/test/java/com/klaviyo/forms/bridge/NativeBridgeMessageTest.kt
```diff
@@ -389,6 +389,28 @@ class NativeBridgeMessageTest : BaseTest() {
...
+            NativeBridgeMessage.BadJwt,
+            NativeBridgeMessage.decodeWebviewMessage(badJwtMessage)
+        )
+    }
+
+    @Test
+    fun `test decodeWebviewMessage BadJWT match is case-sensitive`() {
+        // Fender sends the exact PascalCase "BadJWT" wire type, which deliberately breaks the
+        // lower-camelCase keyName convention. A differently-cased variant must fall through to the
+        // unrecognized-type branch rather than silently matching.
```
amber-klaviyo, line 406: would it be a problem to match any variation if the meaning (bad jwt) is the same? I'm wondering if this will ever change on fender side, and if we would be better catching any version of this regardless of capitalization

amber-klaviyo (approved):

belleklaviyo (approved):

sdk/forms/src/test/java/com/klaviyo/forms/bridge/KlaviyoNativeBridgeTest.kt
ab1470: this was Claude Clauding but good catch. In Kotlin, `/** */` should be reserved for code documentation, while `//` is for code comments. This is a comment, not documentation, so it should be `//`. I'll fix it

sdk/forms/src/test/java/com/klaviyo/forms/bridge/NativeBridgeMessageTest.kt
ab1470: Nice catch, but I'd say we keep it as-is. The other bridge messages are decoded as case-sensitive, the same way this is. Also, Fender really shouldn't be randomly changing the capitalization on message keys. I'd like to be alerted (in the form of a console warning) in the event that Fender somehow does change the message key, even if the change is just a capitalization change

PR #506, opened 2026-07-08
feat(push): add X-Klaviyo-Sdk-Features header to push-token-register (MAGE-765)

Atharva8168:
# Description
Adds an `X-Klaviyo-Sdk-Features` header to `PushTokenApiRequest` only, carrying SDK feature-adoption flags for backend telemetry ([MAGE-766](https://linear.app/klaviyo/issue/MAGE-766)).

## Due Diligence

## Changelog / Code Overview
- `Config`/`KlaviyoConfig`: added `hasManifestKey(key): Boolean` (mirrors `getManifestInt`/`getManifestBoolean`) to distinguish "manifest key absent" from "present and false" — `getManifestBoolean`'s single default-value return can't tell those apart.
- `Constants`: added `DISABLE_AUTOMATIC_TOKEN_FORWARDING` (`com.klaviyo.disable_automatic_token_forwarding`) alongside the existing `AUTOMATIC_PUSH_TRACKING`.
- `PushTokenApiRequest`: attaches `X-Klaviyo-Sdk-Features` with `auto_push_tracking` and/or `auto_push_token_forwarding`, **each included only if its manifest key is present** in the host's manifest. Header is omitted entirely if neither key is set.
  - `auto_push_tracking` reflects `com.klaviyo.automatic_push_tracking` directly.
  - `auto_push_token_forwarding` mirrors `!com.klaviyo.disable_automatic_token_forwarding` directly (independent of the tracking flag, per product decision — reports the disable flag as-is).
- Scoped to `PushTokenApiRequest` only, not the shared `KlaviyoApiRequest` base — verified via a dedicated `EventApiRequestTest` assertion that the header never leaks to other request types.

## Test Plan
- `./gradlew :sdk:core:testDebugUnitTest :sdk:analytics:testDebugUnitTest` — new coverage in `KlaviyoConfigTest`, `PushTokenApiRequestTest` (all manifest-key presence/value combinations), `EventApiRequestTest` (header does not leak to other request types).
- `./gradlew test` (full suite) and `./gradlew ktlintCheck` both pass.
- Manual: ran the sample app on an emulator with each combination of `com.klaviyo.automatic_push_tracking` / `com.klaviyo.disable_automatic_token_forwarding` present/absent/true/false in the manifest, confirmed via `adb logcat` that the outgoing `client/push-tokens` request carries exactly the expected `X-Klaviyo-Sdk-Features` value (or omits it) in each case.

## Related Issues/Tickets
[MAGE-765](https://linear.app/klaviyo/issue/MAGE-765)

## Summary by CodeRabbit

* **New Features**
  * Added support for reporting SDK feature details in a request header when relevant manifest settings are present.
  * Improved manifest handling so the app can detect whether a setting exists, regardless of its value.

* **Bug Fixes**
  * Ensured only the intended request type includes the SDK feature header.
  * Added coverage for header generation across different manifest combinations and value states.

### DIFF
--- a/sdk/core/src/test/java/com/klaviyo/core/SdkFeaturesTest.kt
+++ b/sdk/core/src/test/java/com/klaviyo/core/SdkFeaturesTest.kt
@@ -1,0 +1,95 @@
+package com.klaviyo.core
+
+import com.klaviyo.fixtures.BaseTest
+import io.mockk.every
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertNull
+import org.junit.Test
+
+internal class SdkFeaturesTest : BaseTest() {
+
+    /**
+     * Simulate a host declaring [manifestKey] in the manifest with the given boolean [value].
+     * Absent keys are left at BaseTest's defaults (hasManifestKey → false).
+     */
+    private fun stubManifestKey(manifestKey: String, value: Boolean) {
+        every { mockConfig.hasManifestKey(manifestKey) } returns true
+        every { mockConfig.getManifestBoolean(manifestKey, false) } returns value
+    }
+
+    @Test
+    fun `Header omitted when no manifest keys are present`() {
+        assertNull(SdkFeatures.headerValue(SdkFeatureScope.PUSH_TOKEN_REGISTRATION))
+    }
+
+    @Test
+    fun `Reports only auto_push_tracking when only that key is present and true`() {
+        stubManifestKey(SdkFeatureKey.AUTO_PUSH_TRACKING.manifestKey, true)
+        assertEquals(
+            "auto_push_tracking=1;",
+            SdkFeatures.headerValue(SdkFeatureScope.PUSH_TOKEN_REGISTRATION)
+        )
+    }
+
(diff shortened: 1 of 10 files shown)


ajaysubra (commented):
more tests than code nice!

sdk/core/src/test/java/com/klaviyo/core/SdkFeaturesTest.kt
```diff
@@ -0,0 +1,95 @@
...
+        stubManifestKey(SdkFeatureKey.AUTO_PUSH_TRACKING.manifestKey, false)
+        // Forwarding-disable key left absent
+        assertEquals(
+            "auto_push_tracking=0;",
+            SdkFeatures.headerValue(SdkFeatureScope.PUSH_TOKEN_REGISTRATION)
+        )
+    }
+
+    @Test
+    fun `Every feature key resolves exactly one scope`() {
```
glenn-klaviyo, line 83: I suggest we drop this test, I don't think it validates very much that the compiler already does.

glenn-klaviyo (approved):
Overall LGTM. Tested locally e2e and looked good

PR #519, opened 2026-07-22
feat: API client network-behavior RCA items (MAGE-694, MAGE-496, MAGE-500)

evan-masseau:
# Description

Note: All this code was previously reviewed in individual PRs.

Consolidated API-client network-behavior fixes from the post-incident RCA: widen 5xx retry coverage, cap the request queue at 200 with oldest-first eviction, and wait `max(Retry-After, backoff)`. Merged down from the stacked chain into one feature branch targeting `rel/4.5.0`.

## Due Diligence

Public surface changes only via a `@Deprecated` no-op on `Config.Builder.networkFlushDepth(Int)` (source/binary compatible); all other changes are backwards-compatible behavior. Targeting `rel/4.5.0`.

## Changelog / Code Overview

- **Retry the entire 5xx range (MAGE-694).** Retry `429` + all `500–599` with no exclusions, so CDN/edge failures (e.g. Cloudflare `520–527`) like those seen during the cannot-access-klaviyo.com incident are retried instead of dropped. `501`/`505` are deliberately included — for the SDK's fixed request shapes a genuine origin 501/505 is unreachable, so any 5xx we see is edge noise worth retrying (data-retention > a wasted retry).
- **Align queue-depth behavior with iOS (MAGE-496).** Cap the request queue at **200** (`MAX_QUEUE_SIZE`); at capacity, evict the **oldest request by enqueue timestamp** (`queuedTime`) rather than the front of the deque, so freshly-inserted head-of-line/priority events are protected. Eviction drains fully to the cap even if the queue was already oversized (init-merge / in-flight reinsertion). **Retire the depth-triggered flush** — flushing is now interval + forced only, matching iOS. `Config.Builder.networkFlushDepth(Int)` is kept as a `@Deprecated` no-op (no source/binary break, no forced major bump).
- **Greater-of Retry-After vs exponential backoff (MAGE-500).** Wait `max(Retry-After, cappedExponentialBackoff)` instead of Retry-After alone, and raise the backoff cap **180s → 300s**. Damps sustained fleet-wide retry volume during rate-limit storms (esp. `Retry-After: 0` bursts) while never retrying sooner than the server asked.

## Test Plan

`:sdk:core` + `:sdk:analytics` unit suites green; eviction covered by mocked-clock tests (FIFO order, head-of-line protection, true 200-cap). ktlint clean. CI green on the merged branch.

**On-device UAT** — automated, Proxyman MITM fault-injection + mobile-mcp on the emulator (in-repo sample app, SDK 4.4.1 / Android 36). Full test plan + per-case evidence are posted in the **MAGE-957** issue comments.

| Case | Validates | Result |
|------|-----------|--------|
| TC-001 | Cloudflare 522 retried → delivered on recovery | ✅ Pass |
| TC-002 | 403 load-shed stays terminal (no retry) | ✅ Pass |
| TC-003 | 5xx bounds — 500/599 retried, 499/600 terminal | ✅ Pass |
| TC-004 | Transient network/IO failure retried | ✅ Pass |
| TC-005 | Retry-After honored as a floor | ✅ Pass |
| TC-006 | Exponential backoff on deep retries | ✅ Pass |
| TC-007 | 300s retry-interval ceiling | ⏭️ Skipped — unit-covered |
| TC-008 | Queue caps at 200, evicts oldest | ✅ Pass |
| TC-009 | Depth-flush trigger removed (no-op) | ✅ Pass |

No regressions observed; safe to land pending standard review/CI.

## Related Issues/Tickets

- **Test + release tracking:** MAGE-957 — full on-device UAT summary + per-case evidence are posted in the issue comments.
- **RCA items implemented here:** MAGE-694 (widen 5xx retry range), MAGE-496 (queue cap + evict-oldest), MAGE-500 (max(Retry-After, backoff) + 300s cap).

## Summary by CodeRabbit

* **Improvements**
  * Enforced a maximum queued analytics request limit; when exceeded, oldest requests are evicted (with a warning), while newer priority requests are protected.
  * Queue flushing now relies on the timer interval or explicit forcing (no depth-based gating).
  * Expanded retry logic to treat HTTP 429 and all HTTP 5xx (500–599) responses as retryable.
  * Retry delays now use the longer of exponential backoff and `Retry-After` (case-insensitive), with a higher max retry cap (5 minutes).
  * Debug sample builds now trust user-added CA certificates to enable HTTPS interception during testing.
* **Deprecations**
  * Depth-based flushing configuration (`networkFlushDepth`) is deprecated and removed from runtime behavior; it no longer affects behavior and is scheduled for future removal.

### DIFF
--- a/sample/src/debug/res/xml/network_security_config.xml
+++ b/sample/src/debug/res/xml/network_security_config.xml
@@ -4,9 +4,15 @@
   into release artifacts). Trust the system CA store plus user-added CAs so an
   intercepting proxy (e.g. Proxyman) can decrypt the sample's HTTPS traffic for
   network testing.
+
+  cleartextTrafficPermitted="true": attaching ANY network-security-config makes
+  Android ignore android:usesCleartextTraffic from the manifest, and cleartext
+  defaults to disallowed on targetSdk 28+. Re-permitting it here (debug only)
+  preserves plain-HTTP endpoints in debug builds (e.g. a local backend/proxy over
+  http). Release trust/cleartext behavior is unchanged.
 -->
 <network-security-config>
-    <base-config>
+    <base-config cleartextTrafficPermitted="true">
         <trust-anchors>
             <certificates src="system" />
             <certificates src="user" />
--- a/sample/src/debug/res/xml/network_security_config.xml
+++ b/sample/src/debug/res/xml/network_security_config.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  Debug-only source set: present in debug builds exclusively (never packaged
+  into release artifacts). Trust the system CA store plus user-added CAs so an
+  intercepting proxy (e.g. Proxyman) can decrypt the sample's HTTPS traffic for
+  network testing.
+-->
+<network-security-config>
+    <base-config>
+        <trust-anchors>
+            <certificates src="system" />
+            <certificates src="user" />
+        </trust-anchors>
+    </base-config>
+</network-security-config>
--- a/sample/src/debug/res/xml/network_security_config.xml
+++ b/sample/src/debug/res/xml/network_security_config.xml
@@ -4,12 +4,6 @@
   into release artifacts). Trust the system CA store plus user-added CAs so an
   intercepting proxy (e.g. Proxyman) can decrypt the sample's HTTPS traffic for
   network testing.
-
-  cleartextTrafficPermitted="true": attaching ANY network-security-config makes
-  Android ignore android:usesCleartextTraffic from the manifest, and cleartext
-  defaults to disallowed on targetSdk 28+. Re-permitting it here (debug only)
-  preserves plain-HTTP endpoints in debug builds (e.g. a local backend/proxy over
-  http). Release trust/cleartext behavior is unchanged.
 -->
 <network-security-config>
     <base-config cleartextTrafficPermitted="true">
--- a/sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
@@ -256,6 +256,17 @@ object KlaviyoConfig : Config {
             }
         }
 
+        @Deprecated(
+            message = "Depth-triggered flushing has been removed. The queue now flushes only on " +
+                "the timer interval (see networkFlushInterval) and is internally bounded by a " +
+                "size cap. This setter has no effect and will be removed in a future major release.",
+            level = DeprecationLevel.WARNING
+        )
+        override fun networkFlushDepth(networkFlushDepth: Int) = apply {
+            // No-op: depth-triggered flushing was removed; retained for one release as a
+            // deprecated setter so existing caller code keeps compiling.
+        }
+
         override fun build(): Config {
             val context = applicationContext ?: throw MissingContext()
             val packageInfo = context.packageManager.getPackageInfoCompat(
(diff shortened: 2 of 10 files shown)


sample/src/debug/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,21 @@
...
+  into release artifacts). Trust the system CA store plus user-added CAs so an
+  intercepting proxy (e.g. Proxyman) can decrypt the sample's HTTPS traffic for
+  network testing.
+
+  cleartextTrafficPermitted="true": attaching ANY network-security-config makes
+  Android ignore android:usesCleartextTraffic from the manifest, and cleartext
+  defaults to disallowed on targetSdk 28+. Re-permitting it here (debug only)
+  preserves plain-HTTP endpoints in debug builds (e.g. a local backend/proxy over
+  http). Release trust/cleartext behavior is unchanged.
+-->
```
evan-masseau, line 7:
```suggestion
  network testing.
-->
```

sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt
```diff
@@ -277,6 +257,18 @@ object KlaviyoConfig : Config {
...
         }
 
+        @Deprecated(
+            message = "Depth-triggered flushing has been removed. The queue now flushes on the " +
+                "timer interval (see networkFlushInterval) or when explicitly forced, and is " +
+                "internally bounded by a size cap. This setter has no effect and will be " +
+                "removed in a future major release.",
+            level = DeprecationLevel.WARNING
+        )
+        override fun networkFlushDepth(networkFlushDepth: Int) = apply {
```
Atharva8168, line 267: Nit: now that this is a no-op, should it still log an error/warning for out-of-range input (e.g. `networkFlushDepth <= 0`), the way the old setter did? As written, any input — including negative values — is silently swallowed with no feedback to the caller. Might be worth a debug-level log so callers who haven't migrated yet at least get a hint their config isn't taking effect.

evan-masseau: Good catch — done in 8474653630, and I went a level louder than your suggestion: `warning` rather than `debug`.

Reasoning: the caller stated an explicit configuration intent and we're silently discarding it, which is actionable on their end (delete the call). That reads as "degraded functionality" in the repo's logging table rather than routine config diagnostics, and it matches the deprecation convention. Happy to drop it to `debug` if you think warning is too loud for a setter that's harmless to leave in place.

The message names the setting, **echoes the value passed**, and points at the replacement — so your negative-value case now surfaces concretely instead of vanishing:

```
networkFlushDepth(-5) is deprecated and has no effect: depth-triggered flushing was removed.
The queue flushes on the timer interval (see networkFlushInterval) or when explicitly forced,
and is internally bounded by a size cap. Remove this call — the setter will be deleted in a
future major release.
```

I deliberately didn't re-add a separate `<= 0` validation branch: since the value is ignored entirely, one unconditional warning covers valid and out-of-range input alike, and range-checking a value we discard would just be noise. Added a unit test asserting a warning fires per call (one valid, one negative); `:sdk:core` suite and ktlint are green.

One implementation note in case it bites someone later: the sibling setters use the `"${KlaviyoConfig::prop.name} ..."` idiom, but that doesn't work here — `networkFlushDepth` no longer exists as a `KlaviyoConfig` property (only as the Builder function), so the name is a plain string literal.

-- Claude

Atharva8168 (approved):
Solid RCA consolidation — 5xx widening, queue cap/eviction, and max(Retry-After, backoff) are all well-reasoned and well-tested. Left one nit on the deprecated networkFlushDepth no-op.

PR #525, opened 2026-07-27
fix(push): collapse duplicate foreground push-token registration paths [MAGE-995]

evan-masseau:
# Description

On each foreground the SDK made two overlapping push-token calls, which could enqueue two `client/push-tokens` requests for a single foreground — the first carrying a token the provider had already rotated away from. This collapses them into one path, and moves the lifecycle hook off `ActivityEvent.Resumed` (which fires on every activity transition) onto `FirstStarted` (once per real foreground).

Suggested fix for a bug found while reviewing #517 — targets that feature branch so it can land before #517 merges to `rel/4.5.0`.

## Due Diligence

> **Not device-tested — flagging honestly.** This is verified by unit tests only. The bug's trigger condition (token rotation *and* a device-property change in the same background window) is difficult to reproduce naturally on a device; see Test Plan for how to force it if someone wants device confirmation before merge. Android compatibility: no new platform APIs are used, and `ActivityEvent.FirstStarted` is pre-existing SDK infrastructure.

> Marked `Minor` because `State` gains `refreshPushState()` and `PushTokenFetcher.fetchAndSetPushToken` gains a defaulted parameter. Both are SDK-internal-by-convention types that are `public` only because Kotlin's `internal` is module-scoped. Net effect on the *released* API surface is nil: `PushTokenFetcher` is itself new in 4.5.0 via #517 and has never shipped. Existing call sites are unaffected (default arg), and no host app implements `State`.

### The bug

`StateSideEffects.onLifecycleEvent` called **both**:
1. `Klaviyo.setPushToken(storedToken)` — to force recomputation of push state so notification-permission changes get picked up, and
2. `Klaviyo.maybeAutoRegisterPushToken()` — which asynchronously fetches a fresh token and calls `setPushToken` again.

Push-state dedup is string equality on the entire serialized `PushTokenApiRequest` body, and that body embeds device values injected at access time by the `requestBody` getter (`ENABLEMENT_STATUS` from `DeviceProperties.notificationPermissionGranted`, and `BACKGROUND`). So the two calls can produce two *different* bodies.

Requests enqueued per foreground:

| Scenario | Sync `setPushToken(stored)` | Async fetch | Before | After |
|---|---|---|---|---|
| Nothing changed | no-op | no-op | 0 | 0 |
| Permission changed only | enqueue | no-op | 1 | 1 |
| Token rotated only | no-op | enqueue | 1 | 1 |
| **Both changed** | **enqueue (stale token)** | **enqueue (fresh)** | **2** | **1** |

Row 4 is the defect. Final state converges — the fresh token supersedes the stale one — so it self-corrects, but it burns an extra request and momentarily attaches a dead token to the profile. Worth avoiding on this endpoint in particular given the prior Wyze push-token 429 incident.

### The fix

1. **Mutually exclusive paths.** Attempt the fetch first; recompute from the stored token only when no fetch will deliver one (forwarding disabled, `push-fcm` absent, or the fetch failed).

2. **New `State.refreshPushState()`.** Recomputing derived state becomes an explicit operation instead of a side effect of re-assigning a value already in state. It no-ops without a token: `PersistentObservableString` rejects empty values outright, so the previous `?: ""` could never clear push state and only emitted a spurious `"Empty string value will be ignored"` warning on every foreground for integrators with no push token.

3. **New `onUnavailable` callback on `PushTokenFetcher`.** Routes an async fetch failure back to `refreshPushState()`, so a misconfigured Firebase doesn't silently stop permission changes from being detected. Defaulted, so existing call sites are unchanged.

4. **`ActivityEvent.Resumed` → `FirstStarted`.** `KlaviyoLifecycleMonitor.onActivityResumed` broadcasts `Resumed` unconditionally, so a `FirebaseMessaging.getInstance().token` IPC and an uncached `PackageManager` binder read fired on *every* in-app navigation. `FirstStarted` is gated on `activeActivities == 0`, matching the intent of the original code comment ("re-pull the token on foreground").

Additionally, `maybeAutoRegisterPushToken` moved from the public `Klaviyo` facade onto `PushTokenFetcher`'s companion object in `core`. It only ever touched `Registry` and `Constants` — both core-owned — so this keeps the policy beside the abstraction it governs and removes internal orchestration from the public API surface.

### Reviewer note

The `onUnavailable` callback (#3) is the one arguably optional piece. It covers a narrow case: broken Firebase config **and** a manually-set token **and** reliance on permission detection. Without it that combination stops detecting permission changes, which is a regression against today's unconditional refresh — but if the team would rather keep `PushTokenFetcher` minimal, it can be dropped independently of the rest.

## Test Plan

Automated (all passing):

```
./gradlew :sdk:analytics:testDebugUnitTest :sdk:core:testDebugUnitTest :sdk:push-fcm:testDebugUnitTest ktlintCheck
```

| Suite | Result |
|---|---|
| `PushTokenFetcherTest` (new, core) | 4/4 |
| `KlaviyoTest` | 54/54 |
| `StateSideEffectsTest` | 21/21 |
| `KlaviyoStateTest` | 29/29 |
| `KlaviyoPushTokenFetcherTest` | 7/7 |
| `ktlintCheck` (full project) | BUILD SUCCESSFUL |

Key regression guards:
- `FirstStarted` with forwarding enabled asserts `refreshPushState()` is **not** called — this is the anti-double-report assertion.
- `Resumed` asserts neither path fires, proving the hook moved.
- Fetcher-reports-unavailable asserts `refreshPushState()` runs **exactly once**, not "at least once".

To reproduce row 4 manually on a device:
1. Integrate with `automatic_push_token_forwarding` enabled (the 4.5.0 default) and grant notification permission.
2. Background the app.
3. Revoke notification permission in system settings, and force an FCM token rotation (clear Google Play Services app data, or call `FirebaseMessaging.deleteToken()` from a debug hook).
4. Foreground the app and observe outgoing requests.

Before: two `client/push-tokens` requests, the first with the old token. After: one, with the fresh token.

## Related Issues/Tickets

- Closes MAGE-995
- Follow-up to #517 — found during cross-platform review against the Swift counterpart (klaviyo/klaviyo-swift-sdk#665)

## Summary by CodeRabbit

* **New Features**
  * Automatic push-token forwarding is now centralized and can report when token retrieval is unavailable.
  * Push state refreshes automatically when the app first starts, including changes to notification permissions or device conditions.
  * Push-token retrieval failures are handled without interrupting SDK initialization or app execution.

* **Bug Fixes**
  * Push state is no longer cleared unnecessarily when no token is available.
  * Improved resilience when Firebase or token forwarding encounters errors.

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
@@ -16,7 +16,6 @@ import com.klaviyo.analytics.networking.KlaviyoApiClient
 import com.klaviyo.analytics.state.KlaviyoState
 import com.klaviyo.analytics.state.State
 import com.klaviyo.analytics.state.StateSideEffects
-import com.klaviyo.core.Constants
 import com.klaviyo.core.Constants.PACKAGE_PREFIX
 import com.klaviyo.core.Constants.TRACKING_PARAMETER
 import com.klaviyo.core.Operation
@@ -114,43 +113,7 @@ object Klaviyo {
         }
 
         // Optional side effect, kept last so it can never interfere with core initialization
-        maybeAutoRegisterPushToken()
-    }
-
-    /**
-     * Automatic push token registration (via [Constants.AUTOMATIC_PUSH_TOKEN_FORWARDING],
-     * default [Constants.AUTOMATIC_PUSH_TOKEN_FORWARDING_DEFAULT]): pull the current token and forward
-     * it to Klaviyo. Called from [initialize] and on each app foreground so rotations are picked up.
-     * No-op when the flag is explicitly off or when `push-fcm` is absent.
-     *
-     * Reads the flag and default that also govern `KlaviyoPushService.onNewToken` (which reads them
-     * from its own Context, safe before init), so the two automatic-collection paths can't drift.
-     *
-     * Independent of [Constants.AUTOMATIC_PUSH_OPEN_TRACKING] (which gates only automatic open tracking) —
-     * this flag alone controls token forwarding.
-     */
-    internal fun maybeAutoRegisterPushToken() {
-        // Reading via Registry.config is safe here: this only runs from initialize() and on foreground,
-        // both post-initialization (unlike KlaviyoPushService.onNewToken, which uses its Context).
-        val forwardingEnabled = Registry.config.getManifestBoolean(
-            Constants.AUTOMATIC_PUSH_TOKEN_FORWARDING,
-            Constants.AUTOMATIC_PUSH_TOKEN_FORWARDING_DEFAULT
-        )
-        if (!forwardingEnabled) {
-            Registry.log.verbose(
-                "Skipping automatic push token registration (automaticTokenForwarding=false)"
-            )
-            return
-        }
-
-        Registry.getOrNull<PushTokenFetcher>()?.also { fetcher ->
-            Registry.log.debug("Automatically fetching push token")
-            // Contain any fetcher failure so this optional side effect cannot disrupt the caller
-            runCatching { fetcher.fetchAndSetPushToken() }
-                .onFailure { Registry.log.warning("Automatic push token fetch failed", it) }
-        } ?: Registry.log.verbose(
-            "No push token fetcher registered; skipping automatic registration"
-        )
+        PushTokenFetcher.maybeAutoRegisterPushToken()
     }
 
     /**
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/state/StateSideEffects.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/state/StateSideEffects.kt
@@ -1,6 +1,5 @@
 package com.klaviyo.analytics.state
 
-import com.klaviyo.analytics.Klaviyo
 import com.klaviyo.analytics.model.ImmutableProfile
 import com.klaviyo.analytics.model.Profile
 import com.klaviyo.analytics.model.ProfileKey
@@ -12,10 +11,12 @@ import com.klaviyo.analytics.networking.requests.KlaviyoApiRequest.Companion.HTT
 import com.klaviyo.analytics.networking.requests.KlaviyoErrorResponse
 import com.klaviyo.analytics.networking.requests.KlaviyoErrorSource
 import com.klaviyo.analytics.networking.requests.PushTokenApiRequest
+import com.klaviyo.core.PushTokenFetcher
 import com.klaviyo.core.Registry
 import com.klaviyo.core.config.Clock
 import com.klaviyo.core.lifecycle.ActivityEvent
 import com.klaviyo.core.lifecycle.LifecycleMonitor
+import com.klaviyo.core.safeApply
 import com.klaviyo.core.utils.takeIf
 
 internal class StateSideEffects(
@@ -174,14 +175,31 @@ internal class StateSideEffects(
         }
     }
 
+    /**
+     * Bring push state up to date when the app returns to the foreground, via exactly one path.
+     *
+     * Two things can go stale while backgrounded: the push token (the provider may rotate it) and
+     * the device properties embedded in push state (notification permission, background
+     * availability). A successful token fetch already covers both, because assigning the token
+     * recomputes push state — so the fetch is attempted first, and a direct refresh runs only when
+     * no fetch will deliver one (forwarding disabled, `push-fcm` absent, or the fetch failed).
+     *
+     * Doing both unconditionally would double-report: when the token *and* a device property have
+     * both changed, refreshing from the token already in state enqueues a push-token request
+     * carrying the token the provider has already rotated away from, which the newly fetched token
+     * then immediately supersedes — two requests for one foreground, the first already stale.
+     */
     private fun onLifecycleEvent(activity: ActivityEvent) {
-        activity.takeIf<ActivityEvent.Resumed>()?.run {
-            Registry.get<State>().pushToken?.let {
-                // Trigger the token in state to refresh overall push state, if changed
-                Klaviyo.setPushToken(it)
+        // FirstStarted, not Resumed: Resumed broadcasts on every activity transition, whereas
+        // FirstStarted is gated on activeActivities == 0, so this runs once per actual foreground.
+        activity.takeIf<ActivityEvent.FirstStarted>()?.run {
+            safeApply {
+                val refreshFromStoredToken = { state.refreshPushState() }
+
+                if (!PushTokenFetcher.maybeAutoRegisterPushToken(refreshFromStoredToken)) {
+                    refreshFromStoredToken()
+                }
             }
-            // Re-pull the token on foreground so rotations while backgrounded are caught
-            Klaviyo.maybeAutoRegisterPushToken()
         }
     }
 }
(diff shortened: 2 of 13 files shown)


evan-masseau:
bugbot run

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoPushTokenFetcher.kt
```diff
@@ -9,19 +9,21 @@ import com.klaviyo.core.Registry
...
  */
 internal class KlaviyoPushTokenFetcher : PushTokenFetcher {
-    override fun fetchAndSetPushToken() {
+    override fun fetchAndSetPushToken(onUnavailable: () -> Unit) {
         try {
             FirebaseMessaging.getInstance().token
                 .addOnSuccessListener { token -> Klaviyo.setPushToken(token) }
                 .addOnFailureListener { e ->
                     Registry.log.warning("Failed to fetch push token for automatic registration", e)
+                    onUnavailable()
```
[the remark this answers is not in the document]
evan-masseau: Good catch — confirmed and fixed in 6bf77d2256.

Traced it to be sure: the success path was already safe because `Klaviyo.setPushToken` is itself `= safeApply { }`, so only the new failure path was exposed. By the time `addOnFailureListener` fires, both the `safeApply` in `StateSideEffects.onLifecycleEvent` and the `runCatching` in `maybeAutoRegisterPushToken` have unwound, and `refreshPushState()` reads device properties over binder — so a throw would have landed on whichever thread the Task completed on.

Guarded at both ends rather than just one:

- **Definition site** (`StateSideEffects`) — the callback is now self-contained: `val refreshFromStoredToken: () -> Unit = { safeApply { state.refreshPushState() } }`, so it's safe regardless of who invokes it or when.
- **Boundary** (`KlaviyoPushTokenFetcher`) — a private `notifyUnavailable` wraps the invocation in `runCatching`, so `fetchAndSetPushToken`'s must-not-throw contract holds no matter what a caller hands it. Applied on both the async and synchronous failure paths.

Also documented on the `PushTokenFetcher` interface that `onUnavailable` may be invoked asynchronously, off the caller's stack, so future callers don't assume an enclosing `safeApply` covers them.

Two regression tests added (`KlaviyoPushTokenFetcherTest` 7 → 9): a throwing `onUnavailable` is contained on the async failure path, and when `FirebaseMessaging.getInstance()` itself throws. Full suite green, ktlint clean.

evan-masseau:
bugbot run

evan-masseau:
bugbot run

evan-masseau:
bugbot run

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -114,43 +113,7 @@ object Klaviyo {
         }
 
         // Optional side effect, kept last so it can never interfere with core initialization
-        maybeAutoRegisterPushToken()
```
evan-masseau, line 117: This chunk is just code organization really

sdk/analytics/src/main/java/com/klaviyo/analytics/state/StateSideEffects.kt
```diff
@@ -174,14 +175,29 @@ internal class StateSideEffects(
...
+     * both changed, refreshing from the token already in state enqueues a push-token request
+     * carrying the token the provider has already rotated away from, which the newly fetched token
+     * then immediately supersedes — two requests for one foreground, the first already stale.
+     */
     private fun onLifecycleEvent(activity: ActivityEvent) {
-        activity.takeIf<ActivityEvent.Resumed>()?.run {
-            Registry.get<State>().pushToken?.let {
-                // Trigger the token in state to refresh overall push state, if changed
-                Klaviyo.setPushToken(it)
+        // FirstStarted, not Resumed: Resumed broadcasts on every activity transition, whereas
```
evan-masseau, line 193: This is the main bug that I noticed and wanted to avoid. Refreshing push state on resume AND refreshing token from FCM would be duplicative, so we only need to do one or the other.

evan-masseau: State machinery under the hood would generally have de-duped in this situation anyway, but not always,

glenn-klaviyo: Nice and good to know. I had assumed it was ok given our state machine de-dupe but if it's not always then this is much better 👍

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
glenn-klaviyo: I like this clean up / organization!

sdk/core/src/main/java/com/klaviyo/core/PushTokenFetcher.kt
```diff
@@ -9,6 +9,66 @@ interface PushTokenFetcher {
...
+     *
+     * May be invoked **asynchronously**, after this method has returned and off the caller's
+     * stack — so any guard wrapping the call site will no longer be in scope. Implementations
+     * must contain failures it raises to honor the must-not-throw contract above, and callers
+     * should not assume an enclosing `safeApply` will catch them.
      */
-    fun fetchAndSetPushToken()
+    fun fetchAndSetPushToken(onUnavailable: () -> Unit = {})
+
+    companion object {
```
Atharva8168, line 25: Good relocation — this policy belongs beside the abstraction it governs, not on the public facade.

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -114,43 +113,7 @@ object Klaviyo {
...
-
-        Registry.getOrNull<PushTokenFetcher>()?.also { fetcher ->
-            Registry.log.debug("Automatically fetching push token")
-            // Contain any fetcher failure so this optional side effect cannot disrupt the caller
-            runCatching { fetcher.fetchAndSetPushToken() }
-                .onFailure { Registry.log.warning("Automatic push token fetch failed", it) }
-        } ?: Registry.log.verbose(
-            "No push token fetcher registered; skipping automatic registration"
-        )
+        PushTokenFetcher.maybeAutoRegisterPushToken()
```
Atharva8168, line 116: Nice trim on the facade — 37 fewer lines in the first file integrators read.

sdk/core/src/test/java/com/klaviyo/core/PushTokenFetcherTest.kt
```diff
@@ -0,0 +1,66 @@
+package com.klaviyo.core
+
+import com.klaviyo.fixtures.BaseTest
+import io.mockk.every
+import io.mockk.verify
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Test
+
+internal class PushTokenFetcherTest : BaseTest() {
```
Atharva8168, line 10: Test counts in the PR description are stale — this suite is listed as 4/4 but has 5 tests. Worth updating the Test Plan table so the numbers match what actually runs.

Atharva8168 (approved):

glenn-klaviyo (approved):
LGTM, did some manual verification on this branch

PR #528, opened 2026-07-31
fix(push): keep action buttons when the open_url link is unusable (PUSH-1095)

josephzhou-klaviyo:
# Description
An `Open External URL` action button configured with a scheme-less link (e.g. `www.cnn.com`) rendered **no action buttons at all** on Android. The `open_url` scheme allowlist was being used as a render-time eligibility gate; url-shaped buttons with an unusable url now render as a new `ActionButton.Degraded` variant that launches the app, instead of being dropped.

Android counterpart to klaviyo/klaviyo-swift-sdk#691 (PUSH-1085).

## Due Diligence

> Not yet verified on a device/emulator — `ktlint` + the full `push-fcm` unit suite pass locally (see Test Plan). No version-specific APIs were introduced.

### ⚠️ Reviewer decision needed: sealed-class source compatibility
`object KlaviyoRemoteMessage` and its nested `sealed class ActionButton` are **public**, and `RemoteMessage.actionButtons` is a public extension property — so this is a real consumer surface, not hypothetical.

Adding the `Degraded` variant is **source-breaking** for any host app with an *exhaustive* `when` over `ActionButton`: their code will stop compiling until they add a branch. Nothing breaks at runtime, and binary compatibility is unaffected.

I judged this worth it, because a distinct variant is what makes the compiler force every dispatch site to handle degraded buttons — that exhaustiveness is the safety property this fix leans on. **If we'd rather keep this a `Patch`,** the alternative is a boolean flag plus declared-action/url fields on `OpenApp`, which gets identical analytics parity with no API break, at the cost of losing the compiler guarantee. Happy to switch if you prefer.

### Root cause
`"www.cnn.com".toUri().scheme` is `null`, which is not in `ALLOWED_OPEN_URL_SCHEMES`, so `hasAllowedOpenUrlScheme()` failed and the `TYPE_OPEN_URL` branch of `RemoteMessage.actionButtons` returned `null` — dropping the button. When it was the only button, `buttons.takeIf { it.isNotEmpty() }` returned `null` and `addActionButtons` returned early, so the notification rendered with **no buttons at all**.

Same layering mistake as iOS: **a dispatch-time security allowlist was doing render-time eligibility work.**

### Why Android's fix differs from iOS
This is the important review point. On iOS the `UNNotificationAction` carries no URL and `.foreground` opens the app regardless, so the allowlist can simply move to tap time.

Android has no equivalent seam. `KlaviyoNotification.createButtonAction` bakes the `PendingIntent` at **render time** (`is ActionButton.OpenUrl -> KlaviyoTrampolineActivity.forBrowserUrl(context, button.url)`), and `KlaviyoTrampolineActivity.dispatchDestination` hands the url straight to `DeepLinking.makeExternalIntent(...).startActivityIfResolved(context)` with no re-check. So the allowlist **must** stay in the parser. Keeping a bad url as `OpenUrl` would either bake a browser intent for `javascript:alert(1)`, or — for `www.cnn.com` — produce an intent nothing resolves, leaving a button that does *nothing at all*.

Hence `Degraded`: the button renders, a tap launches the app, and the unusable url structurally cannot reach `makeExternalIntent` because it never becomes an `OpenUrl` instance.

### The change
`sdk/push-fcm/.../KlaviyoRemoteMessage.kt`
- New `ActionButton.Degraded(id, label, declaredAction, declaredUrl)`.
- `TYPE_DEEP_LINK` missing url, `TYPE_OPEN_URL` missing url, and `TYPE_OPEN_URL` disallowed/missing scheme now produce `Degraded` instead of `null`. Structural problems (missing `id`/`label`, unrecognized `action`) still skip the button.
- `appendActionButtonExtras` reports `declaredAction`/`declaredUrl`, so `$opened_push` still carries `Button Action = "Open URL"` and `Button Link = "www.cnn.com"` — **matching iOS**. An earlier revision collapsed to `OpenApp`, which reported `"Open App"` with no `Button Link`; that was rejected because it hid the misconfiguration from analytics and disagreed with iOS about the same message.
- The link-selection `when` enumerates the sealed arms rather than matching `UrlBearing`, so adding a variant is a compile error rather than a silent `null`. `Degraded` deliberately does **not** implement `UrlBearing` — its `declaredUrl` is precisely the url we refused to dispatch, so it must not look dispatchable.

`sdk/push-fcm/.../KlaviyoNotification.kt`
- `is ActionButton.OpenApp, is ActionButton.Degraded ->` share the app-launch branch.
- Verbose log reports the declared action and notes the degradation.

**Unchanged, deliberately:** `ALLOWED_OPEN_URL_SCHEMES`, `hasAllowedOpenUrlScheme()`, `RemoteMessage.webUrl` (the body-tap gate), and `KlaviyoTrampolineActivity`.

### Why urls are not normalized to `https://`
Fender's validation (klaviyo/fender#66002) blocks scheme-less input rather than normalizing. Normalizing here would contradict that contract and diverge from iOS. That fender fix is editor-only — it notes API-created messages bypass validation entirely — which is why the SDK must still degrade gracefully.

### Automated — `./gradlew :sdk:push-fcm:ktlintCheck :sdk:push-fcm:testDebugUnitTest`
```
BUILD SUCCESSFUL
```

| Suite | Tests | Failures |
|---|---|---|
| `KlaviyoNotificationTest` | 44 | 0 |
| `KlaviyoRemoteMessageTest` | 38 | 0 |
| `KlaviyoTrampolineActivityTest` | 9 | 0 |
| `KlaviyoPushServiceTest` | 7 | 0 |
| `KlaviyoPushTokenFetcherTest` | 9 | 0 |
| `KlaviyoPushTokenInitProviderTest` | 1 | 0 |
| **Total** | **108** | **0** |

**Added**
- `actionButtons downgrades open_url with scheme-less url to a degraded app-launch button` — the exact ticket repro, asserting `declaredAction`/`declaredUrl` survive.
- `actionButtons downgrades open_url with blank url to a degraded app-launch button` — closes a gap that predates this PR (no test ever exercised a truly blank `""` url; `optNonBlankString` collapses blank and null today, so this pins that).
- `appendActionButtonExtras for Degraded reports the declared action and link` and `… omits the link when the payload had none`.
- `degraded action button routes to app launch and never to the browser` (`KlaviyoNotificationTest`) — the **security regression guard**: a `Degraded` button with `javascript:alert(1)` must hit `forDestination` and never `forBrowserUrl`, while still emitting the declared analytics extras.

**Flipped (6)** — these pinned the drop-the-button bug; renamed and re-asserted against `Degraded`: two `DEEP_LINK … without/null URL`, the mixed valid/invalid parser test, and three `open_url` blocked-scheme / missing-url / null-url tests.

**Also adjusted:** the `MAX_ACTION_BUTTONS` cap test now uses an unsupported action type for its "invalid button" case, since a url-less `deep_link` no longer drops. Degraded buttons correctly occupy cap slots (shared `buttons.add` path).

### Manual repro (for the reviewer / QA)
1. Push with an action button, action = **Open External URL**, url = `www.cnn.com`.
2. Send preview, expand the notification.
   - **Before:** no buttons at all. **After:** the button renders.
3. Tap it → the app launches (a browser is not opened). Logcat: `OPEN_URL action button 0 has disallowed scheme 'null'; rendering as an app-launch button instead`.
4. Check the `$opened_push` event carries `Button Action = "Open URL"` and `Button Link = "www.cnn.com"`.
5. Regression: `https://www.cnn.com` still opens the browser; `mailto:`/`tel:`/`sms:`/`smsto:` still work; `javascript:alert(1)` renders a button that only launches the app.

## Related Issues/Tickets
- Related to PUSH-1095
- iOS counterpart: klaviyo/klaviyo-swift-sdk#691 (Related to PUSH-1085)
- Sender-side validation: klaviyo/fender#66002 (PUSH-1084)

### DIFF
--- a/sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
+++ b/sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoNotification.kt
@@ -372,6 +372,9 @@ class KlaviyoNotification(private val message: RemoteMessage) {
                 is ActionButton.DeepLink -> ActionButton.DISPLAY_NAME_DEEP_LINK to " (scheme: ${button.url.toUri().scheme})"
                 is ActionButton.OpenUrl -> ActionButton.DISPLAY_NAME_OPEN_URL to " (scheme: ${button.url.toUri().scheme})"
                 is ActionButton.OpenApp -> ActionButton.DISPLAY_NAME_OPEN_APP to ""
+                is ActionButton.Degraded ->
+                    button.declaredAction to
+                        " (degraded to app launch${button.declaredUrl?.let { ": $it" } ?: ""})"
             }
             Registry.log.verbose(
                 "Added action button $index: '${button.label}' ($actionType)$destination"
@@ -414,7 +417,9 @@ class KlaviyoNotification(private val message: RemoteMessage) {
                 // and dismisses the notification — the browser would otherwise swallow the intent.
                 KlaviyoTrampolineActivity.forBrowserUrl(context, button.url)
             }
-            is ActionButton.OpenApp -> {
+            // A degraded button's declared destination is unusable, so it launches the app.
+            // Its declared action/url are still reported via appendActionButtonExtras.
+            is ActionButton.OpenApp, is ActionButton.Degraded -> {
                 if (autoTracking) {
                     KlaviyoTrampolineActivity.forDestination(context)
                 } else {
--- a/sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoRemoteMessage.kt
+++ b/sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoRemoteMessage.kt
@@ -65,12 +65,22 @@ object KlaviyoRemoteMessage {
             is ActionButton.DeepLink -> ActionButton.DISPLAY_NAME_DEEP_LINK
             is ActionButton.OpenUrl -> ActionButton.DISPLAY_NAME_OPEN_URL
             is ActionButton.OpenApp -> ActionButton.DISPLAY_NAME_OPEN_APP
+            // Report what the sender declared, not the app-launch we degraded to.
+            is ActionButton.Degraded -> button.declaredAction
         }
         putExtra(PACKAGE_PREFIX + "Button Action", actionName)
 
-        (button as? ActionButton.UrlBearing)?.let {
-            putExtra(PACKAGE_PREFIX + "Button Link", it.url)
+        // Enumerated over the sealed arms rather than matching on [ActionButton.UrlBearing], so
+        // adding a variant is a compile error here instead of silently yielding no link.
+        // [ActionButton.UrlBearing] remains part of the public surface for consumers that want
+        // to test for a destination without enumerating variants themselves.
+        val link = when (button) {
+            is ActionButton.DeepLink -> button.url
+            is ActionButton.OpenUrl -> button.url
+            is ActionButton.Degraded -> button.declaredUrl
+            is ActionButton.OpenApp -> null
         }
+        link?.let { putExtra(PACKAGE_PREFIX + "Button Link", it) }
     }
 
     /**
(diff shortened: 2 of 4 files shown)


josephzhou-klaviyo:
bugbot run

PR #529, opened 2026-08-03
fix(push): keep trampoline from clearing DeepLinkHandler navigation (MAGE-1004)

Atharva8168:
# Description

With `automatic_push_open_tracking` enabled, the notification trampoline invoked the host's registered `DeepLinkHandler` and then started the host launcher intent with `CLEAR_TOP`, destroying whatever the handler had just navigated to. The launcher intent now only brings the host task to the front when the handler owns navigation.

## Due Diligence

  - `FLAG_ACTIVITY_NEW_TASK` is available on every supported level. The flags are *assigned* rather than added specifically to avoid `Intent.removeFlags`, which is API 26 and above our minSdk of 23.

  - Targeting `rel/4.5.0` rather than `master`, so the README change only goes live with the 4.5.0 release.
  - Fixes a defect in the unreleased automatic push integration, so it should ship in 4.5.0 alongside the feature it affects.

No version bump in this PR — deferred to the release-branch merge per convention.

## Changelog / Code Overview

**Root cause.** `KlaviyoTrampolineActivity.handleTrampolineIntent` calls `Klaviyo.handlePush(intent)`, which reaches `DeepLinking.handleDeepLink` and invokes a registered `DeepLinkHandler` via `runOnUiThread`. Since the trampoline's `onCreate` is already on the main looper, that runs **inline** — the host navigates before the trampoline continues. `startDestination` then started the launcher intent with `NEW_TASK | SINGLE_TOP | CLEAR_TOP`:

- `CLEAR_TOP` on an `ACTION_MAIN`/`CATEGORY_LAUNCHER` intent finishes every Activity above the launcher Activity, destroying the handler's destination.
- `SINGLE_TOP` re-delivers `onNewIntent(ACTION_MAIN)` to the launcher Activity, which single-Activity apps commonly treat as a fresh home launch and use to reset routing.

Deterministic, not a race. Only affects hosts that enable the flag **and** register a handler.

**Fix.** `startDestination` now hoists `handlerOwnsNavigation` (a deep link is present *and* a handler is registered) and branches the flags on it:

- Handler owns navigation → `flags = FLAG_ACTIVITY_NEW_TASK` alone. Launcher-icon semantics: resume the existing task exactly as the user left it. Nothing finished, no `onNewIntent` delivered.
- Every other branch (`ACTION_VIEW` into the host, unresolvable deep-link fallback, `open_app`) → unchanged `NEW_TASK | SINGLE_TOP | CLEAR_TOP`.

Notes on the implementation:

- Flags are **assigned**, not added, because `DeepLinking.makeLaunchIntent` bakes in `SINGLE_TOP` and `Intent.removeFlags` is above our minSdk. Verified against AOSP that `getLaunchIntentForPackage` sets exactly `NEW_TASK` (with `setClassName` after), so action, category, package, and component all survive the overwrite.
- `makeLaunchIntent`'s public signature is untouched — a new parameter would be a binary break, and its two other callers legitimately want `CLEAR_TOP` + `SINGLE_TOP`.
- Extras are still forwarded. `Intent.filterEquals` ignores them, so they can't affect task matching, and on cold start they carry the delivery id that dedupes against a host also calling `handlePush`.

**Tests.** Intent flags now have coverage for the first time. The renamed handler-registered test is the regression test; the three untouched branches gained assertions that they keep the old flags; and a new test pins that `open_app` with a handler registered still resets to root (guarding against a "simplification" of the condition to just `isHandlerRegistered`).

**Sample app.** The deep link handler previously only showed a toast, which `CLEAR_TOP` cannot pop — so the sample could not reproduce this bug or demonstrate the fix. It now navigates to a new `SampleDetailActivity`, producing the `[SampleActivity, SampleDetailActivity]` stack the bug destroyed. Sample-only; nothing published changes.

**Docs.** The handler-owns-navigation contract was undocumented; added a note to the README's deep linking section, including that hosts should route from the handler callback rather than `onNewIntent` on this path.

## Test Plan

Unit:

```
./gradlew :sdk:push-fcm:testDebugUnitTest --tests "com.klaviyo.pushFcm.KlaviyoTrampolineActivityTest"
```

Full `./gradlew test` and `./gradlew ktlintCheck` are green.

To confirm the regression test is not vacuous, revert only the production file and re-run — exactly one test fails (the new regression test) and the other nine pass:

```
git checkout HEAD~1 -- sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoTrampolineActivity.kt
```

Device (install the `automatic` flavor — flag on, and the sample always registers a handler):

```
./gradlew :sample:installAutomaticDebug
```

Send a push with `url` = `klaviyotest://detail`, then:

1. Tap it → `SampleDetailActivity` opens on top of `SampleActivity`.
2. Press Home so the detail screen stays top of the task.
3. Send and tap a second identical push.
4. **Expected: still on `SampleDetailActivity`.** Before the fix, `CLEAR_TOP` finished it and you land back on `SampleActivity`.

Verified. The trampoline log line distinguishes the paths: `Trampoline dispatched deep link via handler; bringing host to front` (fixed) vs `Trampoline dispatching deep link via handler; launching host` (previous). Stack shape via:

```
adb shell dumpsys activity activities | grep -E "ActivityRecord.*com.klaviyo.sample"
```

Reviewer note — remaining scenarios worth a second pair of eyes:

- **Task created three different ways** (launcher icon / previous notification tap / external `klaviyotest://` link), confirming the launcher intent brings the task forward rather than stacking a duplicate launcher Activity. This depends on task base-intent matching (`Intent.filterEquals` against the task's base intent) and can vary by OEM, so it is worth checking on a non-Pixel device. It is the one assumption that could not be settled from source.
- **`open_app` push** (no `url`) should still reset to root — the deliberately unchanged branch.
- **Cold start** ends with `SampleActivity` on top and `SampleDetailActivity` behind it. That is MAGE-1020, **not** a regression from this PR — `CLEAR_TOP` is a no-op with no task to clear, so before/after are identical.

## Related Issues/Tickets

- MAGE-1004 — https://linear.app/klaviyo/issue/MAGE-1004
- Originating review thread: https://github.com/klaviyo/klaviyo-android-sdk/pull/517#discussion_r3675630546
- MAGE-1020 — https://linear.app/klaviyo/issue/MAGE-1020 — related but **not** fixed here: on cold start the handler fires before any Activity exists, so the link is dropped. Independent of this fix and needs its own design.

## Summary by CodeRabbit

* **New Features**
  * Deep links handled by a registered handler now open the destination while preserving the existing app navigation state.
  * The sample app includes a detail screen for viewing deep-linked content, with back navigation.
* **Bug Fixes**
  * Open-app actions continue to reset navigation appropriately, while handler-managed deep links avoid unnecessary relaunches or back-stack clearing.
* **Documentation**
  * Expanded deep-linking guidance to explain handler and default intent behavior.

### DIFF
--- a/sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoTrampolineActivity.kt
+++ b/sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoTrampolineActivity.kt
@@ -127,13 +127,23 @@ internal class KlaviyoTrampolineActivity : Activity() {
          *
          * - Deep link present and no [DeepLinkHandler][com.klaviyo.analytics.linking.DeepLinkHandler]
          *   registered → `ACTION_VIEW` into the host, falling back to the launcher if unresolvable.
-         * - Handler registered, or no deep link → launcher only. `handlePush` already dispatched the
-         *   handler, so an additional `ACTION_VIEW` would double-deliver navigation.
+         * - Deep link present and a handler registered → launcher intent flagged only to bring the
+         *   host's task to the front. `handlePush` already dispatched the handler, so an additional
+         *   `ACTION_VIEW` would double-deliver navigation, and clearing the back stack would undo it.
+         * - No deep link (`open_app`) → launcher intent with the same flags as the non-trampoline
+         *   content intent it replaces.
          */
         private fun startDestination(intent: Intent, context: Context) {
             val deepLink = intent.data
+            // Read once so the destination and its flags can't be decided from different state.
+            // `handlePush` above dispatches this link to the host's handler synchronously on the
+            // main thread, so by now the host may already have navigated. (A repeat tap of the
+            // same delivery is deduped inside `handlePush` and navigates nothing — still the right
+            // branch, since there is no reason to tear down what the first tap navigated to.)
+            val handlerOwnsNavigation = deepLink != null && DeepLinking.isHandlerRegistered
+
             val destination: Intent? = when {
-                deepLink != null && !DeepLinking.isHandlerRegistered -> {
+                deepLink != null && !handlerOwnsNavigation -> {
                     val viewIntent = DeepLinking.makeDeepLinkIntent(
                         deepLink,
                         context,
@@ -150,9 +160,9 @@ internal class KlaviyoTrampolineActivity : Activity() {
                     }
                 }
                 else -> {
-                    if (deepLink != null) {
+                    if (handlerOwnsNavigation) {
                         Registry.log.verbose(
-                            "Trampoline dispatching deep link via handler; launching host"
+                            "Trampoline dispatched deep link via handler; bringing host to front"
                         )
                     } else {
                         Registry.log.verbose("Trampoline dispatching launch intent")
@@ -162,15 +172,27 @@ internal class KlaviyoTrampolineActivity : Activity() {
             }
 
             destination?.apply {
-                // CLEAR_TOP mirrors the non-trampoline path (KlaviyoNotification adds it to the
-                // contentIntent, but it's consumed launching the trampoline rather than forwarded),
-                // preserving back-stack behavior. NEW_TASK is required here since we launch from the
-                // trampoline's own (taskAffinity="") task.
-                addFlags(
-                    Intent.FLAG_ACTIVITY_NEW_TASK or
-                        Intent.FLAG_ACTIVITY_SINGLE_TOP or
-                        Intent.FLAG_ACTIVITY_CLEAR_TOP
-                )
+                if (handlerOwnsNavigation) {
+                    // Launcher-icon semantics: resume the host's task exactly as the user left it.
+                    // The handler owns navigation, so CLEAR_TOP would finish whatever it just
+                    // navigated to, and SINGLE_TOP would re-deliver an ACTION_MAIN intent to the
+                    // launcher activity, resetting in-activity (Fragment/Compose) routing.
+                    //
+                    // Assigned rather than added because makeLaunchIntent bakes in SINGLE_TOP (see
+                    // DeepLinkingTest) and Intent.removeFlags requires API 26, above our minSdk.
+                    // getLaunchIntentForPackage sets exactly NEW_TASK, so nothing else is dropped.
+                    flags = Intent.FLAG_ACTIVITY_NEW_TASK
+                } else {
+                    // CLEAR_TOP mirrors the non-trampoline path (KlaviyoNotification adds it to the
+                    // contentIntent, but it's consumed launching the trampoline rather than
+                    // forwarded), preserving back-stack behavior. NEW_TASK is required here since we
+                    // launch from the trampoline's own (taskAffinity="") task.
+                    addFlags(
+                        Intent.FLAG_ACTIVITY_NEW_TASK or
+                            Intent.FLAG_ACTIVITY_SINGLE_TOP or
+                            Intent.FLAG_ACTIVITY_CLEAR_TOP
+                    )
+                }
             }?.startActivityIfResolved(context)
                 ?: Registry.log.warning("No launch intent found for host app; nothing to start")
         }
--- a/README.md
+++ b/README.md
@@ -1149,6 +1149,13 @@ You should register this from your `Application` or main `Activity`'s `.onCreate
 the application lifecycle to handle any link that launches the app from a terminated state. This handler will be invoked for
 *any* deep link originating from the Klaviyo SDK, including push notifications, universal tracking links, or In-App Forms.
 
+> **Push notification taps with automatic open tracking:** When `automatic_push_open_tracking` is enabled and you have
+> registered a handler, the SDK invokes your handler and then brings your app's existing task to the foreground **without**
+> clearing its back stack or delivering an intent to your launcher `Activity`. Whatever you navigate to from the handler —
+> another `Activity`, a `Fragment` transaction, a Compose `NavController` route — stays on top. Because no intent is delivered
+> on this path, do your routing in the handler callback rather than in `onNewIntent`. If no handler is registered, the SDK
+> instead sends your app an `ACTION_VIEW` intent carrying the link, which you should handle in `onCreate`/`onNewIntent`.
+
 <details open>
    <summary>Kotlin</summary>
 
--- a/sample/src/main/AndroidManifest.xml
+++ b/sample/src/main/AndroidManifest.xml
@@ -53,6 +53,16 @@
             </intent-filter>
         </activity>
 
+        <!--
+            Sample App Only: the destination the deep link handler navigates to. Declared with the
+            default task affinity so it stacks on top of SampleActivity in the same task, which is
+            what makes the back-stack behavior after a notification tap observable.
+        -->
+        <activity
+                android:name=".SampleDetailActivity"
+                android:exported="false"
+                android:theme="@style/Theme.KlaviyoAndroidSdk"/>
+
         <!-- Optional: Set the Klaviyo SDK logging level (1=verbose, 2=debug etc) -->
         <meta-data
                 android:name="com.klaviyo.core.log_level"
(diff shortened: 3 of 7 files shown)


glenn-klaviyo (approved):
Did some local testing on this branch around Push Notifications, Deep Links, Open Web URL project and all seemed to open / navigate just fine. 

just looping in @morrilltim since we're touching the `KlaviyoTrampolineActivity`  in this PR, and so he's aware. i don't believe this changes anything with open web url as i tested things locally but wanted to keep you posted.

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoTrampolineActivity.kt
```diff
@@ -162,15 +172,27 @@ internal class KlaviyoTrampolineActivity : Activity() {
...
-                addFlags(
-                    Intent.FLAG_ACTIVITY_NEW_TASK or
-                        Intent.FLAG_ACTIVITY_SINGLE_TOP or
-                        Intent.FLAG_ACTIVITY_CLEAR_TOP
-                )
+                if (handlerOwnsNavigation) {
+                    // Launcher-icon semantics: resume the host's task exactly as the user left it.
+                    // The handler owns navigation, so CLEAR_TOP would finish whatever it just
+                    // navigated to, and SINGLE_TOP would re-deliver an ACTION_MAIN intent to the
+                    // launcher activity, resetting in-activity (Fragment/Compose) routing.
```
evan-masseau, line 179: I need to re-check, but pretty sure this is exactly how stock FCM notifications behave, which is why we have the `CLEAR_TOP` flag elsewhere, and so this behavior is intentional.

evan-masseau:
> **Correction (edited in place):** an earlier version of this comment claimed the bare `NEW_TASK` flag buries the destination under a duplicate launcher instance, and proposed a cold/warm split on that basis. **That was wrong — it was an artifact of my test harness, not real behavior.** I had been launching the app with `am start -n <pkg>/.SampleActivity`, which omits `CATEGORY_LAUNCHER` and so never matches the task root; each invocation pushed a fresh `SampleActivity`, and I then attributed those instances to the trampoline. Re-measured properly, **your flag choice and its comment are correct.** The corrected findings are below. Sorry for the noise.

I spent a while on this one with Evan, including on-device measurement. Longer writeup covering both PRs is on #530: https://github.com/klaviyo/klaviyo-android-sdk/pull/530#issuecomment-5197247941

## The `handlerOwnsNavigation` carve-out is right, and so is the reasoning

Re-measured on a clean install, baselining the stack immediately before each tap (API 36 emulator, real notifications posted through `KlaviyoNotification.displayNotification` and tapped via UI, so the genuine PendingIntent path runs). Bare `NEW_TASK` resumes the host's task exactly as the user left it in all three states — no duplicate launcher instance anywhere:

| State | Before tap | After tap |
|---|---|---|
| Foreground | `[SampleActivity]` | `[SampleActivity, SampleDetailActivity]`, same task id |
| Backgrounded (process alive) | `[SampleActivity, SampleDetailActivity]` | resumed as-left, handler navigated on top, same task id |
| Cold (process killed) | — | task created, handler navigated on top |

So "Launcher-icon semantics: resume the host's task exactly as the user left it" is accurate as written.

One caveat worth knowing for future test-writing: this behavior **inverts in a same-class stack**, and it is also sensitive to how the host's handler navigates. A handler registered in `Application.onCreate` that calls `startActivity` from the *application* context must pass `NEW_TASK`, which puts the destination in a **separate task** and makes any stack measurement meaningless. That is what the sample app did, and it confounded my first round of testing. Worth a note in the sample so the next person measuring this doesn't repeat it.

## One thing worth adding here

Whichever way the dispatch timing lands, there is a non-obvious invariant in this file that deserves a comment and a test: **the deep link dispatch must be registered from `onCreate`, not `onResume`.**

The app-level `onActivityResumed` callback fires *before* an activity's own `onResume` body, so from `onResume` the trampoline itself would be `currentActivity`, and a postponed dispatch would run immediately against a dying translucent activity in a separate task. What makes `onCreate` correct is precisely the `finish()` in your `finally`: finishing during `onCreate` skips `onStart` and `onResume` entirely, so the trampoline never broadcasts a `Resumed` of its own and the first one observed is guaranteed to belong to the host. Verified on device — the trampoline never reaches either callback.

That is easy to break in a later refactor without any test noticing.

## Measured timings, for reference

The postponed dispatch fires against the host's activity ~70ms after a warm tap and ~112ms after a cold start. Useful for sizing any timeout: 10s is far more headroom than needed.

-- Claude

evan-masseau:
Cross-linking: I've put up an alternative approach as #533 → #534 (stacked), for comparison rather than replacement — Evan wants to talk both options over with @aahiltn before deciding, so **nothing here is being closed**.

**The difference in one line:** instead of deferring the `DeepLinkHandler` until an activity resumes, the trampoline sends the host an `ACTION_VIEW` intent carrying the link and doesn't invoke the handler at all under `automatic_push_open_tracking`.

That removes the postponed lifecycle observer, the timeout, the drop-on-timeout policy, the crash containment inside `onActivityResumed`, and the "register from `onCreate`, never `onResume`" invariant — Android does the routing instead.

**The fact that convinced me it isn't a capability reduction:** in released 4.4.x, `makeResolvedDeepLinkIntent` discards an unresolvable deep link and substitutes the launcher intent, whose `data` is null. `handlePush` reads `intent?.data`, so on that path the handler was **never invoked at all**. A registered handler in 4.4.x never receives a link an intent couldn't have delivered. The only apps that would lose anything are ones relying on the carve-out `rel/4.5.0` added under auto-tracking — which is unreleased.

Two things from this branch are carried forward:
- The `handlePush(intent, dispatchDeepLink)` overload, in #533, with de-duplication split per stage so a host that still calls `handlePush` reaches their handler while `$opened_push` stays exactly-once.
- The `SampleDetailActivity` scaffolding, which is what makes back-stack behavior observable on device.

Also worth carrying over regardless of which approach wins: **`onNewIntent` handling is mandatory, not optional.** Measured — when the process is killed with its task still in recents, Android restores the *old* intent into `onCreate` and delivers the new one to `onNewIntent`. Platform behavior, reproduces without any deep link, but the intent path leans on it much harder.

-- Claude

evan-masseau:
Closing — the team has landed on the intent-based approach for 4.5.0, implemented in #533 → #534. Not because this PR was wrong.

**Your flag choice and its comment were correct.** I want to be unambiguous about that, because an earlier comment of mine on this PR claimed otherwise and it was wrong. I said bare `NEW_TASK` buries the destination under a duplicate launcher instance, and that "launcher-icon semantics: resume the host's task exactly as the user left it" was unsupported.

That was a test-harness artifact. I was launching the app with `am start -n <pkg>/.SampleActivity`, which omits `CATEGORY_LAUNCHER` and therefore never matches the task's root intent — so every invocation pushed a *fresh* `SampleActivity`. I ran it repeatedly across tests and then attributed the instances I had created to the trampoline's launcher intent.

Re-measured on a clean install, baselining the stack immediately before each tap and launching with `-a android.intent.action.MAIN -c android.intent.category.LAUNCHER`, bare `NEW_TASK` resumes the host's task exactly as the user left it — in all three app states: foreground, backgrounded, and cold. No duplicate instance in any case. My comment on this PR has been corrected in place, and the same retraction is on MAGE-1039.

**Why the direction changed anyway:** the carve-out here is coupled to the deferral in #530 — it is correct *given* that the handler owns navigation. Once you stop invoking the handler for a push tap and let an `ACTION_VIEW` intent carry the link instead, the branch is unnecessary and the flags go back to `NEW_TASK|SINGLE_TOP|CLEAR_TOP`, matching what 4.4.x sent.

The decisive fact for that direction: in released 4.4.x, `makeResolvedDeepLinkIntent` discards an unresolvable deep link and substitutes the launcher intent, whose `data` is null — and `handlePush` reads `intent.data`. So on that path the handler was never invoked at all. A registered handler in 4.4.x never receives a link an intent could not have delivered, which is what makes intent-only delivery not a capability reduction.

`SampleDetailActivity` from this branch carried forward into #534 — it is what made all of the on-device back-stack measurement possible.

-- Claude

PR #531, opened 2026-08-06
fix(core): settle postponed-activity waits atomically

evan-masseau:
# Description
Fixes a race in `LifecycleMonitor.runWithCurrentOrNextActivity` that could let two outcomes run for a single wait, corrects `runNow` so it actually runs the job, and adds an `onTimeout` overload for callers that must not silently drop deferred work.

Extracted from #530 so it can be reviewed and land on its own. It is independent of the deep-link dispatch work in #533 → #534 — those branches do not touch this code — and it benefits the three callers that already exist on `rel/4.5.0`. Original commits by @aahiltn, preserved via co-authorship.

## Due Diligence

`Minor` because it adds a public overload to `LifecycleMonitor`. The existing two-argument signature is deliberately untouched — see below.

### 1. The race

`runWithCurrentOrNextActivity` had three possible outcomes — job runs, timeout expires, caller cancels — with nothing arbitrating between them:

- The timeout fires on the clock's own thread while activity resumes broadcast on the main thread, so a resume racing the timer could invoke the job while the observer was already being torn down.
- The returned token didn't abandon a pending wait. `cancel()` cancelled the timeout but left the observer registered, so a later resume still ran the job the caller had given up on.

An `AtomicBoolean` now claims the outcome, so exactly one of job / timeout / cancellation wins.

**This is reachable today, and one instance is user-visible.** `SystemClock.schedule` uses a real `java.util.Timer` thread, and `KlaviyoPresentationManager.dismiss()` → `clearTimers()` can run on that thread while a postponed present fires on the main thread. Concretely, the `cancel()` bug meant dismissing an in-app form that was waiting to re-present did **not** abandon that wait — the observer stayed registered, so a later unrelated activity resume could re-present a form the user had already dismissed.

### 2. `runNow` now runs the job

`runNow()` aliased `cancel()`, so a caller asking to run the job immediately got nothing at all. That is inconsistent with both other `Clock.Cancellable` implementations — `SystemClock` and `StaticClock` each invoke their task — and the original justification ("there is no way to run now a job waiting precisely because it has no activity") was wrong: by the time a caller reaches for `runNow`, there often *is* a current activity.

It now invokes the job against `currentActivity`, falling back to `cancel()` when there is none, so the token always resolves to something rather than silently no-oping.

**This exposed a latent bug in forms.** `KlaviyoPresentationManager.clearTimers()` called `runNow()` *intending it to cancel* — its comment read "Run this cancel job now to stop any postponed form presentation." With `runNow` fixed, that would have presented the form the user just dismissed. It now calls `cancel()` directly, which is what it always meant. Covered by a new regression test in `KlaviyoPresentationManagerTest`.

### 3. Observer registration precedes its timeout

The timeout task captured `observer` before it was assigned, so a timeout elapsing during setup removed nothing and left the observer registered on the process-scoped monitor permanently. Registering first means the timeout always has a real reference to remove. Reported by CodeRabbit.

No regression test: the leaked observer never invokes the job (`settled` is already true, so dispatch short-circuits), and intercepting `onActivityEvent`/`offActivityEvent` does not work because `runWithCurrentOrNextActivity` is a default method on the interface — its internal calls dispatch on `this`, which neither `spyk` nor `mockkObject` intercepts. Asserting it would require exposing `activityObservers` or restructuring the function, which is more invasive than the two lines it guards.

### 4. What was deliberately **not** kept

An earlier revision reconciled a resume landing in the window between the initial `currentActivity` check and the observer's registration. That window requires a caller running off the main thread *and* a timeout short enough to race its own scheduling. Judged too remote to carry the complexity; the KDoc now warns callers against timeouts near `ACTIVITY_TRANSITION_GRACE_PERIOD` instead. Two tests that only covered that block were removed with it.

### 5. Why an overload rather than a defaulted parameter

Inserting `onTimeout` into the existing signature would change that function's JVM descriptor and break binary compatibility for already-compiled callers — `LifecycleMonitor` is public and reachable via `Registry.lifecycleMonitor`. Verified by compiling `:sdk:core` on both `origin/rel/4.5.0` and this branch and diffing `javap`: the 2-arg overload's descriptor and its `$default` synthetic bridge in `LifecycleMonitor$DefaultImpls` are byte-for-byte identical. `-Xjvm-default` is not set anywhere in the repo, so interface defaults compile to the `DefaultImpls` pattern already used by `PushTokenFetcher`. `onTimeout` has no default value, so a two-argument call site resolves only to the two-argument overload.

That choice is load-bearing for review: `BaseTest.kt` is **unchanged** in this PR. In #530, where the parameter was inserted mid-signature, it needed edits to add an extra `any()` matcher. Its absence here is the evidence that existing call sites still compile untouched.

**Existing callers** (all pre-date this change): `DeepLinking.sendDeepLinkIntent`, and two in `KlaviyoPresentationManager`. None adopt `onTimeout` yet — it is plumbing for future use.

## Test Plan

`KlaviyoLifecycleMonitorTest` covers resume-vs-timeout arbitration, cancellation abandoning a pending wait, `runNow` invoking the job against the current activity, and `runNow` abandoning the wait when there is none. `KlaviyoPresentationManagerTest` adds the dismiss-doesn't-present regression.

**Being precise about what the tests prove.** `StaticClock` is fully synchronous, so no test exercises a real cross-thread interleaving — the arbitration is verified by reasoning about the `AtomicBoolean` gate, not by a concurrency test. Reviewers should not read the suite as proof the race itself is covered.

Both behavioral fixes were mutation-checked: reverting `clearTimers` to `runNow()` fails the new forms test, and the `runNow` tests fail against the old alias.

## Related Issues/Tickets
Extracted from #530 (closed). See https://github.com/klaviyo/klaviyo-android-sdk/pull/530#issuecomment-5197247941 for the full investigation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved postponed activity actions so they execute only once when an activity resumes.
  - Added timeout handling for actions that cannot run within the configured period.
  - Cancelled actions no longer execute later or trigger timeout callbacks.
  - Dismissing a pending form now properly cancels its scheduled presentation instead of displaying it unexpectedly.
  - Added support for manually triggering eligible postponed actions immediately.

### DIFF
--- a/sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
@@ -138,18 +138,14 @@ interface LifecycleMonitor {
      * next resumed activity, falling back to [onTimeout] if none resumes within [timeout].
      *
      * Exactly one of [job] and [onTimeout] is invoked, unless the returned token cancels the wait
-     * first.
-     *
-     * [onTimeout] has no default value, so a call supplying only a timeout and a trailing lambda
-     * resolves to the two-argument overload.
+     * first. The timeout should be at least long enough to avoid race conditions between the
+     * scheduling of the tasks, e.g. min [ACTIVITY_TRANSITION_GRACE_PERIOD] milliseconds.
      *
      * @param timeout How long to wait for a resumed activity, or null to wait indefinitely
      * @param onTimeout Invoked instead of [job] if [timeout] elapses with no resumed activity
      * @param job Invoked with the current activity, or the next one to resume
-     * @return null if the wait already settled before returning, otherwise a token that abandons
-     *  the pending wait. Both [Clock.Cancellable.cancel] and [Clock.Cancellable.runNow] abandon it
-     *  — there is no way to "run now" a job that is waiting precisely because it has no activity to
-     *  run against — and neither runs [job] or [onTimeout].
+     * @return A token to cancel the pending wait or attempt to run immediately against the
+     * current activity and abandon the wait, or null if [job] ran immediately.
      */
     fun runWithCurrentOrNextActivity(
         timeout: Long?,
@@ -161,11 +157,11 @@ interface LifecycleMonitor {
             return null
         }
 
-        // The timeout fires on the clock's own thread, so claim the outcome atomically: a resume
-        // racing the timer — or a caller abandoning the wait — must not let two branches run.
+        // Track atomically whether the task or timeout have run
         val settled = AtomicBoolean(false)
         var observer: ActivityObserver? = null
 
+        // Cancel the task if a specified timeout elapses, and invoke the optional timeout callback
         val timeoutTask: Clock.Cancellable? = timeout?.let { delay ->
             Registry.clock.schedule(delay) {
                 if (!settled.compareAndSet(false, true)) return@schedule
@@ -175,41 +171,31 @@ interface LifecycleMonitor {
             }
         }
 
+        // Invoke the job when the next activity resumes, and cancel the timeout task
+        val waitingTask: (activity: Activity, reason: String) -> Unit = { activity, reason ->
+            if (settled.compareAndSet(false, true)) {
+                Registry.log.verbose("Invoking postponed observer on $reason")
+                observer?.let { offActivityEvent(it) }
+                timeoutTask?.cancel()
+                job(activity)
+            } else {
+                Registry.log.verbose("Postponed observer already settled, ignoring $reason")
+            }
+        }
+
         observer = { event ->
             event.takeIf<ActivityEvent.Resumed>()?.let { resumed ->
-                if (settled.compareAndSet(false, true)) {
-                    Registry.log.verbose("Invoking postponed observer on resume")
-                    observer?.let { offActivityEvent(it) }
-                    timeoutTask?.cancel()
-                    job(resumed.activity)
-                }
+                waitingTask(resumed.activity, "resume")
             }
         }
 
         onActivityEvent(observer)
 
-        // Reconcile with whatever raced us while we were registering. The timeout runs on the
-        // clock's own thread, and callers can be off the main thread while activity resumes
-        // broadcast on it.
-        if (settled.get()) {
-            // Something claimed the outcome while we registered — a timeout that de-registered an
-            // observer that did not exist yet, or a resume that already ran the job. Either way the
-            // wait is over, so there is nothing left for a token to abandon.
-            offActivityEvent(observer)
-            return null
-        }
-
-        currentActivity?.let { activity ->
-            if (settled.compareAndSet(false, true)) {
-                Registry.log.verbose("Activity resumed while registering postponed observer")
-                offActivityEvent(observer)
-                timeoutTask?.cancel()
-                job(activity)
-                return null
-            }
-        }
-
         return object : Clock.Cancellable {
+            /**
+             * Cancel the wait for the next resumed activity, and cancel the timeout task if any.
+             * [onTimeout] is not invoked automatically, caller can do that manually if needed.
+             */
             override fun cancel(): Boolean {
                 if (!settled.compareAndSet(false, true)) return false
                 Registry.log.verbose("Abandoning postponed observer")
@@ -218,8 +204,16 @@ interface LifecycleMonitor {
                 return true
             }
 
+            /**
+             * Contract correctness: an option to attempt to invoke immediately. Not recommended.
+             */
             override fun runNow() {
-                cancel()
+                currentActivity?.let { activity ->
+                    waitingTask(activity, "runNow")
+                } ?: run {
+                    Registry.log.verbose("Postponed observer has no current activity, abandoning")
+                    cancel()
+                }
             }
         }
     }
--- a/sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
@@ -145,7 +145,8 @@ interface LifecycleMonitor {
      * @param onTimeout Invoked instead of [job] if [timeout] elapses with no resumed activity
      * @param job Invoked with the current activity, or the next one to resume
      * @return A token to cancel the pending wait or attempt to run immediately against the
-     * current activity and abandon the wait, or null if [job] ran immediately.
+     * current activity and abandon the wait, or null if the wait already settled — [job] ran
+     * against the current activity, or a resume or the timeout claimed it during registration.
      */
     fun runWithCurrentOrNextActivity(
         timeout: Long?,
@@ -190,6 +191,15 @@ interface LifecycleMonitor {
                 offActivityEvent(observer)
                 onTimeout?.invoke()
             }
+        }
+
+        // Activity resumes broadcast on the main thread and the timeout runs on the clock's own
+        // thread, so either can settle the wait before this returns. A settled wait has nothing
+        // left to abandon, and a resume that beat the assignment above left the timeout scheduled.
+        if (settled.get()) {
+            offActivityEvent(observer)
+            timeoutTask?.cancel()
+            return null
         }
 
         return object : Clock.Cancellable {
--- a/sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
+++ b/sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
@@ -160,16 +160,7 @@ interface LifecycleMonitor {
         // Track atomically whether the task or timeout have run
         val settled = AtomicBoolean(false)
         var observer: ActivityObserver? = null
-
-        // Cancel the task if a specified timeout elapses, and invoke the optional timeout callback
-        val timeoutTask: Clock.Cancellable? = timeout?.let { delay ->
-            Registry.clock.schedule(delay) {
-                if (!settled.compareAndSet(false, true)) return@schedule
-                Registry.log.verbose("Removing postponed observer after timeout ${delay}ms")
-                observer?.let { offActivityEvent(it) }
-                onTimeout?.invoke()
-            }
-        }
+        var timeoutTask: Clock.Cancellable? = null
 
         // Invoke the job when the next activity resumes, and cancel the timeout task
         val waitingTask: (activity: Activity, reason: String) -> Unit = { activity, reason ->
@@ -190,6 +181,16 @@ interface LifecycleMonitor {
         }
 
         onActivityEvent(observer)
+
+        // Cancel the task if a specified timeout elapses, and invoke the optional timeout callback
+        timeoutTask = timeout?.let { delay ->
+            Registry.clock.schedule(delay) {
+                if (!settled.compareAndSet(false, true)) return@schedule
+                Registry.log.verbose("Removing postponed observer after timeout ${delay}ms")
+                offActivityEvent(observer)
+                onTimeout?.invoke()
+            }
+        }
 
         return object : Clock.Cancellable {
             /**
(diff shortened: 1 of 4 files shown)


sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
```diff
@@ -118,35 +119,113 @@ interface LifecycleMonitor {
...
+                    Registry.log.verbose("Activity resumed while registering postponed observer")
+                    offActivityEvent(observer)
+                    timeoutTask?.cancel()
+                    job(activity)
+                    return null
+                }
+            }
+        }
+
+        return object : Clock.Cancellable {
```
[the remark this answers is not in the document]
evan-masseau: Confirmed and fixed in 189077e4.

You're right that the settled branch fell through to the token return. Worth noting the reachable case is broader than the timeout: `settled` can also be claimed by the observer firing on the main thread while an off-main caller is still inside `onActivityEvent` — so the job could have already run, which contradicts the return contract directly rather than just being redundant.

The settled branch now returns `null`, and the KDoc says "null if the wait already settled before returning" rather than singling out the job-ran case. Added a regression test that drives the timeout during registration; it fails if the early return is removed.

Practical impact was low — the returned token's `cancel()` would `compareAndSet` against an already-true flag and report `false` — but the inconsistency was real and this is the function the PR rewrites, so it belongs here.

-- Claude

evan-masseau:
bugbot run

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoPresentationManager.kt
```diff
@@ -436,8 +436,8 @@ internal class KlaviyoPresentationManager() : PresentationManager {
      * Clear timers and observers to stop any delayed side effects
      */
     private fun clearTimers() {
-        // Run this cancel job now to stop any postponed form presentation
-        cancelPostponedPresent?.runNow().also { cancelPostponedPresent = null }
+        // Abandon any postponed form presentation. Not runNow(), which would present the form.
```
evan-masseau, line 439: The semantics of the `Cancellable` were incorrectly applied before this PR. `cancel` should cancel the job you were trying to schedule, and `runNow` should invoke it immediately. Previously we were returning the *timeout's* token, so `runNow` meant cancel the task

sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
```diff
@@ -118,35 +119,103 @@ interface LifecycleMonitor {
...
+            }
+        }
+
+        observer = { event ->
+            event.takeIf<ActivityEvent.Resumed>()?.let { resumed ->
+                waitingTask(resumed.activity, "resume")
             }
         }
+
         onActivityEvent(observer)
```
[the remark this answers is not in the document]
evan-masseau: This is a virtual impossibility IRL, and I added a warning to the KDoc about using a sane timeout value in the argument.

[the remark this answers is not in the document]
evan-masseau: Push back on me if you think this is more likely scenario IRL than I anticipate, to me that sounds like a microsecond race condition unlikely to happen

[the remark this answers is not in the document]
evan-masseau: If I did flip it, and register the observer first, the timeout second, would that eliminate the concern? Would it not just flip it, and leave open the possibility of registering a timeout wait that could be registered when the activity observer has already run?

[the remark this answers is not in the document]
evan-masseau: Good catch — fixed in e8e38cb2 by registering the observer before scheduling the timeout, so the timeout always has a real reference to remove. `observer` is definitely-assigned by that point, so the nullable `?.let` inside the timeout collapses to a direct call.

Worth separating this from a nearby race that was **deliberately** dropped in this PR: an earlier revision also reconciled a resume landing between the initial `currentActivity` check and the registration. That one needs a caller off the main thread *and* is only a dropped link, not a leak — judged too remote to carry the complexity, and the KDoc now warns against timeouts short enough to race their own scheduling. This finding is different: it strands an observer on the process-scoped monitor permanently, and the fix is free, so it is worth doing regardless of how likely it is.

Two corrections to the report itself:

**Not a heavy lift.** It is a two-line move — no new state, no reconciliation block, no serialization.

**No regression test, and I want to be straight about why rather than quietly skipping it.** I tried three approaches and all three failed to detect the mutation:
- The leaked observer never invokes the job (`settled` is already true, so the dispatch short-circuits), so no job-invocation assertion can see it. Verified by probe: `calledAfterResume=false` both with and without the fix.
- Asserting on observer registration/removal requires intercepting `onActivityEvent`/`offActivityEvent`, but `runWithCurrentOrNextActivity` is a **default method on the interface** — its internal calls dispatch on `this`, not on a `spyk` or `mockkObject` wrapper, so neither sees them.

Making this observable would mean exposing `activityObservers` for testing or restructuring the function, which is more invasive than the two lines it would guard. If you see a way to assert it without that, I am interested.

-- Claude

sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
```diff
@@ -118,35 +119,104 @@ interface LifecycleMonitor {
...
+        timeoutTask = timeout?.let { delay ->
+            Registry.clock.schedule(delay) {
+                if (!settled.compareAndSet(false, true)) return@schedule
+                Registry.log.verbose("Removing postponed observer after timeout ${delay}ms")
+                offActivityEvent(observer)
+                onTimeout?.invoke()
+            }
+        }
+
+        return object : Clock.Cancellable {
```
Atharva8168, line 206: Looks like the Bugbot fix from 189077e4 (`if (settled.get()) { offActivityEvent(observer); return null }` + its test) got deleted in 55c6578ad along with the legitimately-dropped reconciliation block — but this is a separate, already-acknowledged fix, and its removal isn't mentioned in the PR description. Currently a settled wait can still return a live token instead of `null`. Intentional, or swept up in the runNow refactor?

sdk/core/src/test/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitorTest.kt
```diff
@@ -212,4 +212,144 @@ class KlaviyoLifecycleMonitorTest : BaseTest() {
         KlaviyoLifecycleMonitor.onActivityResumed(mockk())
         assert(!called) { "Callback should not be called if timed out" }
     }
+
+    @Test
+    fun `runWithCurrentOrNextActivity invokes onTimeout if activity is not resumed in time`() {
```
Atharva8168, line 219: Nice coverage overall! One small gap I noticed: there's no test here for the settled-during-registration race I flagged on LifecycleMonitor.kt:195 — there actually was one (`returns no token when the wait settles while registering`, from 189077e4) before it got removed along with the fix in 55c6578ad. Might be worth bringing both back if you agree the fix should stay?

sdk/forms/src/main/java/com/klaviyo/forms/presentation/KlaviyoPresentationManager.kt
Atharva8168, line 439: This one's spot on — calling `cancel()` here is exactly what the old comment always meant, and catching that the `runNow()` fix would otherwise re-present a form the user just dismissed is a great catch. Nice, minimal fix.

Atharva8168 (approved):
Approving — Overall LGTM. Added some comments too.

sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
evan-masseau: Swept up, not intentional — good catch. Restored in `51c14a1219`.

It was collateral from the `runNow` refactor in `55c6578ad`: that commit deliberately dropped the *reconciliation* block (the `currentActivity?.let { ... }` re-check, which only ever dropped a job and needed an off-main caller), and the settled-check early return was sitting directly above it in the same hunk. The description called out the reconciliation removal and said nothing about this one, which is exactly why it read as accidental.

One thing that changed while it was gone, and made the restored version slightly different from `189077e4`: `e8e38cb2` moved observer registration *before* the timeout scheduling. So the reachable settler here is now a **resume**, not the timeout — and a resume landing in that window runs `waitingTask` before `timeoutTask` has been assigned, so its `timeoutTask?.cancel()` is a no-op against `null`. The wait settles with its timer still scheduled. The restored branch therefore cancels the timeout too, which the original didn't need to:

```kotlin
if (settled.get()) {
    offActivityEvent(observer)
    timeoutTask?.cancel()
    return null
}
```

Regression test rewritten for the same reason — the old one drove `staticClock.execute` from inside `onActivityEvent`, which no longer settles anything now that the timeout is scheduled afterward. It fires a resume during registration instead, and asserts the leftover timer is gone via `staticClock.scheduledTasks`.

Mutation-verified: removing the branch fails exactly one test.

-- Claude

sdk/core/src/test/java/com/klaviyo/core/lifecycle/KlaviyoLifecycleMonitorTest.kt
evan-masseau: Both are back as of `51c14a1219` — details on the sibling thread.

The test isn't a straight revert though. `e8e38cb2` reordered registration ahead of the timeout scheduling, so the original version's mechanism (driving `staticClock.execute(150)` from inside `onActivityEvent`) can no longer settle anything at that point — the timeout isn't scheduled yet, so there's nothing for the clock to fire. It would have passed vacuously.

Rewritten to fire a resume during registration instead, which is the window that's actually reachable now, plus an assertion that no orphaned timer is left scheduled.

-- Claude

sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
```diff
@@ -118,35 +119,114 @@ interface LifecycleMonitor {
...
      * Helper function to run a task immediately if there is a current activity,
      * or wait for the next resumed activity if resumed within the optional timeout.
-     * Returns a token that can be used to cancel the pending task if needed.
+     *
+     * A job that outlives its timeout is dropped. Callers that need a fallback should use the
+     * [onTimeout] overload rather than relying on the resume arriving in time.
+     *
+     * @param timeout How long to wait for a resumed activity, or null to wait indefinitely
+     * @param job Invoked with the current activity, or the next one to resume
+     * @return null if [job] already ran against the current activity, otherwise a token that
```
Atharva8168: Surfacing this inline since CodeRabbit only raised it in the collapsed "outside diff range" section of its last review, where there's no thread to track it against.

This `@return` looks stale as of `51c14a121`: the 3-arg overload can now also return `null` when a resume or the timeout settles the wait during registration, and this overload delegates straight to it — so "null if [job] already ran against the current activity" no longer covers every null case. Might be worth mirroring the wording you already added to the 3-arg overload's `@return`?

sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
evan-masseau: Meh, that isn't really true, I did intentionally drop it, it seemed like an insane level of edge case that only an AI would consider. But it doesn't hurt to have it I guess.

sdk/core/src/main/java/com/klaviyo/core/lifecycle/LifecycleMonitor.kt
```diff
@@ -118,35 +119,114 @@ interface LifecycleMonitor {
...
+     *  abandons the pending wait.
```
evan-masseau:
```suggestion
     * @return A token to cancel the pending wait or attempt to run immediately against the
     * current activity and abandon the wait, or null if the wait already settled — [job] ran
     * against the current activity, or a resume or the timeout claimed it during registration.
```

PR #532, opened 2026-08-06
fix(push): dispatch trampoline deep links once the host has resumed

evan-masseau:
# Description
A notification tap that cold-starts the app reached a registered `DeepLinkHandler` before any `Activity` existed, so a handler routing through a `NavController`, a fragment transaction, or a held `Activity` had nothing to navigate and the link was lost. The trampoline now postpones the handler invocation until the host has a resumed `Activity`.

**Stacked on #531** — review that one first. This PR's diff is against `ecm/lifecycle-monitor-postponed-activity`.

Supersedes #529 and #530, which diagnose the same bug correctly. The difference is scope: those put the wait inside `DeepLinking.handleDeepLink`, the shared funnel for all three deep-link origins, which retimes form CTAs, universal links, and manual `Klaviyo.handlePush` calls. Original commits from those branches are preserved here via co-authorship.

## Due Diligence

`Minor`: adds a `handlePush` overload. It is `@RestrictTo(LIBRARY_GROUP)` + `@JvmSynthetic`, so it is lint-blocked externally and invisible from Java, but it is still a Kotlin-visible symbol.

## Changelog / Code Overview

**Scoping the retime.** `DeepLinking.handleDeepLink` is unchanged. Instead, `handlePush` gains an overload that runs tracking and dismissal but suppresses the final deep-link dispatch, so the trampoline can sequence dispatch itself:

```kotlin
@JvmStatic fun handlePush(intent: Intent?): Klaviyo            // unchanged
@RestrictTo(LIBRARY_GROUP) @JvmSynthetic
fun handlePush(intent: Intent?, dispatchDeepLink: Boolean): Boolean
```

The trampoline then registers the handler invocation against the next resumed `Activity`, and brings the host's task to the front with `FLAG_ACTIVITY_NEW_TASK`.

**Ordering is load-bearing in two places:**

1. The postponed dispatch is registered *before* the host is started, so a resume cannot beat the registration.
2. Dispatch is registered from `onCreate`, not `onResume`. `Application.ActivityLifecycleCallbacks.onActivityResumed` fires *before* an `Activity`'s own `onResume` body, so from `onResume` the trampoline itself would already be the tracked current activity and the handler would run against a dying translucent activity in a separate task. What makes `onCreate` safe is the existing `finish()` in the `finally`: finishing during `onCreate` skips `onStart`/`onResume` entirely, so the trampoline never broadcasts a resume of its own.

**Dedup.** Moving dispatch out of `handlePush` also moved it out from behind that function's de-duplication guard, so `handlePush` now reports whether it handled the delivery and the trampoline only postpones a dispatch when it did. A duplicate still brings the host to the front — matching what a repeat tap did before — without re-running navigation.

**Timeout.** If no activity resumes within 2s the link is dropped with a warning rather than delivered. Invoking a handler that needs an activity when none resumed would lose the link anyway, and falling back to `ACTION_VIEW` would send an intent a handler-registered host likely never wired up. Measured worst case on-device was ~112ms.

**Also fixed:** `sendDeepLinkIntent` waited up to 50ms for a `currentActivity` it used only for `packageName` and `packageManager`. It now uses the resumed activity when there is one and the application context with `NEW_TASK` otherwise, so it never waits.

## Behavior changes, stated precisely

`handleDeepLink` itself is untouched, but `sendDeepLinkIntent` is on its no-handler path, so:

| Flow | Changed? |
|---|---|
| Push open (handler registered) | Retimed — this is the fix |
| Push open (no handler) | No. `KlaviyoPushOpenHandler` only calls `handleDeepLink` when a handler is registered, so this never reaches `sendDeepLinkIntent` |
| Manual `Klaviyo.handlePush` | No |
| Form CTA / universal link (handler registered) | No |
| Form CTA / universal link (no handler) | Yes — no longer waits up to 50ms. Identical when an activity is resumed; previously the link could be dropped when one was not |

No `MIGRATION_GUIDE` entry for callback retiming, because no released callback path changes timing.

## Test Plan

Verified on an API 36 emulator (`automaticDebug`), posting real notifications through `KlaviyoNotification.displayNotification` and tapping them via UI so the genuine PendingIntent path runs, with the stack baselined immediately before each tap:

| State | Result |
|---|---|
| Foreground | `[SampleActivity]` → `[SampleActivity, SampleDetailActivity]`, same task, dispatch ~70ms |
| Backgrounded (process alive) | task resumed as-left, handler navigated on top, same task |
| Cold (process killed) | task created, handler navigated on top, dispatch ~112ms |

Two tests were mutation-checked to confirm they fail when the behavior they guard is removed: the crash-containment `safeApply` guard, and the duplicate-delivery guard.

**Known test gap:** the `onCreate`-not-`onResume` invariant is verified on-device and documented in the source, but not covered by a unit test — every test calls the static `handleTrampolineIntent` helper directly, and asserting it properly would need Robolectric, which this repo does not currently use.

## Open question for reviewers

`@RestrictTo(LIBRARY_GROUP)` is the first use of that annotation in this SDK. Neighboring cross-module-internal functions (`DeepLinking.sendLaunchIntent`, `makeLaunchIntent`, `makeDeepLinkIntent`) solve the same problem with plain `public fun` + KDoc. Kotlin `internal` is not an option here — it is compilation-module scoped and `push-fcm` is a separate Gradle module. Happy to drop the annotation for consistency if the team prefers the established pattern.

## Related Issues/Tickets
Supersedes #529 and #530. See https://github.com/klaviyo/klaviyo-android-sdk/pull/530#issuecomment-5197247941 for the full investigation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1149,6 +1149,18 @@ You should register this from your `Application` or main `Activity`'s `.onCreate
 the application lifecycle to handle any link that launches the app from a terminated state. This handler will be invoked for
 *any* deep link originating from the Klaviyo SDK, including push notifications, universal tracking links, or In-App Forms.
 
+> **Push notification taps with automatic open tracking:** When `automatic_push_open_tracking` is enabled and you have
+> registered a handler, the SDK brings your app to the foreground if it is not already running, then invokes your handler
+> once your app has a resumed `Activity` — so it is safe to navigate from the callback even when the tap cold-starts your
+> app from a terminated state. Your back stack is never cleared, and the SDK does **not** send an `ACTION_VIEW` intent
+> carrying the link, so whatever you navigate to from the handler — another `Activity`, a `Fragment` transaction, a
+> Compose `NavController` route — stays on top. On a warm tap no new intent reaches your launcher `Activity` at all; on a
+> cold start it is created by the ordinary launch intent, which carries no deep link data. Either way the link reaches
+> you *only* through the handler, so do your routing in the callback rather than from `getIntent()` or `onNewIntent()`.
+> If your app fails to produce a resumed `Activity` within a couple of seconds, the link is dropped rather than delivered
+> to a handler that would have nothing to navigate. If no handler is registered, the SDK instead sends your app an
+> `ACTION_VIEW` intent carrying the link, which you should handle in `onCreate()`/`onNewIntent()`.
+
 <details open>
    <summary>Kotlin</summary>
 
--- a/sample/src/main/AndroidManifest.xml
+++ b/sample/src/main/AndroidManifest.xml
@@ -53,6 +53,16 @@
             </intent-filter>
         </activity>
 
+        <!--
+            Sample App Only: the destination the deep link handler navigates to. Declared with the
+            default task affinity so it stacks on top of SampleActivity in the same task, which is
+            what makes the back-stack behavior after a notification tap observable.
+        -->
+        <activity
+                android:name=".SampleDetailActivity"
+                android:exported="false"
+                android:theme="@style/Theme.KlaviyoAndroidSdk"/>
+
         <!-- Optional: Set the Klaviyo SDK logging level (1=verbose, 2=debug etc) -->
         <meta-data
                 android:name="com.klaviyo.core.log_level"
--- a/sample/src/main/java/com/klaviyo/sample/SampleApplication.kt
+++ b/sample/src/main/java/com/klaviyo/sample/SampleApplication.kt
@@ -33,6 +33,7 @@ class SampleApplication : Application() {
                 // OPTIONAL SETUP NOTE: Register a callback to handle any deep links from Klaviyo notifications, in-app forms, or universal tracking links
                 // If not using a deep link handler, Klaviyo will send an Intent to your app with the deep link in intent.data
                 showToast("Deep link to: $uri")
+                startActivity(SampleDetailActivity.intent(this, uri.toString()))
             }
             .registerFormLifecycleHandler { event ->
                 // OPTIONAL SETUP NOTE: Register a callback to receive form lifecycle events
--- a/sample/src/main/java/com/klaviyo/sample/SampleDetailActivity.kt
+++ b/sample/src/main/java/com/klaviyo/sample/SampleDetailActivity.kt
@@ -1,0 +1,79 @@
+package com.klaviyo.sample
+
+import android.content.Context
+import android.content.Intent
+import android.os.Bundle
+import androidx.activity.ComponentActivity
+import androidx.activity.compose.setContent
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.systemBarsPadding
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.OutlinedButton
+import androidx.compose.material3.Surface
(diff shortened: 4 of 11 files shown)


evan-masseau:
bugbot run

evan-masseau:
Cross-linking: I've put up an alternative approach as #533 → #534 (stacked), for comparison rather than replacement — Evan wants to talk both options over with @aahiltn before deciding, so **nothing here is being closed**.

**The difference in one line:** instead of deferring the `DeepLinkHandler` until an activity resumes, the trampoline sends the host an `ACTION_VIEW` intent carrying the link and doesn't invoke the handler at all under `automatic_push_open_tracking`.

That removes the postponed lifecycle observer, the timeout, the drop-on-timeout policy, the crash containment inside `onActivityResumed`, and the "register from `onCreate`, never `onResume`" invariant — Android does the routing instead.

**The fact that convinced me it isn't a capability reduction:** in released 4.4.x, `makeResolvedDeepLinkIntent` discards an unresolvable deep link and substitutes the launcher intent, whose `data` is null. `handlePush` reads `intent?.data`, so on that path the handler was **never invoked at all**. A registered handler in 4.4.x never receives a link an intent couldn't have delivered. The only apps that would lose anything are ones relying on the carve-out `rel/4.5.0` added under auto-tracking — which is unreleased.

Two things from this branch are carried forward:
- The `handlePush(intent, dispatchDeepLink)` overload, in #533, with de-duplication split per stage so a host that still calls `handlePush` reaches their handler while `$opened_push` stays exactly-once.
- The `SampleDetailActivity` scaffolding, which is what makes back-stack behavior observable on device.

Also worth carrying over regardless of which approach wins: **`onNewIntent` handling is mandatory, not optional.** Measured — when the process is killed with its task still in recents, Android restores the *old* intent into `onCreate` and delivers the new one to `onNewIntent`. Platform behavior, reproduces without any deep link, but the intent path leans on it much harder.

-- Claude

evan-masseau:
Closing — superseded by #534, which is the same fix without the deferral machinery.

This PR worked: verified on device, the postponed dispatch fired against the host's activity ~70ms after a warm tap and ~112ms after a cold start, and the handler navigated correctly. It just needed a lifecycle observer, a timeout, a drop-on-expiry policy, crash containment inside `onActivityResumed`, and a non-obvious "register from `onCreate`, never `onResume`" invariant that could not be unit-tested without Robolectric. #534 sends an `ACTION_VIEW` intent instead and lets Android do the routing, so none of that exists.

Two things from this branch live on:
- The `handlePush(intent, dispatchDeepLink)` overload is #533, now with de-duplication split per stage so a host that still calls `handlePush` reaches their handler while `$opened_push` stays exactly-once.
- The `sendDeepLinkIntent` fix — it waited up to 50ms for an activity it only used for `packageName` and `packageManager` — is folded into #533 as well.

**#531 stays open and is unaffected.** Its `LifecycleMonitor` race fix is independent of the deep-link approach: three forms and analytics callers use that helper today, and it fixes a real bug where dismissing a form that was waiting to re-present did not deregister the observer, so a later unrelated resume could re-present it.

-- Claude

evan-masseau:
Correction to my close comment above: I said the `sendDeepLinkIntent` fix was "folded into #533 as well." **It is not** — #533 and #534 both still carry the original, which waits up to `ACTIVITY_TRANSITION_GRACE_PERIOD` for an activity it only needs for `packageName` and `packageManager`.

It is a real improvement and still worth doing, just not currently anywhere. Tracked on MAGE-1039 alongside the other carry-forward items rather than being smuggled into an unrelated PR.

Everything else in that comment stands.

-- Claude

PR #533, opened 2026-08-07
feat(analytics): dedupe push open tracking without suppressing deep links

evan-masseau:
# Description
Splits push-open de-duplication into two independently-counted stages — open tracking/dismissal, and deep-link dispatch — each at-most-once per delivery, and lets a caller that delivers the deep link itself suppress dispatch via an intent extra.

Prerequisite for #534, which changes notification taps to deliver deep links by intent. No behavior change on its own beyond the dedup split described below.

## Due Diligence

`handlePush` keeps its single published signature — no public API change. It is now expression-bodied (`= apply { ... }`) to match every other `: Klaviyo` builder on the facade, which compiles to the identical JVM signature (verified with `javap`); plain `apply` rather than `safeApply`, since `KlaviyoPushOpenHandler.handle` already wraps its own work and the codebase warns against nesting the two.

## Changelog / Code Overview

`KlaviyoPushOpenHandler.handle` deduped all of its work behind one guard — a second `handlePush` call for the same delivery returned immediately having done nothing. That single guard is wrong once an SDK entry point handles a tap and then forwards the same intent to the host: the host's own `handlePush` call could no longer reach their registered `DeepLinkHandler`.

The fix is two counters, not zero. Each stage is at-most-once per delivery, tracked in its own `BoundedIdSet`:

| Stage | Deduped | Counter |
|---|---|---|
| `$opened_push` event | yes, first caller wins | `handledPushDeliveries` |
| Notification dismissal | yes | (same) |
| `DeepLinkHandler` invocation | yes, first caller wins | `dispatchedDeepLinks` |

They need separate counters because they do not advance together: a suppressed intent records the open while leaving dispatch unconsumed.

**Ordering constraint.** The suppression check returns *before* the dispatch mark. Marking there would consume the slot belonging to the unflagged intent forwarded to the host — silently dropping the handler call and reintroducing the exact bug this PR fixes. Mutation-checked in both directions.

Resulting semantics, all covered by tests:

- No delivery id (no `tm`, no `NOTIFICATION_UID_EXTRA`) → neither stage is deduped. Unchanged.
- A host calling `handlePush` twice → one event, one navigation.
- Suppressed call, then two unflagged calls → one event, one navigation.
- Non-Klaviyo or null intent → still returns before any stage.

### Suppressing dispatch

A caller that delivers the deep link by another means stamps `Constants.SUPPRESS_DEEP_LINK_HANDLER_EXTRA` on the intent it hands to `handlePush`, and forwards a copy without it. `handle` skips the dispatch stage when the flag is present; tracking and dismissal are unaffected.

An earlier revision of this PR added a `handlePush(intent, dispatchDeepLink)` overload instead. The extra is a better fit: it needs no new public method on the facade (Kotlin `internal` does not cross Gradle module boundaries, so the overload had to be `public` + `@RestrictTo` + `@JvmSynthetic`), it keeps this a `Patch`, and it matches how `BROWSER_URL_EXTRA` already carries internal trampoline routing state on the same intents. Both use the `_klaviyo.` prefix so they stay out of analytics event properties.

The flag must not survive the hop to the host — #534 removes it before forwarding, with a test per destination.

## Test Plan

New tests cover: a suppressed intent still tracks/dismisses/marks the delivery; a later unflagged call invokes the handler without tracking again; a suppressed call followed by two unflagged calls yields exactly one navigation; a delivery with no id dispatches every time; dismissal happens once across repeat deliveries.

Mutation-checked three ways — restoring the whole-function short-circuit, disabling the dispatch counter, and marking dispatch inside the suppression branch each fail the relevant tests.

Exercised end-to-end on an API 36 emulator as part of #534: exactly one `$opened_push` per notification tap.

### Unrelated test-isolation fix

`handlePush continues deep link handling even if opened_push processing fails` stubs `enqueueEvent` to throw `MissingAPIKey`. That is a `KlaviyoException`, so `safeApply(preInitQueue)` queues the operation for replay — and because `Klaviyo` is an object, the queue outlives `unmockkAll()`. The next test's `initialize()` drained it and enqueued a stale `$opened_push` against that test's mock.

This was latent on `rel/4.5.0`; any test asserting `exactly = 1` that happened to run next would fail. Cleanup now clears the queue between tests.

## Related Issues/Tickets
Prerequisite for #534.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved push-open handling to prevent duplicate tracking and notification dismissal for the same delivery.
  * Ensured deep links are dispatched once per delivery while preserving navigation for forwarded notifications.
  * Maintained support for repeated deep-link dispatches when delivery identifiers are unavailable.
  * Added handling for suppressed deep-link callbacks without blocking link delivery.

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
@@ -4,6 +4,7 @@ import android.app.Application
 import android.content.Context
 import android.content.Intent
 import android.net.Uri
+import androidx.annotation.RestrictTo
 import androidx.core.net.toUri
 import com.klaviyo.analytics.linking.DeepLinkHandler
 import com.klaviyo.analytics.linking.DeepLinking
@@ -344,9 +345,26 @@ object Klaviyo {
      * @param intent the [Intent] from opening a notification
      */
     @JvmStatic
-    fun handlePush(intent: Intent?): Klaviyo {
-        KlaviyoPushOpenHandler.handle(intent, preInitQueue)
-        return this
+    fun handlePush(intent: Intent?): Klaviyo = handlePush(intent, dispatchDeepLink = true)
+
+    /**
+     * [handlePush] with control over the deep-link dispatch stage, for SDK entry points that
+     * deliver the link to the host themselves.
+     *
+     * When [dispatchDeepLink] is false the push open is still tracked and the notification still
+     * dismissed, and the delivery is still recorded, so a later [handlePush] call for the same
+     * delivery does not track it again. That later call does still invoke a registered
+     * [DeepLinkHandler], since dispatch is never deduplicated.
+     *
+     * Public only to cross module boundaries within the SDK; not part of the supported API.
+     *
+     * @param intent the [Intent] from opening a notification
+     * @param dispatchDeepLink whether to invoke a registered [DeepLinkHandler] with the link
+     */
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    @JvmSynthetic
+    fun handlePush(intent: Intent?, dispatchDeepLink: Boolean): Klaviyo = apply {
+        KlaviyoPushOpenHandler.handle(intent, preInitQueue, dispatchDeepLink)
     }
 
     /**
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
@@ -4,7 +4,6 @@ import android.app.Application
 import android.content.Context
 import android.content.Intent
 import android.net.Uri
-import androidx.annotation.RestrictTo
 import androidx.core.net.toUri
 import com.klaviyo.analytics.linking.DeepLinkHandler
 import com.klaviyo.analytics.linking.DeepLinking
@@ -345,26 +344,9 @@ object Klaviyo {
      * @param intent the [Intent] from opening a notification
      */
     @JvmStatic
-    fun handlePush(intent: Intent?): Klaviyo = handlePush(intent, dispatchDeepLink = true)
-
-    /**
-     * [handlePush] with control over the deep-link dispatch stage, for SDK entry points that
-     * deliver the link to the host themselves.
-     *
-     * When [dispatchDeepLink] is false the push open is still tracked and the notification still
-     * dismissed, and the delivery is still recorded, so a later [handlePush] call for the same
-     * delivery does not track it again. That later call does still invoke a registered
-     * [DeepLinkHandler], since dispatch is never deduplicated.
-     *
-     * Public only to cross module boundaries within the SDK; not part of the supported API.
-     *
-     * @param intent the [Intent] from opening a notification
-     * @param dispatchDeepLink whether to invoke a registered [DeepLinkHandler] with the link
-     */
-    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-    @JvmSynthetic
-    fun handlePush(intent: Intent?, dispatchDeepLink: Boolean): Klaviyo = apply {
-        KlaviyoPushOpenHandler.handle(intent, preInitQueue, dispatchDeepLink)
+    fun handlePush(intent: Intent?): Klaviyo {
+        KlaviyoPushOpenHandler.handle(intent, preInitQueue)
+        return this
     }
 
     /**
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
@@ -344,9 +344,8 @@ object Klaviyo {
      * @param intent the [Intent] from opening a notification
      */
     @JvmStatic
-    fun handlePush(intent: Intent?): Klaviyo {
+    fun handlePush(intent: Intent?): Klaviyo = apply {
         KlaviyoPushOpenHandler.handle(intent, preInitQueue)
-        return this
     }
 
     /**
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/KlaviyoPushOpenHandler.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/KlaviyoPushOpenHandler.kt
@@ -58,9 +58,8 @@ internal object KlaviyoPushOpenHandler {
         // forwards the same intent to the host, so a manual handlePush call (or singleTask
         // re-entry) would otherwise double-track one tap. A delivery with no id is never deduped.
         val deliveryId = intent.pushDeliveryId
-        val isNewDelivery = deliveryId == null || handledPushDeliveries.markOnce(deliveryId)
 
-        if (isNewDelivery) {
+        if (deliveryId == null || handledPushDeliveries.markOnce(deliveryId)) {
             // Create and enqueue an $opened_push. safeApply(preInitQueue) buffers this for replay
             // if handlePush runs before initialize(), and guards against unexpected failures.
             safeApply(preInitQueue) {
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/KlaviyoPushOpenHandler.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/KlaviyoPushOpenHandler.kt
@@ -37,18 +37,11 @@ internal object KlaviyoPushOpenHandler {
      * Called by [Klaviyo.handlePush]; not meant for direct use outside this module.
      *
      * Tracking and dismissal happen at most once per delivery. Deep-link dispatch is not deduped:
-     * an SDK entry point that suppresses it may forward the same intent to the host, whose own
-     * [Klaviyo.handlePush] call must still reach a registered
+     * an intent flagged with [Constants.SUPPRESS_DEEP_LINK_EXTRA] may be forwarded to the host
+     * without the flag, so the host's own [Klaviyo.handlePush] call still reaches a registered
      * [DeepLinkHandler][com.klaviyo.analytics.linking.DeepLinkHandler].
-     *
-     * @param dispatchDeepLink Whether to invoke a registered handler with the intent's deep link.
-     *  Pass false when the caller delivers the link itself.
      */
-    internal fun handle(
-        intent: Intent?,
-        preInitQueue: Queue<Operation<Unit>>,
-        dispatchDeepLink: Boolean = true
-    ) {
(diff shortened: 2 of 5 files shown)


sdk/analytics/src/main/java/com/klaviyo/analytics/KlaviyoPushOpenHandler.kt
```diff
@@ -35,52 +42,69 @@ internal object KlaviyoPushOpenHandler {
     /**
      * Core push-open handling: guards, event enqueue, notification dismissal, deep-link dispatch.
      * Called by [Klaviyo.handlePush]; not meant for direct use outside this module.
+     *
+     * Tracking/dismissal and deep-link dispatch each happen at most once per delivery, counted
+     * separately: an intent flagged with [Constants.SUPPRESS_DEEP_LINK_HANDLER_EXTRA] records the open
+     * without consuming the dispatch, so the unflagged copy forwarded to the host still reaches a
+     * registered [DeepLinkHandler][com.klaviyo.analytics.linking.DeepLinkHandler].
      */
```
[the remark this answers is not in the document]
evan-masseau: Fixed in `ab7d76b85a` — imported `DeepLinkHandler` and `KlaviyoException` and switched both KDoc links to the short form.

Worth noting the second one was not just verbose: the fully qualified name I had written (`com.klaviyo.core.config.KlaviyoException`) pointed at a package that does not contain that class — it lives in `com.klaviyo.core`. So the link was broken, and importing it surfaced that.

sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoPushOpenHandlerTest.kt
```diff
@@ -260,6 +282,86 @@ internal class KlaviyoPushOpenHandlerTest : BaseTest() {
...
+    fun `handlePush on a suppressed intent tracks and dismisses without invoking the handler`() {
+        val (getCapturedUri) = setupDeepLinkHandler()
+        val testUri = mockk<Uri>()
+
+        Klaviyo.handlePush(suppressedIntent("suppressed-dispatch", testUri))
+
+        verifyOpenedPushEventEnqueued()
+        assertEquals(null, getCapturedUri())
+        verify(exactly = 0) { DeepLinking.handleDeepLink(any()) }
+    }
```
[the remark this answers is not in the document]
evan-masseau: Good catch — fixed in `ab7d76b85a`. The test name promised dismissal and asserted none, so I added the tag, mocked `NotificationManagerCompat`, and now verify exactly one `cancel`.

Asserting rather than renaming is deliberate: suppression is scoped to the deep link handler only, so the open event *and* the dismissal both have to survive it. That is the property this test should be pinning.

Mutation-checked — stubbing out the dismissal branch in `handle` fails this test, so the new assertion is not vacuous.

sdk/core/src/main/java/com/klaviyo/core/Constants.kt
```diff
@@ -44,6 +44,17 @@ object Constants {
...
+    /**
+     * Intent extra marking an intent whose deep link is delivered to the host by another means, so
+     * `Klaviyo.handlePush` skips invoking a registered `DeepLinkHandler` for it. The link itself is
+     * still delivered — only the handler call is suppressed. Absent → handler is invoked.
+     *
+     * Set on the trampoline's own intents and removed before the intent is forwarded to the host, so
+     * a host that calls `handlePush` itself still reaches its handler.
+     * Uses [INTERNAL_PREFIX] to stay out of analytics event properties, like [NOTIFICATION_TAG_EXTRA].
+     */
+    const val SUPPRESS_DEEP_LINK_HANDLER_EXTRA = INTERNAL_PREFIX + "suppress_deep_link_handler"
```
Atharva8168, line 56: Nice work on this one — the KDoc here does a lot of heavy lifting. Spelling out that the link is still delivered and only the handler call is suppressed, plus the note that the trampoline strips the flag before forwarding, means a future reader doesn't have to reconstruct the two-stage contract from the handler. Placing it in core alongside `NOTIFICATION_TAG_EXTRA`/`NOTIFICATION_UID_EXTRA` with the same `INTERNAL_PREFIX` rationale keeps it consistent with what's already here. Thanks for the care on the naming and the docs.

sdk/analytics/src/main/java/com/klaviyo/analytics/Klaviyo.kt
```diff
@@ -344,9 +344,8 @@ object Klaviyo {
      * @param intent the [Intent] from opening a notification
      */
     @JvmStatic
-    fun handlePush(intent: Intent?): Klaviyo {
+    fun handlePush(intent: Intent?): Klaviyo = apply {
```
Atharva8168, line 347: Small change, but a good one — this was the last `: Klaviyo` builder on the facade still using a block body, so it now matches `setProfile` / `createEvent` / `createSubscription`. I independently checked the `apply` vs `safeApply` call: the nesting warning in `KlaviyoException.kt` is real, and `KlaviyoPushOpenHandler.handle` already wraps each of its stages, so plain `apply` is the right choice here rather than an oversight. Thanks for spelling the reasoning out in the description — saved a reviewer the round trip.

sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoPushOpenHandlerTest.kt
```diff
@@ -112,6 +116,17 @@ internal class KlaviyoPushOpenHandlerTest : BaseTest() {
...
+    /**
+     * Empty [Klaviyo]'s private pre-init queue between tests. A test that makes `enqueueEvent` throw
+     * a [KlaviyoException] leaves the operation queued for replay, and since [Klaviyo] is an object
+     * that queue outlives `unmockkAll`. The next `initialize` would drain it and enqueue that stale
+     * event against the following test's mock.
+     */
+    private fun drainPreInitQueue() = Klaviyo::class.java
+        .getDeclaredField("preInitQueue")
+        .also { it.isAccessible = true }
+        .let { (it.get(Klaviyo) as Queue<*>).clear() }
```
Atharva8168: Really glad you chased this down rather than papering over it — the write-up in the description explains the failure mode better than the code alone could, and the reflection is well-documented at the call site.

One question, non-blocking: the hazard itself is latent on `rel/4.5.0` for any test class that calls `Klaviyo.initialize` after something queues a failed operation, but the fix here is scoped to this class. Do you think that's worth a follow-up ticket — either lifting `drainPreInitQueue` into `BaseTest` or otherwise making the queue reset the default — so the next class to hit it doesn't have to rediscover it?

Atharva8168 (approved):
Overall LGTM. Just mentioned few small non blocking suggestions.

sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoPushOpenHandlerTest.kt
evan-masseau: Fixed here rather than ticketed — `c0151a04fc`. You're right that the hazard is latent for any class, and "easy to fix later" tends to mean "rediscovered later at the cost of another debugging session," which is what this one already cost once.

`drainPreInitQueue` now lives in `BaseTest.cleanup()`, so every test class gets it. `fixtures` already depends on `:sdk:analytics`, so no new dependency edge. Wrapped in `runCatching` — the field is private and reflective, and a fixture teardown shouldn't be able to fail a passing test if that field is ever renamed.

Verified it isn't just moved dead code: deleting the call from `BaseTest.cleanup` fails exactly one test in `:sdk:analytics` (489 completed, 1 failed), the same isolation failure as before. So the guard is load-bearing from its new home.

Also dropped two imports from `KlaviyoPushOpenHandlerTest` that the moved KDoc was the last user of.

-- Claude

sdk/fixtures/src/main/java/com/klaviyo/fixtures/BaseTest.kt
```diff
@@ -173,9 +175,23 @@ abstract class BaseTest {
...
+     * queued for replay, and since [Klaviyo] is an object that queue outlives `unmockkAll` and the
+     * test class itself. The next `initialize` in any test would drain it and replay that stale
+     * operation against a different test's mocks.
+     */
+    protected fun drainPreInitQueue() = runCatching {
+        Klaviyo::class.java
+            .getDeclaredField("preInitQueue")
+            .also { it.isAccessible = true }
+            .let { (it.get(Klaviyo) as Queue<*>).clear() }
+    }
```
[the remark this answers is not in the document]
evan-masseau: Valid, and it resolved itself: `524b0664b5` reverts the `BaseTest` hoist entirely, and the restored class-local helper never had the `runCatching` — it throws on reflection failure, which is the behavior you're asking for.

The underlying point was right and I'd argued the wrong way for it. My rationale was "a fixture teardown shouldn't fail a passing test if the field gets renamed," which is backwards for this specific helper: it exists *because* silent cross-test pollution cost a debugging session. A guard that fails silently restores that original bug while still looking present in the source — strictly worse than the failure it was avoiding.

One note on the suggested diff: `runCatching { ... }.getOrThrow()` is behaviorally identical to not wrapping at all, so the committable suggestion is the convoluted form of the fix. The first option in your prose — "remove `runCatching`" — was the clean one.

-- Claude

sdk/analytics/src/test/java/com/klaviyo/analytics/KlaviyoPushOpenHandlerTest.kt
evan-masseau: Reversing this — Evan pushed back and he's right. Reverted in `524b0664b5`; the helper is class-local again, as it was when you first reviewed it.

The hoist traded one problem for a worse one. To spare a hypothetical future class from rediscovering this, it made **every test class in every module** run the drain in teardown — including `core` and `location` classes that never touch `Klaviyo` — and introduced a fixture→analytics coupling that didn't exist before. That's `BaseTest` drifting toward an uber-fixture that does a bit of everything for everyone, which is a slow leak nobody unwinds later.

So the answer to your original question is "yes it's a latent hazard, no it shouldn't be solved by making it everyone's problem." If a second class hits it, the helper is three lines to copy or to lift *then* — with two real call sites to justify it rather than one plus a hypothetical.

CodeRabbit separately flagged that my hoisted version wrapped the reflection in `runCatching`, swallowing failures. That was a real regression and the revert removes it too: the class-local version throws, which is correct for a guard whose entire purpose is making silent pollution loud.

-- Claude

evan-masseau: To add some human input to this conversation... 

My longer term vision for `BaseTest.kt` would be to make it _less_ of a monolithic uber-class, and this would have been a step in the wrong direction IMO. Its easy enough to extract something into composable, shareable pieces later anyway.

PR #536, opened 2026-08-19
feat(push): make automatic_push_token_forwarding three-valued

evan-masseau:
Part of MAGE-1113

## Summary

`automatic_push_token_forwarding` gates two paths: the **reactive** forward in `KlaviyoPushService.onNewToken`, and the **proactive** fetch in `PushTokenFetcher.maybeAutoRegisterPushToken` (at `initialize` and each foreground). Only the reactive path existed before the flag, so a boolean defaulting to `true` opts every existing integration into new behavior, and no value means "what I had before."

This makes the flag three-valued via `AutomaticPushTokenForwarding`:

| Value | Reactive `onNewToken` | Proactive fetch |
|---|---|---|
| not set | yes | no |
| `true` | yes | yes |
| `false` | no | no |

Call sites read `forwardsOnTokenRotation` / `fetchesProactively` instead of comparing enum entries. Resolved from a `Context` in push-fcm (safe before `initialize`) and from `Config` in analytics. `Klaviyo.setPushToken` is unaffected in all three states. `Constants.AUTOMATIC_PUSH_TOKEN_FORWARDING_DEFAULT` is removed — nothing outside tests referenced it.

Telemetry is unchanged: `SdkFeatures.headerValue` already gated on `hasManifestKey` and omitted absent keys, so behavior now matches what telemetry already modeled.

## Changelog

- `automatic_push_token_forwarding` is now three-valued; leaving it unset preserves the SDK's previous token-forwarding behavior, and `true` additionally registers the token at `initialize` and on each foreground.
- Added an `unset` sample flavor, so the sample demonstrates all three states.
- Docs updated to describe both automatic push flags as opt-in.

## Test plan

Unit tests: `:sdk:core`, `:sdk:push-fcm`, `:sdk:analytics` all pass (forced full rerun). Six existing test sites converted from a boolean default to explicit three-state helpers, with new cases for the unset state in `PushTokenFetcherTest`, `KlaviyoPushServiceTest`, and `StateSideEffectsTest`.

Manual verification on an emulator, comparing each state against tag `4.4.1` on an identical scenario (cold start, then two background/foreground cycles), counting `Enqueuing Push Token request`:

1. `./gradlew :sample:assembleUnsetDebug :sample:assembleManualDebug :sample:assembleAutomaticDebug`
2. For each flavor: uninstall, install, grant `POST_NOTIFICATIONS`, cold start, settle 20s, then two background/foreground cycles.

| Build | cold start | resume 1 | resume 2 |
|---|---|---|---|
| `4.4.1` (baseline) | 1 | 0 | 0 |
| `unset` | 1 | 0 | 0 |
| `automatic` | 1 (4 fetches, deduped) | 0 | 0 |
| `manual` | 1 (explicit call only; both automatic paths skipped) | 0 | 0 |

Also verified with `initialize` deliberately delayed past FCM's `onNewToken` (the wrapper-SDK ordering): `unset` matches pre-flag behavior, and `true` registers the token via the fetch at `initialize`.

Merged-manifest check confirms no SDK library manifest declares the key, so the unset state cannot be introduced accidentally by manifest merging.

## Note

Under an unset flag, a token delivered before `Klaviyo.initialize` is dropped, because `Klaviyo.setPushToken` does not participate in `preInitQueue`. That is pre-existing behavior this change preserves deliberately; `true` heals it via the fetch at initialize. Fixing the drop would change unset behavior and create profiles for integrations that never opted in, so it belongs with the next major, alongside removing the unset state.

## Additional verification — API 36

Re-ran on API 36 (`ProxyFriendly_36`, Android 16) to exercise the `POST_NOTIFICATIONS` runtime-permission path, which API 27 cannot.

**Notification-permission changes are still detected with the flag unset**, matching `4.4.1` at every step. `push_state` is assigned from `PushTokenApiRequest(...).requestBody`, whose getter injects `enablement_status`, so the permission state is inside the string change detection compares:

| Step | `unset` | `4.4.1` | `enablement_status` |
|---|---|---|---|
| Cold start, granted | 1 | 1 | `AUTHORIZED` |
| Revoke → foreground | 1 | 1 | `UNAUTHORIZED` |
| Re-grant → foreground | 1 | 1 | `AUTHORIZED` |
| Foreground, no change | 0 | 0 | suppressed |

The proactive fetch never dispatched in any step, so all three requests came from the `refreshPushState()` fallback. Note that revoking a runtime permission kills the process on API 36, so the revoke step is a cold start in both builds; the re-grant step is a genuine resume and demonstrates the fallback path directly.

**Explicit `false` produces zero requests**, observed directly rather than by elimination, using the `unset` flavor's Activity (which does not call `setPushToken` itself) with the flag forced to `false`: 0 requests enqueued, 0 outbound push-token traffic, and `push_state` never written.

**The flavor matrix is identical to API 27** on every axis, including gate-skip and fetch counts.

## Summary by CodeRabbit

* **New Features**
  * Added three-state automatic push-token forwarding: unset, enabled, or disabled.
  * Preserved manual push-token registration in all configurations.
  * Added a sample integration demonstrating unset forwarding and tracking settings.

* **Bug Fixes**
  * Push-token forwarding and proactive fetching now follow the configured state independently.
  * Explicitly disabling automatic forwarding no longer affects manual token registration.

* **Documentation**
  * Updated setup guides with revised forwarding behavior, integration options, and anonymous-profile guidance.


MIGRATION_GUIDE.md
```diff
@@ -49,28 +47,38 @@ override fun onNewIntent(intent: Intent?) {
...
 
-When this flag is enabled (which is the **default**), the SDK automatically registers the device's
-FCM push token with Klaviyo on your behalf — you no longer need to call `Klaviyo.setPushToken`
-manually on app startup.
+This flag has three states, because leaving it unset is different from setting it to `false`:
 
-To opt out and retain manual control, see
+| Value | Behavior |
+|---|---|
+| **not set** | The SDK registers the token whenever FCM issues a new one. This is what the SDK has always done, so upgrading changes nothing. |
```
evan-masseau: More specifically I think this should say "behavior unchanged from 4.4.1 and earlier".

MIGRATION_GUIDE.md
```diff
@@ -49,28 +47,38 @@ override fun onNewIntent(intent: Intent?) {
...
+
+Your own `Klaviyo.setPushToken` calls always work, in all three states.
+
+For setup instructions, see
 [Option A — Automatic Integration](./README.md#option-a--automatic-integration) in the README.
 
 ### Rollout summary
 
-| Behavior | This release (4.5.0) | Future major release |
+| Behavior | This release (4.5.1) | Future major release |
```
evan-masseau: Hm this table feels potentially confusing, I'd just drop it, seems duplicative of the one above.

README.md
```diff
@@ -522,12 +522,24 @@ push integration that otherwise require code in your app:
...
+
+| Value | Behavior |
+|---|---|
+| **not set** | The SDK registers the token whenever FCM issues a new one. This is what the SDK has always done. |
+| **`true`** | Additionally registers the current token at `Klaviyo.initialize` and on each foreground. |
+| **`false`** | The SDK never registers the token automatically — see [Option B](#option-b--manual-integration). |
+
+Set it to `true` if you want the SDK to own the token completely. That also covers a token that
+already existed before you integrated the SDK, which the unset behavior does not: FCM calls
+`onNewToken` only when it generates or rotates a token.
```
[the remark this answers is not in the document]
evan-masseau: I think this is already covered in how we talk about "Advanced" push integration and how you should subclass our push service

sample/src/unset/java/com/klaviyo/sample/SampleActivity.kt
```diff
@@ -0,0 +1,175 @@
...
+ * predates those flags gets:
+ *  - Push token: this app registers no token itself, and the SDK does not fetch one at
+ *    Klaviyo.initialize() or on foreground. Tokens reach Klaviyo only when the push provider delivers
+ *    one to KlaviyoPushService.onNewToken. Contrast with `automatic` (proactive fetch) and `manual`
+ *    (this app fetches and forwards).
+ *  - Push opens: no automatic open tracking, so this Activity calls Klaviyo.handlePush() itself, the
+ *    same as the `manual` flavor. Deep links arrive via the handler registered in SampleApplication.
+ * See sample/README.md.
+ */
+class SampleActivity : ComponentActivity() {
```
evan-masseau, line 31: Maybe not worth doing in this PR, but instead of re-declaring a lot of the same Activity code in multiple flavors, could have a single activity, but use a `BuildConfig` or something and use a flag to conditionally show the differences.

sdk/push-fcm/src/main/java/com/klaviyo/pushFcm/KlaviyoPushService.kt
```diff
@@ -50,15 +51,12 @@ open class KlaviyoPushService : FirebaseMessagingService() {
...
-        )
-        if (forwardingEnabled) {
+        val forwarding = applicationContext.automaticPushTokenForwarding()
+        if (forwarding.forwardsOnTokenRotation) {
             Klaviyo.setPushToken(newToken)
         } else {
             Registry.log.verbose(
-                "Skipping automatic push token forwarding (automaticTokenForwarding=false)"
+                "Skipping automatic push token forwarding (automaticTokenForwarding=$forwarding)"
             )
```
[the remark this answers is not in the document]
glenn-klaviyo: This was how token forwarding worked prior to the 4.5.0 and 4.5.1 releases. We encouraged folks to fetch their latest token `onCreate` to ensure the token was set, even if it wasn't picked up by the KlaviyoPushService. https://github.com/klaviyo/klaviyo-android-sdk/tree/rel/4.4.1#collecting-push-tokens

> In order to send push notifications to your users, you must collect their push tokens and register them with Klaviyo. This is done via the Klaviyo.setPushToken method, which registers the push token and current authorization state via the [Create Client Push Token API](https://developers.klaviyo.com/en/reference/create_client_push_token). The SDK's KlaviyoPushService will automatically receive new push tokens via the onNewToken method. We also recommend retrieving the latest token value on app startup and registering it with the Klaviyo SDK. Add the following to your application or main activity's .onCreate() method:

```
import com.google.firebase.messaging.FirebaseMessaging
import com.klaviyo.analytics.Klaviyo

// Fetches the current push token and registers with Push SDK
FirebaseMessaging.getInstance().token.addOnSuccessListener { pushToken ->
    Klaviyo.setPushToken(pushToken)
}
```

MIGRATION_GUIDE.md
glenn-klaviyo: yeah we agree with this to drop this table in favor of the one above. cc: @Atharva8168

MIGRATION_GUIDE.md
```diff
@@ -2,22 +2,20 @@
...
 ## New Automatic Push Behaviors (Manifest Flags)
 
-SDK 4.5.0 introduces two push behaviors controlled via `AndroidManifest.xml` metadata:
-`automatic_push_open_tracking` (new, opt-in) and `automatic_push_token_forwarding` (**on by
-default**, formalizing the SDK's existing automatic token forwarding). No action is required to
-preserve your current behavior. If you manage push tokens manually and want to disable automatic
-forwarding, opt out as described below.
+SDK 4.5.1 introduces two push behaviors controlled via `AndroidManifest.xml` metadata:
+`automatic_push_open_tracking` and `automatic_push_token_forwarding`. Both are opt-in, so
+**no action is required** — leave them unset and your current behavior is unchanged.
```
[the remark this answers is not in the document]
glenn-klaviyo: we don't believe we need to address this as the read me explains this. cc: @Atharva8168

glenn-klaviyo (approved):
I feel good about these changes. tested locally across the various flavors. 

@evan-masseau i feel like we should address the migration guide updates you pointed out and then we should be in a good place. Nothing else standing out to me at this time.

Atharva8168 (approved):
Went over with @glenn-klaviyo on every Claude suggestion and CodeRabbit findings. Tested locally.
We feel good approving this. Nothing blocking.

glenn-klaviyo:
@evan-masseau should we mention in the ~readme~ migration guide as part of this PR about the previous 4.5.0 version not existing any more and why?

MIGRATION_GUIDE.md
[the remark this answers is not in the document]
evan-masseau: agreed

sample/src/unset/java/com/klaviyo/sample/SampleActivity.kt
evan-masseau: ticketed for later

evan-masseau:
> @evan-masseau should we mention in the ~readme~ migration guide as part of this PR about the previous 4.5.0 version not existing any more and why?

@glenn-klaviyo I'm gonna say lets just put it in the release notes and not the migration guide.

PR #540, opened 2026-09-04
fix(analytics): give each event consumer its own copy (MAGE-1187)

evan-masseau:
# Description
Fixes shared `Event` instance mutation between the event buffer and observers, which could silently null out `$event_id`, `_time`, and `value` on replayed events.

## Due Diligence

## Changelog / Code Overview

`KlaviyoJsBridge.profileEvent()` uses `Event.pop()` to hoist `$event_id`, `_time`, and `value` out of the property map into positional JS arguments — intentional, so those keys aren't duplicated inside the `event.toMap()` passed as the fifth argument. `pop()` is destructive (removes the key), and the `Event` instance being drained was the **same object reference** stored in `GenericEventBuffer`. `KlaviyoState.createEvent` handed one instance to both the buffer and the observers, so a live delivery could drain the buffered copy, and a later replay (e.g. after a webview rebuild on session expiry or API key change) would emit `window.profileEvent("$opened_push", null, null, null, {...})` instead of the real values.

Fix: copy the event at both points where it's published — once per observer in the state broadcast (`KlaviyoState.createEvent`), and once per element in `GenericEventBuffer.getEvents()`. This gives proper isolation so no consumer can affect the buffer or another consumer, rather than a narrower fix targeted at just the forms bridge. Also documented `BaseModel.pop()` as destructive (the name alone reads like a queue pop) and updated the `getEvents()` KDoc to describe the copy-on-read behavior.

`Event` has no `equals` override (identity equality), so ~11 existing assertions in `GenericEventBufferTest` that previously compared event instances directly now compare metric + properties via a small `assertSameEvent` helper, since the buffer now hands out copies rather than the original references. This accounts for most of the test diff size relative to the production diff.

**Regression history:** Introduced in 4.1.0 (`GenericEventBuffer` + the forms `ProfileEventObserver` + the `pop` calls were all added together), present through the current release. Impact is low-to-moderate — `metric` and the full `properties` map survive the pop intact, so trigger matching that keys on those still works (manually verified a post-expiry push-open form still displays). What's lost is `unique_id`, `_time`, and `value`, which may matter for dedupe or time-windowed logic on the onsite (klaviyo.js) side — unconfirmed, since that consumer isn't in this repo. The failure was silent: `window.profileEvent` is a pass-through that returns `true` regardless, so the SDK logged a successful evaluation either way.

This was found while investigating MAGE-1167 (Holafly in-app forms discrepancy), but was ruled out as the cause — Android doesn't have the iOS session-expiry observer bug that ticket describes. Mentioning it here only as investigation context, not as something this PR fixes.

## Test Plan

TDD: wrote failing tests first and confirmed RED for the right reason (`expected:<test-uuid> but was:<null>`), then applied the fix and confirmed GREEN.

- New test: `getEvents returns copies so consumers cannot mutate the buffered event` (`GenericEventBufferTest`)
- New test: `createEvent gives each observer its own copy to consume destructively` (`KlaviyoStateTest`) — verified it fails without the write-side copy
- `./gradlew :sdk:analytics:testDebugUnitTest :sdk:forms:testDebugUnitTest :sdk:core:testDebugUnitTest` — all pass
- `./gradlew ktlintCheck` — clean

## Related Issues/Tickets
Closes MAGE-1187

### DIFF
--- a/sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
+++ b/sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
@@ -209,7 +209,7 @@ internal class KlaviyoState : State {
             // Add enriched event to buffer for multi-consumer access
             GenericEventBuffer.addEvent(enrichedEvent)
             eventObserver.forEach {
-                it?.invoke(enrichedEvent)
+                it?.invoke(enrichedEvent.copy())
             }
         }
 
--- a/sdk/analytics/src/test/java/com/klaviyo/analytics/state/GenericEventBufferTest.kt
+++ b/sdk/analytics/src/test/java/com/klaviyo/analytics/state/GenericEventBufferTest.kt
@@ -44,7 +44,7 @@ internal class GenericEventBufferTest : BaseTest() {
 
         val bufferedEvents = GenericEventBuffer.getEvents()
         assertEquals(1, bufferedEvents.size)
-        assertEquals(event, bufferedEvents[0])
+        assertSameEvent(event, bufferedEvents[0])
     }
 
     @Test
@@ -68,9 +68,9 @@ internal class GenericEventBufferTest : BaseTest() {
 
         val bufferedEvents = GenericEventBuffer.getEvents()
         assertEquals(3, bufferedEvents.size)
-        assertEquals(event1, bufferedEvents[0])
-        assertEquals(event2, bufferedEvents[1])
-        assertEquals(event3, bufferedEvents[2])
+        assertSameEvent(event1, bufferedEvents[0])
+        assertSameEvent(event2, bufferedEvents[1])
+        assertSameEvent(event3, bufferedEvents[2])
     }
 
     @Test
@@ -87,8 +87,8 @@ internal class GenericEventBufferTest : BaseTest() {
 
         val bufferedEvents = GenericEventBuffer.getEvents()
         assertEquals(2, bufferedEvents.size)
-        assertTrue(bufferedEvents.contains(event1))
-        assertTrue(bufferedEvents.contains(event2))
+        assertTrue(bufferedEvents.any { it.metric.name == event1.metric.name })
+        assertTrue(bufferedEvents.any { it.metric.name == event2.metric.name })
     }
 
     @Test
@@ -110,7 +110,7 @@ internal class GenericEventBufferTest : BaseTest() {
 
         val secondRetrieval = GenericEventBuffer.getEvents()
         assertEquals(1, secondRetrieval.size)
-        assertEquals(event, secondRetrieval[0])
+        assertSameEvent(event, secondRetrieval[0])
     }
 
     @Test
@@ -138,7 +138,7 @@ internal class GenericEventBufferTest : BaseTest() {
 
         val bufferedEvents = GenericEventBuffer.getEvents()
         assertEquals(1, bufferedEvents.size)
-        assertEquals(event2, bufferedEvents[0])
+        assertSameEvent(event2, bufferedEvents[0])
     }
 
     @Test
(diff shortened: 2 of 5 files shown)


evan-masseau:
bugbot run

sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
```diff
@@ -209,7 +209,7 @@ internal class KlaviyoState : State {
             // Add enriched event to buffer for multi-consumer access
             GenericEventBuffer.addEvent(enrichedEvent)
             eventObserver.forEach {
-                it?.invoke(enrichedEvent)
+                it?.invoke(enrichedEvent.copy())
```
Atharva8168: Nice catch on this one.

One thing: `createEvent` still returns `enrichedEvent`, which is the same instance we hand to the buffer. Nothing in the repo uses that return value today, but `State` is a public interface, so a caller could still pop the buffered event. Could we copy on the way out too, or note it in the KDoc?

Also worth knowing `copy()` is shallow, so property values are shared references. Fine for `pop()`, just not a deep guarantee.

sdk/analytics/src/test/java/com/klaviyo/analytics/state/GenericEventBufferTest.kt
```diff
@@ -87,8 +87,8 @@ internal class GenericEventBufferTest : BaseTest() {
 
         val bufferedEvents = GenericEventBuffer.getEvents()
         assertEquals(2, bufferedEvents.size)
-        assertTrue(bufferedEvents.contains(event1))
-        assertTrue(bufferedEvents.contains(event2))
+        assertTrue(bufferedEvents.any { it.metric.name == event1.metric.name })
```
Atharva8168: Small one: this pair went to a metric-name-only check while the rest of the file moved to `assertSameEvent`. That drops the properties comparison here. Any reason not to use the helper?

Atharva8168 (approved):
Approving, two small things.

`KlaviyoJsBridge.profileEvent` still pops from the event it's handed. Fine now since both callers pass a copy, but the safety lives in a different module from the code that needs it, and nothing catches a third caller that forgets. 
Those three pops are the only production uses of `pop()` in the repo, so `toMap()` minus the three keys would let us drop it. 
Fine as a follow-up, just didn't want it lost.

Also `KlaviyoStateTest` clears the buffer inside the tests but not in `@After`. It's an object, so the new test leaves an event behind for the next class. `GenericEventBufferTest` clears in both.

sdk/analytics/src/main/java/com/klaviyo/analytics/state/KlaviyoState.kt
evan-masseau: Done in 1b90551 — `createEvent` now returns its own copy, so the buffer, each observer, and the caller all hold separate instances. Added a test that pops off the returned event and checks the buffered one is untouched.

— Claude

sdk/analytics/src/test/java/com/klaviyo/analytics/state/GenericEventBufferTest.kt
evan-masseau: Switched to `assertSameEvent` in 1b90551.

— Claude

evan-masseau:
@Atharva8168 re: your review notes —

- **`KlaviyoStateTest` teardown:** added `GenericEventBuffer.clearBuffer()` to `@After` in 1b90551.
- **Dropping `pop()`:** keeping it for now. Every consumer gets its own copy after this PR (buffer reads, observers, and the `createEvent` return value), so `profileEvent` popping from what it's handed only drains its own copy.

— Claude

PR #541, opened 2026-09-14
fix(analytics): bound the API queue on restore and batch eviction writes

evan-masseau:
# Description
Fixes an `OutOfMemoryError` a customer hit after upgrading to 4.5.1: restoring a persisted API queue from a pre-4.5.0 SDK (unbounded queue) followed by evicting the entire backlog one `SharedPreferences.apply()` call at a time can blow the heap.

## Due Diligence

## Changelog / Code Overview
Two changes:

1. `KlaviyoApiClient.restoreQueue()` now bounds the queue on read. `selectNewestWithinCapacity()` narrows the persisted uuid list to the newest `MAX_QUEUE_SIZE` (200) **before decoding anything** — reading only each entry's `time` field and releasing that JSON before examining the next — then clears the dropped uuids in one write and decodes only what it retained. Selecting before decoding keeps a legacy backlog from being materialized in memory at all.
2. New `DataStore.clear(Collection<String>)` — a batched removal that does one `edit()`/`apply()` instead of one per key. Declared abstract on the interface (a `forEach` default would contradict the documented single-write contract and silently reintroduce the write amplification this change removes) and implemented in both `SharedPreferencesDataStore` and the `InMemoryDataStore` test fixture.

Eviction on the enqueue path is consolidated into `trimToCapacity()`, which `enqueueRequest` calls once per call rather than evicting per request. Both paths select victims by `queuedTime` rather than deque position: head-of-line requests are inserted at the front but are the *newest*, so trimming positionally would discard priority events such as `$opened_push`.

### Root cause
- The queue cap (`MAX_QUEUE_SIZE = 200`) shipped in 4.5.0 and was enforced only in `enqueueRequest`. `restoreQueue()` had no size guard.
- Clients upgrading from <= 4.4.1 (unbounded Android queue) restore their entire persisted backlog on the next launch — potentially thousands of requests.
- The next `enqueueRequest` call then tries to evict that entire backlog in one synchronous main-thread pass, one `SharedPreferences.apply()` per evicted request.
- Per AOSP (`SharedPreferencesImpl.java:539`, verified identical on android-11.0.0_r1 and android-15.0.0_r1), `commitToMemory()` does `mMap = new HashMap<>(mMap)` — a full copy of the entire preferences map — whenever `mDiskWritesInFlight > 0`. `apply()` writes asynchronously, so a tight loop keeps that counter above zero.
- Copies accumulate while the single `queued-work-looper` thread is stalled inside one slow `writeToFile` call (large prefs file → slow XML serialize + fsync). Each pending `MemoryCommitResult` retains its copied map until the drain completes.

Counterintuitive but worth calling out: eviction count alone does not cause the OOM — nearly every queued `apply()` write is skipped by a generation check in `writeToFile`. The accumulation requires the single drain thread to *stall* on one slow write; write duration is the lever, not the number of evictions.

### Explicitly out of scope
Moving the eviction drain off the main thread. Some launch lag remains for clients with a large backlog, but its dominant cost is the blocking load/parse of the oversized prefs file in `restoreQueue` (the first `getString()` blocks on `awaitLoadedLocked()` until the whole XML is parsed), not the deletes — and it's a one-time migration cost, since the file is rewritten small once trimmed. Making `initialize()` async would change SDK startup semantics and belongs in its own change, not a patch release.

## Test Plan
- 642 unit tests pass (core 139, analytics 503), `ktlintCheck` clean.
- 7 new tests:
  - restore trims an over-capacity queue to the newest 200 (and drops the oldest bodies from the store)
  - restore retains the newest by `queuedTime`, not deque position (seeds a head-of-line request at the front of the index with the newest timestamp and asserts it survives)
  - restore at exactly capacity drops nothing
  - restore does not decode the requests it drops (asserts the decoder runs exactly `MAX_QUEUE_SIZE` times)
  - restore reads each persisted body once (guards against a sort selector re-parsing bodies per comparison)
  - batched clear uses exactly one `edit()`/`apply()` while still broadcasting per key (verified via a registered `StoreObserver`, not log output)
  - empty collection does not open preferences
- Reproduced on a physical device (Android 11) with a seeded ~10k-request queue and `android:largeHeap="true"` to match the customer's 512MB ceiling: `OutOfMemoryError ... target footprint 536870912, growth limit 536870912`. With this branch, no crash under identical conditions. Re-verified on the final revision after the review round.

## Related Issues/Tickets
Part of [MAGE-1219](https://linear.app/klaviyo/issue/MAGE-1219/prevent-android-sdk-event-queue-oom-crashes)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved analytics queue recovery when stored requests exceed capacity by retaining the newest requests and removing older ones.
  - Queue limits are now enforced consistently after requests are added.
  - Discarded requests are removed from persistent storage to prevent stale data from accumulating.
  - Queue restoration prioritizes request timestamps over storage order and removes duplicate entries.

- **Reliability**
  - Added more efficient bulk clearing of stored data while preserving change notifications for each removed item.
  - Queues restored at capacity remain intact without unnecessary request removal.


evan-masseau:
@coderabbitai review

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -348,6 +350,22 @@ internal object KlaviyoApiClient : ApiClient {
...
+            if (uuids.size <= MAX_QUEUE_SIZE) return@let uuids
+
+            wasMutated = true
+            val overflow = uuids.size - MAX_QUEUE_SIZE
+            Registry.log.warning(
+                "Persisted queue of ${uuids.size} exceeds capacity ($MAX_QUEUE_SIZE), dropping $overflow oldest"
+            )
+
+            val retained = uuids.takeLast(MAX_QUEUE_SIZE)
+            Registry.dataStore.clear(uuids.take(overflow).toSet() - retained.toSet())
```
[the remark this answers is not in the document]
evan-masseau: Confirmed and fixed in e4fb194. You're right: `offerFirst` puts head-of-line requests at the front and `persistQueue` writes deque order, so `takeLast` dropped exactly the priority requests it should keep.

`restoreQueue` now decodes the full index first, then drops the oldest by `queuedTime` — the same policy `enqueueRequest` uses. Added a regression test (`Restoring an over-capacity queue retains newest by queuedTime not deque position`) that seeds a head-of-line request at the front with the newest timestamp and asserts it survives while the genuinely oldest entry is dropped.

-Claude

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -348,6 +350,22 @@ internal object KlaviyoApiClient : ApiClient {
...
+            // request at a time.
+            if (uuids.size <= MAX_QUEUE_SIZE) return@let uuids
+
+            wasMutated = true
+            val overflow = uuids.size - MAX_QUEUE_SIZE
+            Registry.log.warning(
+                "Persisted queue of ${uuids.size} exceeds capacity ($MAX_QUEUE_SIZE), dropping $overflow oldest"
+            )
+
+            val retained = uuids.takeLast(MAX_QUEUE_SIZE)
```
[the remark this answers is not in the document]
evan-masseau: Same issue Bugbot flagged, and you're both right — fixed in e4fb194. Trimming now happens after decode and selects by `queuedTime` rather than index position, matching `enqueueRequest`'s evict-oldest policy. Regression test added for the head-of-line case.

-Claude

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -221,6 +221,7 @@ internal object KlaviyoApiClient : ApiClient {
             }
         }.forEach { request ->
             if (!apiQueue.contains(request)) {
+                val evictedUuids = mutableListOf<String>()
```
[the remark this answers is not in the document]
evan-masseau: Good catch — fixed in e4fb194. The `evictedUuids` list is now hoisted to cover the entire `enqueueRequest` call and cleared once after the loop, so a multi-request vararg call performs a single batched `apply()` instead of one per request.

-Claude

sdk/core/src/test/java/com/klaviyo/core/model/SharedPreferencesDataStoreTest.kt
```diff
@@ -128,4 +128,32 @@ internal class SharedPreferencesDataStoreTest : BaseTest() {
...
+
+        SharedPreferencesDataStore.clear(keys)
+
+        // One edit/apply pair regardless of how many keys are removed
+        verify(exactly = 1) { preferenceMock.edit() }
+        verify(exactly = 1) { editorMock.apply() }
+        keys.forEach { key -> verify(exactly = 1) { editorMock.remove(key) } }
+
+        // Observers are still notified per key
+        keys.forEach { key -> verify { spyLog.verbose("$key=null") } }
```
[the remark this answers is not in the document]
evan-masseau: Agreed — fixed in e4fb194. The test now registers a MockK `StoreObserver`, asserts exactly one `(key, null)` callback per key, and removes the observer in a `finally` block. Verifying the log text alone would indeed have passed even if the observer broadcast were dropped.

-Claude

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
```diff
@@ -1201,4 +1201,82 @@ internal class KlaviyoApiClientTest : BaseTest() {
...
+    @Test
+    fun `Restoring an over-capacity persisted queue trims to the newest MAX_QUEUE_SIZE`() {
+        mockkObject(KlaviyoApiRequestDecoder)
+        every { KlaviyoApiRequestDecoder.fromJson(any()) } answers { a ->
+            KlaviyoApiRequest(
+                "https://mock.com",
+                RequestMethod.GET,
+                uuid = (a.invocation.args[0] as JSONObject).getString("uuid")
+            )
+        }
```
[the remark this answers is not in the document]
evan-masseau: Fixed in e4fb194. Extracted `seedPersistedQueue(uuids, times)`, which handles the decoder stub and store seeding; each test now contains only its capacity-specific assertions. The helper takes an optional timestamp list so the new head-of-line ordering test can seed a non-positional `queuedTime`.

-Claude

evan-masseau:
@coderabbitai review

evan-masseau:
bugbot run

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -377,6 +384,31 @@ internal object KlaviyoApiClient : ApiClient {
...
+     * Drop the oldest requests by [KlaviyoApiRequest.queuedTime] until the queue is within
+     * [MAX_QUEUE_SIZE], removing them from the persistent store in a single write.
+     *
+     * A store written by an SDK version that predates the queue cap can hold an unbounded number
+     * of requests, which would otherwise leave [enqueueRequest] to evict the backlog one request
+     * at a time.
+     *
+     * @return whether any requests were dropped
+     */
+    private fun trimToCapacity(): Boolean {
```
evan-masseau, line 427: Is this now a bit duplicative of the clearing in `enqueueRequest` i.e. line 251? Would it perhaps be cleaner to just call `trimToCapacity` from `enqueueRequest` to make space and only do queue trimming operations from one place?

evan-masseau: Agreed — done in 195ab855.

`enqueueRequest` now appends unconditionally and calls `trimToCapacity()` once at the end, so victim selection, removal, and the batched store clear all live in one routine. The duplicated `while` loop is gone.

Two things that fell out of it, worth flagging since they are behavior changes rather than pure refactor:

- The queue can now momentarily exceed `MAX_QUEUE_SIZE` within a call (append, then trim) instead of making space first. Net effect is the same at the call boundary, and it fixes a latent edge case: a vararg call larger than the cap previously evicted incrementally per request, whereas one trim at the end retains the genuinely newest N.
- The concurrency gating moved with it. `trimToCapacity` filters on `remove()`'s boolean, so a request another thread already evicted is not double-reported or double-cleared — same best-effort soft bound as before.

640 tests pass, including the two pre-existing capacity tests that assert the head-of-line request survives eviction.

-Claude

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -369,6 +372,10 @@ internal object KlaviyoApiClient : ApiClient {
             }
         }
 
+        if (trimToCapacity()) {
```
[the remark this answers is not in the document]
evan-masseau: I think the issue with this is we can't know the queue times until we've decoded them all so we know which to trim. Maybe there's a creative way around that... Or else, we put in an upper limit above which we ignore the queue time ordering and just hard limit it. This _should_ be a one time concern anyway, since going forward the persisted list should be at or below 200.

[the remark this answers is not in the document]
evan-masseau: Verified and fixed in c920276 — good catch, this one went to the heart of the change set.

You're right that `restoreQueue` decoded every persisted request into `apiQueue` and only trimmed afterward. For a legacy backlog of several thousand requests with large bodies, that materialized all of them before the cap applied — precisely the allocation this PR exists to avoid.

Added `selectNewestWithinCapacity`, which runs on the uuid list before any decoding: it reads only each entry's `time` field, releases that JSON before examining the next, keeps the newest `MAX_QUEUE_SIZE` uuids, and clears the rest in a single batched write. Only retained entries are then decoded. Entries whose timestamp cannot be read sort as oldest and are discarded first, which also prunes corrupt bodies rather than restoring them.

Pinned with a test asserting `KlaviyoApiRequestDecoder.fromJson` is invoked exactly `MAX_QUEUE_SIZE` times when the persisted queue is over capacity — the dropped bodies are never decoded.

One note on your suggested approach: I kept the uuid list itself materialized. It is already parsed as a single `JSONArray` upstream, and a list of 10k uuid strings is a few hundred KB — immaterial next to the request bodies, which is where the memory actually goes. Streaming the index too would add complexity without a measurable benefit.

641 tests pass, ktlint clean.

-Claude

evan-masseau:
@coderabbitai review

evan-masseau:
bugbot run

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -377,6 +368,75 @@ internal object KlaviyoApiClient : ApiClient {
...
+        val retained = uuids.sortedByDescending { uuid ->
+            Registry.dataStore.fetch(uuid)?.let { json ->
+                try {
+                    JSONObject(json).optLong(KlaviyoApiRequest.TIME_JSON_KEY, Long.MIN_VALUE)
+                } catch (exception: JSONException) {
+                    Registry.log.debug("Invalid request JSON $uuid", exception)
+                    Long.MIN_VALUE
+                }
+            } ?: Long.MIN_VALUE
+        }.take(MAX_QUEUE_SIZE).toSet()
```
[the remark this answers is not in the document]
evan-masseau: Confirmed and fixed in d0654b2 — this was a real regression I introduced in the previous commit.

`sortedByDescending` re-invokes its selector on every comparison, so fetching and parsing the request body inside the selector meant re-parsing the whole backlog O(n log n) times. At 10,000 entries that is on the order of 130,000 JSON parses on the calling thread — worse than the behavior it replaced.

Timestamps are now read into a map once via `associateWith`, and the sort reads from that map. Each persisted body is fetched exactly once during selection.

Pinned with a test asserting fetch counts: each dropped uuid is fetched exactly once (selection only) and each retained uuid exactly twice (selection, then decode). A future refactor that moves the read back into a comparator will fail it.

642 tests pass, ktlint clean.

-Claude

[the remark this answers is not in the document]
evan-masseau: Confirmed and fixed in d0654b2 — this was a real regression I introduced in the previous commit.

`sortedByDescending` re-invokes its selector on every comparison, so fetching and parsing the request body inside the selector meant re-parsing the whole backlog O(n log n) times. At 10,000 entries that is on the order of 130,000 JSON parses on the calling thread — worse than the behavior it replaced.

Timestamps are now read into a map once via `associateWith`, and the sort reads from that map. Each persisted body is fetched exactly once during selection.

Pinned with a test asserting fetch counts: each dropped uuid is fetched exactly once (selection only) and each retained uuid exactly twice (selection, then decode). A future refactor that moves the read back into a comparator will fail it.

642 tests pass, ktlint clean.

-Claude

evan-masseau:
@coderabbitai review

evan-masseau:
bugbot run

evan-masseau:
bugbot run

evan-masseau:
@coderabbitai review

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -377,6 +368,81 @@ internal object KlaviyoApiClient : ApiClient {
...
+            .take(overflow)
+            .filter { apiQueue.remove(it) }
+            .map { request ->
+                Registry.log.warning(
+                    "API queue at capacity ($MAX_QUEUE_SIZE), evicting oldest request: ${request.type}"
+                )
+                request.uuid
+            }
+
+        Registry.dataStore.clear(evictedUuids)
```
ndurell, line 477: **The batched write isn't pinned by a test**

This line is the fix — `clear(Collection)` once instead of `clear(String)` N times, i.e. one `apply()` instead of 10,000. As far as I can tell nothing currently holds it in place.

The two capacity tests (`KlaviyoApiClientTest.kt:483` and `:515`) assert only end state:

```kotlin
assertEquals(MAX_QUEUE_SIZE, KlaviyoApiClient.getQueueSize())
assertNull(spyDataStore.fetch("uuid-0"))
assertNotNull(spyDataStore.fetch("uuid-newest"))
verify(atLeast = 1) { spyLog.warning(any(), null) }
```

None of those observe *how many writes* the eviction took. So if someone later "simplifies" this back to a loop:

```kotlin
apiQueue.sortedBy { it.queuedTime }.take(overflow)
    .filter { apiQueue.remove(it) }
    .forEach { Registry.dataStore.clear(it.uuid) }
```

…the queue is still bounded, `uuid-0` is still gone from the store, the warning still fires — every test passes and the crash is back.

`SharedPreferencesDataStoreTest` ("Clearing multiple keys uses a single edit and apply") covers the other half nicely, but it tests the tool rather than whether the eviction path reaches for it:

| Layer | Question | Covered |
|---|---|---|
| `sdk/core` | does `clear(Collection)` batch into one write? | ✅ |
| `sdk/analytics` | does eviction actually *call* `clear(Collection)`? | ❌ |

`spyDataStore` is already a spy (`BaseTest.kt:124`), so one line per capacity test should be enough:

```kotlin
verify(exactly = 1) { spyDataStore.clear(any<Collection<String>>()) }
```

That alone catches it — the loop version calls the `Collection` overload zero times. I'd skip a matching `exactly = 0` on the `clear(String)` overload; `sendQueueSerially` and the decode-failure branch use it legitimately, so it risks a false failure.

Same reasoning applies to the restore path at line 408 if you want the symmetry.

One caveat: I couldn't run this to confirm the overload resolves cleanly in MockK — only Java 21 available in my environment, and CLAUDE.md warns 21 breaks these tests via the reflection-based static field overrides.

---
_Generated by [Claude Code](https://claude.ai/code)_

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -377,6 +368,81 @@ internal object KlaviyoApiClient : ApiClient {
         queueInitialized = true
     }
 
+    /**
+     * Narrow a persisted uuid list to the newest [MAX_QUEUE_SIZE] entries by enqueue timestamp,
+     * clearing the rest from the persistent store.
+     *
+     * A store written by an SDK version that predates the queue cap can hold an unbounded number
+     * of requests. Selecting before decoding keeps the discarded bodies out of memory: only each
+     * entry's timestamp is read here, and the JSON is released before the next entry is examined.
```
ndurell: **The KDoc overstates the memory guarantee**

> Selecting before decoding keeps the discarded bodies out of memory: only each entry's timestamp is read here, and the JSON is released before the next entry is examined.

The second clause is accurate — each `JSONObject` is transient. The first isn't, and the difference matters for whoever maintains this next.

`Registry.dataStore.fetch` is `openSharedPreferences().getString(key, null)` (`SharedPreferencesDataStore.kt:74`). `SharedPreferencesImpl` loads the *entire* prefs file into its in-memory `mMap` on first access and holds it for the process lifetime. So all 10k body strings are already resident before this function reads the first one, and stay resident until the keys are actually removed. Nothing here keeps the discarded bodies out of memory — they were never out of it.

What the change genuinely avoids is materializing 10k **decoded** `KlaviyoApiRequest` objects — parsed headers/query/body per request, which is the real multiplier over the raw strings. That's the win, and I think it's worth stating as the win.

Worth correcting mainly because the PR description already has it right:

> its dominant cost is the blocking load/parse of the oversized prefs file in `restoreQueue` (the first `getString()` blocks on `awaitLoadedLocked()` until the whole XML is parsed), not the deletes

So the description and the KDoc currently contradict each other, and the KDoc is the one that lives on in the codebase. Someone reading it later could reasonably conclude restore is now bounded with respect to persisted file size — it isn't, and can't be while the backing store is SharedPreferences.

Possible rewording:

```kotlin
 * A store written by an SDK version that predates the queue cap can hold an unbounded number
 * of requests. Selecting before decoding keeps the discarded requests from being decoded:
 * only each entry's timestamp is read here, and the JSON is released before the next entry
 * is examined. The raw bodies are already resident in the SharedPreferences map regardless,
 * so this bounds the decoded request objects, not the store's own footprint.
```

Follows on from #discussion_r4009553894, where this mechanism was described but the SharedPreferences residency didn't come up.

---
_Generated by [Claude Code](https://claude.ai/code)_

ndurell (commented):
Took a pass at this — the diagnosis matches the Wyze stack trace exactly and both changes attack the real mechanism. Nice find on the `commitToMemory` full-map copy; that's not an obvious one.

Two non-blocking comments inline, neither a correctness bug:

1. The batched write — the actual fix — has no test pinning it. A refactor back to a per-key loop would keep every test green and reintroduce the crash. One `verify` line per capacity test closes it.
2. The `selectNewestWithinCapacity` KDoc claims a memory bound it doesn't have. The PR description gets this right; the KDoc doesn't, and the KDoc is what survives.

Also worth noting for the release call: `DataStore` is public API and this adds an abstract member to it, which is source- and binary-breaking for any third-party implementor while the PR is marked `Patch`. In practice it looks harmless — `Registry.dataStore` is a hardcoded getter (`Registry.kt:80`), not resolved through the dynamic registry, so nobody can actually substitute a `DataStore` and there's no reason to implement one. Flagging it because there's no binary-compatibility-validator in the build, so CI will never surface this class of change. An interface default (`= keys.forEach(::clear)`) overridden only in `SharedPreferencesDataStore` would sidestep it and drop the `InMemoryDataStore` override, if you'd rather not spend the semver question.

Separately, in answer to the main-thread question from the Slack thread: nothing about `restoreQueue` requires the main thread — it touches no UI and `apiQueue` is a `ConcurrentLinkedDeque`. It's synchronous inside init (`Klaviyo.kt:104` → `KlaviyoApiClient.kt:79`) on the caller's thread. What's actually load-bearing is that `initialize()` returns with the SDK usable, since callers immediately follow it with `setProfile`/`createEvent`. Agreed that's out of scope for a patch — though `Klaviyo.kt:46`'s existing `preInitQueue` means some of the machinery for deferring those calls is already there, which might make the follow-up cheaper than it looks.

---
_Generated by [Claude Code](https://claude.ai/code)_

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
Atharva8168: +1,  the second clause is fine, the `JSONObject` really is transient.

One gap though: the decoded objects aren't "the real multiplier" either. 10k bodies is ~10-20MB, decoded maybe a few times that — neither gets you to 512MB. The heap went to the `new HashMap<>(mMap)` copies: shallow, ~0.5MB each, but ~9,800 of them stacking up while the looper stalls in one slow `writeToFile`.

So this function bounds decode cost; `clear(Collection)` is what actually fixes the OOM.

sdk/core/src/test/java/com/klaviyo/core/model/SharedPreferencesDataStoreTest.kt
```diff
@@ -128,4 +128,39 @@ internal class SharedPreferencesDataStoreTest : BaseTest() {
...
+        keys.forEach { key -> verify(exactly = 1) { observer(key, null) } }
+    }
+
+    @Test
+    fun `Clearing an empty key collection does not open preferences`() {
+        withPreferenceMock()
+
+        SharedPreferencesDataStore.clear(emptyList())
+
+        verify(exactly = 0) { preferenceMock.edit() }
```
Atharva8168: Tiny Nit: the name says `does not open preferences` but this checks edit(), which is a weaker claim, it'd still pass if we opened prefs and just skipped the edit. The early return sits before `openSharedPreferences()`, and `withPreferenceMock` already stubs the context call, so you can assert what the name promises:

```suggestion
        verify(exactly = 0) {
            mockContext.getSharedPreferences(KLAVIYO_PREFS_NAME, Context.MODE_PRIVATE)
        }
```

 Both `KLAVIYO_PREFS_NAME` and `Context` are already imported, so no import churn.

Atharva8168 (commented):
No correctness issues from my side. +1 to @ndurell  on the batched write being untested, that's the one thing that could regress quietly. Rest is nits.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
ndurell: You're right, and the stack trace nails it harder than the estimate does: the failing allocation is **24 bytes** at `HashMap.newNode`, which on ART is exactly one `HashMap.Node` (8-byte header + `hash` + three refs). It died allocating one more Node inside the copy — nothing payload-sized was being requested at the moment of death.

The shallow part is the other half of it: the value `String`s are shared references, so per-copy cost is the table array plus N Nodes — ~305KB at 10k entries by my arithmetic (16384 slots × 4B + 10k × 24B), same order as your 0.5MB. Which means it takes on the order of a thousand *live* copies to reach 512MB, and that lines up with Evan reproducing at ~1K requests plus ballast rather than needing all 9,800 retained at once.

So: withdrawing "the real multiplier." This function bounds decode cost and startup latency; `clear(Collection)` is what fixes the OOM.

That makes the KDoc wording more load-bearing rather than less. If it reads as though selecting-before-decoding is the crash fix, someone could reasonably conclude the batched clear is incidental and fold it back into a per-key loop — which is exactly the regression flagged on line 441, and nothing in the suite would catch it. Revised suggestion, now that it has to carry your point too:

```kotlin
 * A store written by an SDK version that predates the queue cap can hold an unbounded number
 * of requests. Selecting before decoding bounds the decode cost: only each entry's timestamp
 * is read here, and the JSON is released before the next entry is examined. The raw bodies
 * stay resident in the SharedPreferences map either way.
 *
 * This is not what fixes the OOM — the batched [DataStore.clear] is. The crash was thousands
 * of shallow copies of the preferences map piling up while the queued-work looper stalled in
 * one slow write, so clearing these keys one edit at a time would bring it back.
```

---
_Generated by [Claude Code](https://claude.ai/code)_

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
evan-masseau: You're right, and the suggested one-liner works. Fixed in e29c046.

Added `verify(exactly = 1) { spyDataStore.clear(any<Collection<String>>()) }` to both capacity tests and to the restore trim test for the symmetry you mentioned.

I checked it actually catches the regression rather than assuming: I temporarily replaced the batched clear with `evictedUuids.forEach { Registry.dataStore.clear(it) }` — your exact hypothetical — and both capacity tests failed. Reverted after confirming.

Also took your advice to skip an `exactly = 0` on the String overload; `sendQueueSerially` and the decode-failure branch both use it legitimately, so it would be brittle.

On the Java version caveat: the overload resolves fine in MockK on Java 17, no ambiguity.

-Claude

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
evan-masseau: Correct, and this is the more important of the two — the KDoc was making a claim the code doesn't support. Fixed in e29c046, close to your wording:

```
 * of requests. Selecting before decoding keeps the discarded requests from being decoded into
 * [KlaviyoApiRequest] objects: only each entry's timestamp is read here, and the JSON is
 * released before the next entry is examined. The raw bodies are resident in the
 * SharedPreferences map for the process lifetime regardless, so this bounds the decoded
 * request objects, not the store's own footprint.
```

Verified the mechanism: `fetch` is `getString` against the already-loaded `mMap`, so every body string is resident before this function reads its first entry. The win is avoiding N decoded `KlaviyoApiRequest` objects, not avoiding the strings.

Worth adding that on-device profiling agrees with you about which cost dominates. Pixel 2 XL, 10k requests, 55.6MB prefs, median of 3 runs at the default log level: reading the prefs file 2336ms (71%), selecting 895ms (27%), deleting 29ms (0.9%), broadcasting 11ms (0.3%). The file load is the bulk of it and nothing in this PR touches it — which is exactly the misreading your correction prevents.

-Claude

sdk/core/src/test/java/com/klaviyo/core/model/SharedPreferencesDataStoreTest.kt
evan-masseau: Good catch — the name promised more than the assertion checked. Applied your suggestion verbatim in e29c046; the early return does sit before `openSharedPreferences()`, so it now asserts the stronger claim with no import churn.

-Claude

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -377,6 +368,84 @@ internal object KlaviyoApiClient : ApiClient {
...
+            } ?: Long.MIN_VALUE
+        }
+
+        val retained = uuids.sortedByDescending { queuedTimes[it] }
+            .take(MAX_QUEUE_SIZE)
+            .toSet()
+
+        Registry.dataStore.clear(uuids.filterNot(retained::contains))
+
+        return uuids.filter(retained::contains).toTypedArray()
```
ndurell, line 446: **Duplicate uuids in the persisted index bypass the cap**

Surfacing CodeRabbit's finding as an inline thread — it landed as an "outside diff range" note inside a review summary, which is easy to miss.

`retained` is a `Set`, but this line filters the original array, so every *occurrence* of a retained uuid survives:

```kotlin
val retained = uuids.sortedByDescending { queuedTimes[it] }.take(MAX_QUEUE_SIZE).toSet()
return uuids.filter(retained::contains).toTypedArray()
```

250 persisted entries that are all the same uuid: `retained` holds 1, the `filterNot` clears nothing, and the return is all 250. `restoreQueue` then fetches and decodes 250 times.

Two calibrations against CodeRabbit's framing:

- **Not a memory risk.** Line 350's `if (!apiQueue.contains(request))` dedupes by uuid, so `apiQueue` stays bounded at the distinct count. The cost is wasted fetch + decode, not an unbounded queue.
- **But restore won't self-heal it.** `wasMutated` comes from `.also { if (it.size < uuids.size) … }` on line 340. At 250 == 250 that's false, so `persistQueue()` never runs and the duplicated index stays on disk — re-decoded on every cold start until the next `enqueueRequest` rewrites it.

On how much it's worth: normal operation can't produce duplicates, since `enqueueRequest` dedupes by uuid and `persistQueue` writes from the already-deduped queue, so this needs a malformed or legacy index. That said, coping with an untrusted index is this function's whole reason for existing, so "the file would have to be corrupt" is a weaker defense here than it would be elsewhere. There is also at least one real path to duplicates: `queue.optString(i)` returns `""` for any non-string entry, so several malformed entries all collapse to repeated `""` — those happen to be handled already (`fetch` → null → `Long.MIN_VALUE` → sorts oldest → dropped).

Their `uuids.distinct()` fix looks correct, and it makes the function shorter rather than longer. Worth noting it closes the `wasMutated` gap for free: the deduped list *is* shorter than `uuids`, so the flag fires and the normalized index gets persisted.

Heads up that CodeRabbit has auto-paused reviews on this branch, so it won't re-review further pushes without an explicit `@coderabbitai review`.

---
_Generated by [Claude Code](https://claude.ai/code)_

ndurell (commented):
Re-checked `e29c046` — the batched-write assertion and the KDoc correction both look right, and nice job proving the new test actually goes red on the looped version rather than assuming it.

One inline comment: promoting CodeRabbit's duplicate-uuid finding to a real thread, since it posted as an outside-diff note in a summary. I think it's valid, with two adjustments to its framing — it's wasted decode rather than a memory risk, but `wasMutated` also won't fire in the all-duplicate case, so restore doesn't clean up after itself the way the over-capacity path does.

One gap still open on the batched write, if you want it airtight: both capacity tests evict exactly one request, so they prove the `Collection` overload is used but not that N victims collapse into one write. The restore test covers that at N=50; the enqueue path doesn't. `enqueueRequest(*Array(250){…})` on an empty queue would close it — which is the overflow-vararg test CodeRabbit asked for in round one that got lost when the design changed.

---
_Generated by [Claude Code](https://claude.ai/code)_

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
evan-masseau: Confirmed both halves and fixed in fb56ed4. Thanks for pulling it out of the summary — I would have missed it there.

Reproduced the arithmetic before changing anything: 250 entries all sharing one uuid, `retained` holds 1, `filterNot` clears 0, and all 250 come back. And your second point is the one that makes it worth fixing — `wasMutated` compares `250 == 250`, stays false, so `persistQueue()` never runs and the bad index survives every cold start.

Took the `distinct()` approach, applied at the entry point so the whole function works on distinct uuids:

```kotlin
private fun selectNewestWithinCapacity(persisted: Array<String>): Array<String> {
    val uuids = persisted.distinct()
    if (uuids.size <= MAX_QUEUE_SIZE) return uuids.toTypedArray()
```

That closes the `wasMutated` gap as you predicted — the deduped list is shorter, the flag fires, and the normalized index is written back. Added a test asserting a 250-duplicate index restores exactly 1 request and persists as `[\"uuid-dup\"]`.

Agreed on your framing over CodeRabbit's: the `apiQueue.contains` check does bound the queue, so this is wasted fetch/decode rather than a memory risk. And agreed that "the index would have to be malformed" is a weak defense in a function whose entire job is surviving an untrusted index.

Noted on the auto-pause; I am triggering reviews explicitly.

-Claude

evan-masseau:
@coderabbitai review

evan-masseau:
bugbot run

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
evan-masseau: Follow-up on framing: there is no path in the SDK that writes a duplicate uuid — `enqueueRequest` dedupes via `apiQueue.contains` (and `KlaviyoApiRequest.equals` compares by uuid), and `persistQueue` serializes that already-deduped deque. So this is robustness against an index we did not write, not a defect anyone is hitting.

Amended the commit to `refactor` with conditional wording (5fa7cad) so the changelog does not imply a live crash.

The reachable malformed case you mentioned is worth restating: `queue.optString(i)` yields `""` for any non-string element, so several bad entries collapse into repeated empty strings. Those were already survivable — `fetch("")` is null, sorts as `Long.MIN_VALUE`, gets dropped — so `distinct()` only makes that path cheaper rather than fixing it.

Keeping the change on your reasoning rather than on severity: this function's whole purpose is surviving an untrusted index, so "the file would have to be corrupt" is a weaker defense here than elsewhere. It also makes the function shorter and closes the `wasMutated` gap, so it is not carrying complexity for a hypothetical.

-Claude

Atharva8168 (approved):
Nice catch on the commitToMemory map copy, that one was easy to miss.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -377,6 +376,109 @@ internal object KlaviyoApiClient : ApiClient {
...
+     * Reading a request's timestamp costs a JSON parse, so examining every entry of an unbounded
+     * backlog scales without limit — a queue an order of magnitude beyond capacity would spend
+     * seconds on it before the SDK finishes starting up.
+     *
+     * Entries are taken from both ends because each holds requests worth keeping: priority
+     * requests are inserted at the front of the deque, and the most recently enqueued are appended
+     * at the back. Anything in between is older than a full queue's worth of requests on both
+     * sides, and is discarded without being read.
+     */
+    private fun boundCandidates(uuids: List<String>): List<String> {
```
ndurell, line 392: What causes the number of candidates to exceed the queue size?

evan-masseau: Previous versions of the SDK had no upper bound on the queue 😬 
This can be considered a one-time operation, because going forward, the queue size is enforced at writes.

ndurell: ok that makes sense!

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -377,6 +376,109 @@ internal object KlaviyoApiClient : ApiClient {
...
+        // A malformed index can repeat a uuid, and retaining every occurrence of one would both
+        // exceed the cap and re-decode the same request. Deduplicating also marks the index as
+        // mutated, so the normalized form replaces it on disk.
+        val uuids = persisted.distinct()
+        if (uuids.size <= MAX_QUEUE_SIZE) return uuids.toTypedArray()
+
+        val candidates = boundCandidates(uuids)
+
+        // Read each timestamp exactly once: a sort selector is re-invoked per comparison, which
+        // would re-parse every request body O(n log n) times.
```
ndurell, line 428: small nit should this comment about time complexity be on the sort call below?

evan-masseau: I think it belongs here bc the point is we're reading the timestamps ahead of sort, so that the sort isn't re-parsing in the body of the sort function

ndurell (approved):

PR #542, opened 2026-09-15
perf(analytics): bound the work of restoring an oversized queue

evan-masseau:
# Description
Stacked on #541 so the change can be reviewed in isolation. **Base is `bug/MAGE-1219/restore-queue-overload`, not `rel/4.5.2`** — merge #541 first.

Caps the cost of restoring an oversized persisted queue, so startup work is bounded by capacity rather than by how large a backlog happened to grow.

## Due Diligence

## Changelog / Code Overview
`restoreQueue` picks which requests to keep by reading each entry's `queuedTime`, and every read costs a JSON parse. That work scales with the size of the backlog, which nothing bounds.

When the persisted queue exceeds **twice** `MAX_QUEUE_SIZE`, only the first and last `MAX_QUEUE_SIZE` entries are examined; everything between is discarded without being read. Both ends are kept because both hold requests worth saving:

- **Head** — priority requests (`offerFirst`) are inserted at the front
- **Tail** — the most recently enqueued are appended at the back

Anything in the middle is older than a full queue's worth of requests on either side. Below the bound, behavior is unchanged: every entry is read and selection is exact by timestamp.

## Why this is worth doing on its own
The motivation is bounding an unknown, not the time saved in the case we measured. We have no visibility into how large these backlogs actually get on customer devices, and the cost of being wrong is unbounded.

Measured on a Pixel 2 XL (Android 11), 10,000 requests, 55.6MB prefs file, median of 3 runs:

| Backlog | Unbounded selection | Bounded |
|---|---|---|
| 10,000 | ~0.9 s | ~36 ms |
| 100,000 | ~9 s | ~36 ms |
| 1,000,000 | ~90 s | ~36 ms |

At ~0.09ms per entry the unbounded path grows linearly forever; the bounded path is flat.

## What this deliberately does not change
Profiling the same scenario showed where restore time actually goes at the shipping log level:

| Phase | Time | Share |
|---|---|---|
| Reading the prefs file (framework) | 2336 ms | 71.4% |
| Selecting which requests to keep | 895 ms | 27.4% |
| Deleting discarded entries | 29 ms | 0.9% |
| Broadcasting store changes | 11 ms | 0.3% |

Deleting and broadcasting are already negligible — the batched `clear()` in #541 handles the former, and the per-key observer broadcast costs ~1.1µs per key. An earlier reading that suggested broadcasting was expensive turned out to be the cost of verbose logging itself: at log level Verbose it measures ~1054ms, and at the default Warning it is ~11ms, with every other phase unchanged. No change is warranted there.

The dominant cost is the framework's own load and parse of the preferences file, which is unavoidable without moving off SharedPreferences or off the main thread. Both are out of scope for a patch release.

## Test Plan
- 645 unit tests pass, `ktlintCheck` clean.
- 3 new tests:
  - a queue beyond twice capacity examines only the head and tail windows, and entries in between are never read from the store
  - a prioritized request at the front of an oversized index is still retained
  - a queue within twice capacity still reads every entry and selects exactly

## Related Issues/Tickets
Part of [MAGE-1219](https://linear.app/klaviyo/issue/MAGE-1219/prevent-android-sdk-event-queue-oom-crashes)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved restoration of queued analytics requests after app restarts.
  * Large persisted queues now restore up to the supported capacity while prioritizing requests from inspected queue windows.
  * Restoration limits processing for exceptionally large queues to improve reliability.
  * Discarded requests are removed from persistent storage to prevent stale data accumulation.

* **Tests**
  * Added coverage for oversized queues, bounded restoration, request retention, and storage cleanup.


sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -380,17 +380,44 @@ internal object KlaviyoApiClient : ApiClient {
...
+                    "dropping ${uuids.size - MAX_QUEUE_SIZE} oldest"
+            )
+            return uuids.asList()
+        }
 
         Registry.log.warning(
-            "Persisted queue of ${uuids.size} exceeds capacity ($MAX_QUEUE_SIZE), " +
-                "dropping ${uuids.size - MAX_QUEUE_SIZE} oldest"
+            "Persisted queue of ${uuids.size} is more than twice capacity ($MAX_QUEUE_SIZE), " +
+                "discarding all but the first and last $MAX_QUEUE_SIZE without reading them"
```
evan-masseau: i don't think we have to broadcast this algorithm exactly, can instead say something generic like, "discarding the most outdated requests"

evan-masseau: Agreed — done. Both branches now share one message:

```
Persisted queue of N exceeds capacity (200), discarding the most outdated requests
```

Collapsing the two messages also let the two log statements become one, since the only difference between them was the detail you flagged.

-Claude

evan-masseau:
@coderabbitai review

evan-masseau:
bugbot run

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -380,17 +380,38 @@ internal object KlaviyoApiClient : ApiClient {
...
+     * sides, and is discarded without being read.
+     */
+    private fun boundCandidates(uuids: Array<String>): List<String> {
         Registry.log.warning(
             "Persisted queue of ${uuids.size} exceeds capacity ($MAX_QUEUE_SIZE), " +
-                "dropping ${uuids.size - MAX_QUEUE_SIZE} oldest"
+                "discarding the most outdated requests"
         )
 
+        if (uuids.size <= MAX_QUEUE_SIZE * 2) return uuids.asList()
```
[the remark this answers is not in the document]
evan-masseau: Fixed in 4c5c046. Added `MAX_RESTORE_CANDIDATES = MAX_QUEUE_SIZE * 2` with a KDoc explaining what the threshold means, and the check now reads `if (uuids.size <= MAX_RESTORE_CANDIDATES)`. The tests use the constant too, so the relationship is stated once.

-Claude

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
```diff
@@ -1322,4 +1322,72 @@ internal class KlaviyoApiClientTest : BaseTest() {
...
+        uuids.forEach { uuid -> verify(atLeast = 1) { spyDataStore.fetch(uuid) } }
+    }
+
+    @Test
+    fun `Restoring a queue beyond twice capacity clears the unexamined entries from the store`() {
+        val max = KlaviyoApiClient.MAX_QUEUE_SIZE
+        val uuids = (0 until max * 2 + 100).map { "uuid-$it" }
+        seedPersistedQueue(uuids)
+
+        KlaviyoApiClient.restoreQueue(forceRestore = true)
```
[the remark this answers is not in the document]
evan-masseau: Fixed in 4c5c046. Extracted `restoreOversizedQueue()`, which seeds a queue past `MAX_RESTORE_CANDIDATES`, restores it, and returns the uuids in persisted order. Each test keeps only its own assertions.

Agreed on leaving the head-of-line test alone — its `uuids` and `times` fixtures are the point of that test, so folding them into a shared helper would obscure it.

-Claude

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
```diff
@@ -1322,4 +1322,72 @@ internal class KlaviyoApiClientTest : BaseTest() {
...
+
+        // Every uuid absent from the restored queue is also gone from the store, including the
+        // middle band that was discarded without being read. Leaving those behind would keep the
+        // preferences file oversized on every subsequent launch.
+        val restored = spyDataStore.fetch(KlaviyoApiClient.QUEUE_KEY).orEmpty()
+        uuids.filterNot { restored.contains(it) }.forEach { uuid ->
+            assertNull(spyDataStore.fetch(uuid))
```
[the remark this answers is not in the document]
evan-masseau: Good catch on the mechanism — fixed in 4c5c046. The assertion now parses the persisted index into a `Set<String>` and compares exactly:

```kotlin
val restoredJson = JSONArray(spyDataStore.fetch(KlaviyoApiClient.QUEUE_KEY).orEmpty())
val restored = (0 until restoredJson.length()).map { restoredJson.getString(it) }.toSet()
uuids.filterNot(restored::contains).forEach { uuid -> assertNull(spyDataStore.fetch(uuid)) }
```

One correction for the record: the specific example does not apply to this test. With 500 entries the retained windows are `uuid-0..199` and `uuid-300..499`, and the dropped band is `uuid-200..299`. `uuid-30` is retained, not excluded. I checked every dropped uuid against the old substring match and none were masked, so the assertion was not passing vacuously.

Fixing it anyway, because it is latent rather than absent: the masking depends entirely on where the window boundaries land relative to uuid numbering, so changing `MAX_QUEUE_SIZE` or the seed count could silently hollow out the assertion later. Exact membership is the right thing regardless.

-Claude

evan-masseau:
@coderabbitai review

evan-masseau:
bugbot run

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
```diff
@@ -380,17 +388,38 @@ internal object KlaviyoApiClient : ApiClient {
      *
      * @return the uuids to restore, in their original order
      */
-    private fun selectNewestWithinCapacity(uuids: Array<String>): Array<String> {
```
Atharva8168: Two KDoc blocks stacked here, the old `selectNewestWithinCapacity` one (ending in `@return the uuids to restore`) is still sitting right above this new one. 
So `selectNewestWithinCapacity` has no docs now, and its `@return` reads as `boundCandidates` which returns `List<String>`.

Doesn't show up in the diff since those lines are unchanged, which is probably how it slipped past. 

I'd move the old block down above `selectNewestWithinCapacity` rather than delete it and that's the same text @ndurell is asking you to reword on #541, so folding his fix in while you're there saves a conflict.

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
```diff
@@ -1322,4 +1323,79 @@ internal class KlaviyoApiClientTest : BaseTest() {
...
+     */
+    private fun restoreOversizedQueue(): List<String> =
+        (0 until KlaviyoApiClient.MAX_RESTORE_CANDIDATES + 100)
+            .map { "uuid-$it" }
+            .also {
+                seedPersistedQueue(it)
+                KlaviyoApiClient.restoreQueue(forceRestore = true)
+            }
+
+    @Test
```
Atharva8168, line 1372: This only passes because the `verify` comes first, `assertNull(spyDataStore.fetch(ignored))` is itself a call on the spy, so flipping these two lines breaks the `exactly = 0`. 

Nothing says the order matters. The `assertNull` is also mostly covered by the clears-unexamined-entries test below,

```suggestion
        verify(exactly = 0) { spyDataStore.fetch(ignored) }
        // Must stay after the verify above — this call registers on the spy
        assertNull(spyDataStore.fetch(ignored))
```

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
```diff
@@ -1322,4 +1323,79 @@ internal class KlaviyoApiClientTest : BaseTest() {
...
+        assertNull(spyDataStore.fetch(ignored))
+
+        // The most recently enqueued survive
+        assertNotNull(spyDataStore.fetch("uuid-" + (uuids.size - 1)))
+    }
+
+    @Test
+    fun `Restoring a queue beyond twice capacity still retains a prioritized request at the front`() {
+        // A head-of-line request sits at the front of the index with the newest timestamp. The
+        // head window must reach it even though the backlog is too large to examine in full.
```
Atharva8168, line 1392: This is the easy case, priority request at index 0, which the head window always reaches. 
The case the bound actually gives up is one just outside it (say index `max + 50`):- newer than anything in the tail window, dropped unread.

Fine as a trade, but then the KDoc says both ends are kept "because each holds requests worth keeping" and only the reachable end has a test. 
One asserting the stranded request *does* get dropped would make the trade deliberate rather than accidental.

Atharva8168 (commented):
Three notes inline, none blocking. 
The KDoc one is the only real bug, `selectNewestWithinCapacity` lost its docs when the new block went in above it. 
Numbers look convincing and the middle-band clearing checks out.

sdk/analytics/src/main/java/com/klaviyo/analytics/networking/KlaviyoApiClient.kt
evan-masseau: Good catch, and you're right about why it slipped past — the orphaned block sits in unchanged lines, so it never appeared in the diff. Fixed in 3bc8a84.

Moved the original KDoc down onto `selectNewestWithinCapacity` rather than deleting it, and folded in @ndurell's reword from #541 while I was there as you suggested, so the two branches won't conflict on those lines.

-Claude

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
evan-masseau: You're right that the ordering is load-bearing and undocumented — `assertNull(spyDataStore.fetch(ignored))` registers on the spy, so swapping the two lines breaks `exactly = 0`. Took your suggestion verbatim in 3bc8a84.

I hit this exact trap earlier on this PR family: a `verify(exactly = 0)` failed because an `assertNull` above it had already called the spy on the same key. Worth the comment.

Kept the `assertNull` rather than dropping it — agreed it overlaps the clears-unexamined-entries test, but it keeps this test's own claim self-contained.

-Claude

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
evan-masseau: Fair — the existing test covered the easy case, and the trade the bound actually makes had no coverage at all. Added in 3bc8a84:

```kotlin
fun `Restoring a queue beyond twice capacity drops a recent request stranded outside the windows`()
```

It seeds an entry at `max + 50` with `Long.MAX_VALUE` as its timestamp — newer than everything in the tail window — and asserts it is dropped unread while an older entry inside the tail window survives. So the give-up case is now deliberate and pinned rather than implied.

That is the honest cost of using position as a proxy for recency, and I would rather it be stated in a test than discovered later by someone wondering why a recent request vanished.

-Claude

evan-masseau:
@coderabbitai review

evan-masseau:
bugbot run

Atharva8168:
One more small thing on the description, I feel "flat ~36ms regardless of backlog size" reads like the whole restore is bounded, but it's really just the timestamp-parse term. 

The index parse and `optString` copy (:338-339), and the `clear()` of n-200 keys (:440), are all still linear — and the framework's own file load, the 71.4%, is untouched.

Rough math off your own profile: 

0.9s at 27.4% puts total approx 3.3s and framework approx 2.3s at 10k. 

If that scales with file size, 100k is still approx 23s after this. 

Worth shipping either way since it kills the term we actually own

Also cheap and worth it for having a counter when the `> MAX_RESTORE_CANDIDATES` path fires. 
Would turn 200/400 into something we can tune on real data instead of guesses.

evan-masseau:
@Atharva8168 the good news is, this is a one-time operation, fixing old SDK upgrades where the queue was unbounded. Going forward, there will be no path to exceed 200 requests on disk. 

I think claude may have already edits the description you referenced.

Also worth mentioning, there's an upper bound on what is even possible for # of requests in the queue. An app could never have reached _100s of thousands_ because it would have OOMed when it tried to write that many into shared prefs, and/or shared prefs would have limited it. So there's probably a practical upper bound.

Unfortunately, even if we had that counter, we have nowhere to send it. We don't collect telemetrics like that from the host app.

sdk/analytics/src/test/java/com/klaviyo/analytics/networking/KlaviyoApiClientTest.kt
```diff
@@ -1322,4 +1323,101 @@ internal class KlaviyoApiClientTest : BaseTest() {
...
+    }
+
+    @Test
+    fun `Restoring a queue beyond twice capacity drops a recent request stranded outside the windows`() {
+        // The trade this bound makes: an entry just past the head window is newer than everything
+        // in the tail window, but is discarded unread because position stands in for recency.
+        val max = KlaviyoApiClient.MAX_QUEUE_SIZE
+        val uuids = (0 until KlaviyoApiClient.MAX_RESTORE_CANDIDATES + 100).map { "uuid-$it" }
+        val stranded = max + 50
+        val times = uuids.indices.map { if (it == stranded) Long.MAX_VALUE else it.toLong() }
```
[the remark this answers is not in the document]
evan-masseau: Fair, and it matches the repo guideline. Named both as test fields:

```kotlin
/** How far a fixture queue is pushed past the point where entries stop being examined. */
private val excessBeyondCandidates = 100

/** Offset of an entry that falls between the head and tail windows, so is never read. */
private val offsetBetweenWindows = 50
```

The second one earns its name — `max + 50` gave no hint that the point is landing *between* the two windows, which is the whole premise of that test.

-Claude

Atharva8168 (approved):
Gave a second pass on this, LGTM.
Also, looks like CI failures are just infra.

Atharva8168:
Reran CI, All Green Nowwww.

